PR #3, opened 2021-12-02
Add bag2proto script to typescript folder

defunctzombie:
Convert a ros1 bag file to a protobuf mcap file.


defunctzombie:
Might be nice to separate the core library from tools like this?

PR #5, opened 2021-12-16
Enhancements/fixes to bag2proto

defunctzombie:
- Add support for bz2 compressed bags
- Add support for Float64Array handling

### DIFF
--- a/typescript/scripts/bag2proto.ts
+++ b/typescript/scripts/bag2proto.ts
@@ -7,6 +7,7 @@
 import { Bag } from "@foxglove/rosbag";
 import { FileReader } from "@foxglove/rosbag/node";
 import { parse as parseMessageDefinition } from "@foxglove/rosmsg";
+import Bzip2 from "@foxglove/wasm-bz2";
 import { program } from "commander";
 import protobufjs from "protobufjs";
 import descriptor from "protobufjs/ext/descriptor";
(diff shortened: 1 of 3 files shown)


defunctzombie:
Discovered while converting our nuscenes bag to mcap.

typescript/scripts/bag2proto.ts
```diff
@@ -90,13 +104,16 @@ function rosMsgDefinitionToProto(typeName: string, msgDef: string): protobufjs.R
...
+      }
       if (field.arrayLength != undefined) {
         lineComments.push(`length ${field.arrayLength}`);
       }
       fields.push(
-        `${qualifiers.join(" ")} ${field.name} = ${fieldNumber++};${
-          lineComments.length > 0 ? " // " + lineComments.join(", ") : ""
-        }`,
+        `${qualifiers.join(" ")} ${field.name} = ${fieldNumber++} ${
+          isPacked ? "[packed = true]" : ""
```
jtbandes, line 115: I think proto3 uses packed=true by default for scalar types. Why do we need this?

defunctzombie: hm - I was reading docs that made it sound like I should add this.

Re-read the docs and yea we don't need packed. proto3 does it by default for scalar value fields. Will remove.

PR #9, opened 2022-01-10
Support per-channel encryption

wkalt:
Add support for per-channel encryption via key_id and encryption fields
in the channel info record.

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -7,6 +7,7 @@
 [diagram unchunked]: ./diagrams/unchunked.png
 [diagram chunked]: ./diagrams/chunked.png
 [feature explanations]: ./notes/explanatory-notes.md#feature-explanations
+[rfc5652 padding]: https://datatracker.ietf.org/doc/html/rfc5652#section-6.3
 
 > Status: DRAFT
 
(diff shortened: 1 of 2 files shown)


wkalt:
This is currently for discussion purposes:
* Is the feature justified for v1 inclusion?
* Would the solution work? 

I think the main thing to consider with respect to v1 inclusion is that it will be difficult to use this specific implementation (with key ID and encryption in the channel info record) after we have released v1 of the channel info record.

One solution that will remain available after initial release of V1 would be an "encrypted channel info" record, that could store encryption and key ID fields, along with a regular channel info in its data field:

```
encryption string
key_id string
data Channel Info
```

messages could then be encrypted and decrypted with a similar process to the one in this PR.

As for whether this solution works, one thing that has come up in discussion is whether it is acceptable that schemas for encrypted channels are not encrypted. Similar concerns may apply to things like statistics and timestamps. I'm not sure if we have a good reason not to encrypt the schema in the channel info - it seems like we could just use the same key as indicated by the record. I feel that timestamps should remain unencrypted to make it possible for tooling to operate with encrypted and decrypted messages without decrypted (i.e sorting, splitting, merging files). I don't know that I have a feeling on statistics other than that it would be best to minimize the number of cases where data can be present or absent depending on the chosen configuration.

wkalt:
closing - out of near-term scope.

PR #15, opened 2022-01-14
typescript: v0 reading and unindexed writing

jtbandes:
Breaking changes to package exports and new APIs for reading "v0" mcap files as defined in the current spec document.

- Renames existing McapReader/Writer to McapPre0Reader/Writer.
- Adds `Mcap0StreamReader` and `Mcap0IndexedReader`
  - Indexed reader does not yet support overlapping or out of order chunks.
- Adds `Mcap0UnindexedWriter`
- Adds a translation layer for reading pre-0 files and emitting v0 records (`McapPre0LatestStreamReader`)
- Updates `validate` script to try reading v0 files as indexed, and fall back to streamed


typescript/src/Writer.ts
```diff
@@ -0,0 +1,77 @@
+export const LITTLE_ENDIAN = true;
```
defunctzombie, line 1: ? I don't understand why we export this.

typescript/src/getBigUint64.ts
```diff
@@ -0,0 +1,18 @@
+// DataView.getBigUint64 was added to relatively recent versions of Safari. It's pretty easy to
```
defunctzombie, line 1: Maybe a polyfills file is a better home? Or compat?

defunctzombie:
Whats the organization structure here? There are pre0 folders and latest folders but I also see a `typescript/src/latest/McapPre0LatestStreamReader.ts` which lives in a latest folder?

typescript/src/Writer.ts
jtbandes: Probably happened automatically when I did a "move to a separate file" refactor

jtbandes:
> There are pre0 folders and latest folders but I also see a `typescript/src/latest/McapPre0LatestStreamReader.ts` which lives in a latest folder?

Organization is up for discussion– the thinking was since this thing's job is to convert to the latest format, it lives in the latest folder. Or maybe it should live in the pre0 folder instead? 🤷

defunctzombie:
> the thinking was since this thing's job is to convert to the latest format,

👍 I did not realize it is a streaming converter from Pre0 into latest. If it lives in the pre0 folder we can delete that folder when we are ready to drop support for pre0.

typescript/src/Writer.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export default class Writer {
```
defunctzombie, line 3: When working on the bag2proto I felt like this should probably have a better name. Like DatatypeWriter or some such.

typescript/src/getBigUint64.ts
defunctzombie: Also when working on the bag2proto it felt more natural for all the top level files in `src` to live in a _common_ folder since they otherwise add noise to browsing the project

PR #20, opened 2022-01-14
Update bag2proto to write mcap v0

defunctzombie:


typescript/src/v0/Mcap0RecordWriter.ts
```diff
@@ -0,0 +1,90 @@
...
+  constructor(writable: IWritable) {
+    this.writable = writable;
+  }
+
+  async writeMagic(): Promise<void> {
+    await this.writable.write(new Uint8Array(MCAP0_MAGIC));
+  }
+
+  async writeHeader(header: Header): Promise<void> {
+    const serializer = new Writer();
```
jtbandes, line 27:
```suggestion
    const writer = new Writer();
```

defunctzombie: I am not a fan of the class name Writer. In your PR I commented this should be called DatatypeWriter, DatatypeSerializer something other than "writer".

typescript/src/v0/Mcap0UnindexedWriter.ts
```diff
@@ -0,0 +1,60 @@
...
+    await this.recordWriter.writeFooter({
+      indexOffset: 0n,
+      indexCrc: 0,
+    });
+  }
+
+  /**
+   * Add channel info and return a generated channel id. The channel id is used when adding messages.
+   */
+  async registerChannel(info: Omit<ChannelInfo, "channelId">): Promise<number> {
```
jtbandes, line 30: thoughts on `addChannel` vs `registerChannel`? I used `addChannel` for ws-protocol stuff.

defunctzombie: I had addChannel and switched to registerChannel since it returns an ID - don't feel strongly tho

typescript/src/v0/Mcap0RecordWriter.ts
jtbandes: Ok, I don't think "datatype" makes it more clear, but open to other renames. The variable name should match the type if possible though.

typescript/src/v0/Mcap0UnindexedWriter.ts
```diff
@@ -0,0 +1,60 @@
+import { IWritable } from "./IWritable";
+import { Mcap0RecordWriter } from "./Mcap0RecordWriter";
+import { ChannelInfo, Message, Header, Attachment } from "./types";
+
+export class Mcap0UnindexedWriter {
```
jtbandes, line 5: I imagine indexed and unindexed writing could potentially be in the same class, with just a boolean constructor param to indicate whether we should be holding index info in memory and writing it at the end?

jtbandes (approved):
👍 feel free to merge into my branch and we can work on integration there

PR #25, opened 2022-01-19
typescript: add json and protobuf encoding support to validate script

jtbandes:
`validate` script now supports `protobuf` and `json` encodings (similar to [ws-protocol](https://github.com/foxglove/ws-protocol)).


defunctzombie:
👍 

We still need to actually document what encoding means for mcap - right now this validates against a spec that doesn't exist :D

PR #26, opened 2022-01-19
Skeleton C++ project

jhurliman:


cpp/README.md
```diff
@@ -0,0 +1,6 @@
+# C++ implementation of the MCAP (Message Capture) library
```
jtbandes, line 1: we've stopped using the phrase "message capture" so you can just remove that from here and conanfile.py

cpp/mcap/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
jtbandes, line 1: The current LICENSE in the repo root is Apache 2.0

cpp/scripts/format.py
```diff
@@ -0,0 +1,64 @@
+import argparse
```
jtbandes, line 1: any thoughts on how we could avoid copying this between two repos? publish a docker image, or a nodejs binary on npm that can be used with npx (or just pypi), or something?

cpp/.clang-format
```diff
@@ -0,0 +1,30 @@
+---
+Language: Cpp
+Standard: Cpp11
```
jtbandes, line 3: Just realized we can probably change this to Cpp17, or whatever we decide to require...maybe our customers will need support back as far as 14?

cpp/Makefile
```diff
@@ -0,0 +1,17 @@
...
+.PHONY: format-check
+format-check:
+	docker compose run --rm -v .:/src base python3 scripts/format.py /src
+
+.PHONY: format-fix
+format-fix:
+	docker compose run --rm -v .:/src base python3 scripts/format.py --fix /src
+
+.PHONY: bag2mcap
+example_server:
```
jtbandes, line 16:
```suggestion
bag2mcap:
```

cpp/dev.Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+ENV CONAN_V2_MODE=1
+RUN conan config init
+RUN conan profile update settings.compiler.cppstd=17 default
+
+FROM build as build_bag2mcap
+COPY ./examples /src/examples/
+COPY ./mcap /src/mcap/
+COPY ./.clang-format /src/
+RUN conan editable add ./mcap mcap/0.0.1
+RUN conan install examples --install-folder examples/build --build=zlib --build=zstd
```
jtbandes, line 45: Does this work without `--build=zlib --build=zstd`? I only added them for openssl in ws-protocol because there weren't any binaries for this configuration. It would be nice to find a config we can use for development that has prebuilt binaries so we don't have to wait for them to build each time, but I haven't spent time figuring that out yet.

jtbandes (approved):

cpp/dev.Dockerfile
jhurliman: It doesn’t without build=zstd because they only only have gcc binaries not clang

PR #28, opened 2022-01-20
C++ library can write unindexed files

jhurliman:


cpp/examples/bag2mcap.cpp
```diff
@@ -1,15 +1,41 @@
...
-  msg.channelId = 1;
-  msg.sequence = 2;
-  msg.publishTime = 3;
-  msg.recordTime = 4;
-  msg.data.push_back(5);
+  msg.channelId = topic.channelId;
+  msg.sequence = 0;
+  msg.publishTime = now();
+  msg.recordTime = msg.publishTime;
+  msg.data = reinterpret_cast<uint8_t*>(payload.data());
```
jtbandes, line 34: `payload.data()` should already be `uint8_t*`

cpp/mcap/include/mcap/errors.hpp
```diff
@@ -0,0 +1,69 @@
...
+        return "unknown";
+    }
+  }
+
+  virtual std::error_condition default_error_condition(int c) const noexcept override final {
+    switch (static_cast<ErrorCode>(c)) {
+      case ErrorCode::NotOpen:
+        return make_error_condition(std::errc::bad_file_descriptor);
+      case ErrorCode::InvalidChannelId:
+        return make_error_condition(std::errc::invalid_argument);
```
jtbandes, line 51: not very familiar with the `default_error_condition` API, but why would we want to translate to posix error codes when we already have a custom one that exactly describes the error?

cpp/mcap/include/mcap/mcap.hpp
```diff
@@ -1,12 +1,22 @@
 #pragma once
 
+#include "errors.hpp"
+#include <climits>
```
jtbandes, line 4: probably prefer `numeric_limits`?

cpp/mcap/include/mcap/mcap.hpp
```diff
@@ -1,12 +1,22 @@
...
 #include <vector>
 
 namespace mcap {
 
-constexpr char SpecVersionChar = '0';
+#define LIBRARY_VERSION "0.0.1"
+
+constexpr char SpecVersion = '0';
+constexpr char LibraryVersion[] = LIBRARY_VERSION;
+constexpr char Magic[] = {char(137), 77, 67, 65, 80, SpecVersion, 13, 10};  // "\x89MCAP0\r\n"
```
jtbandes, line 19: why does 137 need a cast? does something happen without the cast? maybe this should all be `unsigned char[]`?

cpp/mcap/include/mcap/mcap.hpp
```diff
@@ -45,14 +55,22 @@ struct ChannelInfo {
...
 };
 
 struct Message {
   mcap::ChannelId channelId;
   uint32_t sequence;
   mcap::Timestamp publishTime;
   mcap::Timestamp recordTime;
-  mcap::ByteArray data;
+  uint64_t dataSize;
+  uint8_t* data;
```
jtbandes, line 73:
```suggestion
  uint8_t* data = nullptr;
```

cpp/mcap/include/mcap/mcap.inl
```diff
@@ -0,0 +1,168 @@
...
+}
+
+void McapWriter::registerChannel(mcap::ChannelInfo& info) {
+  info.channelId = uint16_t(channels_.size() + 1);
+  channels_.push_back(info);
+}
+
+std::optional<std::error_code> McapWriter::write(const mcap::Message& message) {
+  if (!stream_) {
+    return std::make_optional(make_error_code(ErrorCode::NotOpen));
```
jtbandes, line 35:
```suggestion
    return make_error_code(ErrorCode::NotOpen);
```

cpp/mcap/include/mcap/mcap.inl
```diff
@@ -0,0 +1,168 @@
...
+  if (!stream_) {
+    return;
+  }
+  write(mcap::Footer{0, 0});
+  writeMagic();
+  stream_->flush();
+  stream_ = nullptr;
+}
+
+void McapWriter::registerChannel(mcap::ChannelInfo& info) {
```
jtbandes, line 28: naming: maybe `addChannel`?

jtbandes: At Cruise we recommended in our style guide to use pointers for output parameters. That way the call site is clear, e.g. `registerChannel(&info)`. I would prefer using the same style in this codebase too

cpp/mcap/include/mcap/mcap.inl
```diff
@@ -0,0 +1,168 @@
...
+}
+
+void McapWriter::write(uint64_t value) {
+  stream_->write(reinterpret_cast<const char*>(&value), sizeof(value));
+}
+
+void McapWriter::write(uint8_t* data, uint64_t size) {
+  stream_->write(reinterpret_cast<const char*>(data), size);
+}
+
```
jtbandes, line 159: Need endianness checks, or (my preference) just use bitwise shift and `&` to write out the bytes you want to write in the order you want them

cpp/mcap/include/mcap/errors.hpp
jhurliman: This is an optional override in the API for users that want to compare your custom error codes against posix error codes. Doing more specific comparisons is always encouraged, but I guess the thinking is if this error happens several levels deep in your codebase and it bubbles up to a generic error handler that has different handling for filesystem or I/O errors vs application logic errors (like `invalid_argument`).

To answer your question, "we" would probably never do this, but it's a common nicety for library users that may do this for whatever reason.

cpp/mcap/include/mcap/mcap.hpp
jhurliman: `numeric_limits` does not define `CHAR_BIT` AFAIK

jtbandes (commented):
ostream is useful as a starting point, but I hope we'll evolve beyond ostream to a simpler API that's more like just write(data,length). In my experience istream is quite slow, but maybe that doesn't apply to ostream or to non-formatted output functions...

cpp/mcap/include/mcap/mcap.hpp
jtbandes: `numeric_limits<unsigned char>::digits`? https://en.cppreference.com/w/cpp/types/numeric_limits/digits

cpp/mcap/include/mcap/mcap.hpp
jhurliman: `char` is a signed 8-bit value, so 137 (being larger than 127) overflows and clang will warn about this. Sadly, our Dockerfile-based build seems to swallow up warnings but I see the warning when I compile directly on MacOS using:

```
Homebrew clang version 13.0.0
Target: arm64-apple-darwin21.1.0
Thread model: posix
InstalledDir: /opt/homebrew/opt/llvm/bin
```

I could make it an unsigned char array, but then I need to typecast when I do an ostream `write()` which wants a char array.

cpp/mcap/include/mcap/mcap.hpp
```diff
@@ -108,4 +127,78 @@ struct UnknownRecord {
...
+  /**
+   * @brief Write an attachment to the output stream.
+   *
+   * @param attachment Attachment to add.
+   * @return An error code on failure.
+   */
+  std::optional<std::error_code> write(const mcap::Attachment& attachment);
+
+private:
+  std::ostream* stream_ = nullptr;
```
jtbandes, line 184: I hate trailing `_` but I'll shut up, it's recommended in the google style guide 😭

cpp/mcap/include/mcap/mcap.inl
jhurliman: I remember that guideline, but I thought there was also a guideline to use `&` for `inout` parameters.

cpp/mcap/include/mcap/mcap.inl
jhurliman: Can we just static_assert() on little-endian?

cpp/mcap/include/mcap/errors.hpp
```diff
@@ -0,0 +1,69 @@
...
+        return std::error_condition(c, *this);
+    }
+  }
+};
+
+}  // namespace detail
+
+const detail::McapErrorCategory& McapErrorCategory() {
+  static detail::McapErrorCategory c;
+  return c;
```
jtbandes, line 62: hm, this probably needs to be `inline` too, but it seems pretty weird to have a `static` local inside an `inline` function 🤔

cpp/mcap/include/mcap/mcap.hpp
jtbandes: > our Dockerfile-based build seems to swallow up warnings

😱

cpp/mcap/include/mcap/mcap.inl
jtbandes: 👍
I do wonder what would go wrong in our code if CHAR_BIT≠8? I guess all these sizeof()s would be wrong?

cpp/mcap/include/mcap/mcap.inl
jtbandes: or like I did in ws-protocol, `ChannelId addChannel(const ChannelInfoWithoutId&)` 🤷

cpp/mcap/include/mcap/errors.hpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+inline std::error_code make_error_code(ErrorCode e) {
```
jtbandes, line 65: sounds like error_code is platform-dependent, maybe we should just always use error_condition?

cpp/mcap/include/mcap/mcap.hpp
jtbandes: If we're using c++17, maybe `std::byte*`?

cpp/mcap/include/mcap/errors.hpp
jhurliman: This is taken from the boilerplate example on how to implement custom error codes from the Boost team, so I'm a bit reticent to change it. I don't think it actually can/will be inlined because of the static local.

cpp/mcap/include/mcap/mcap.hpp
```diff
@@ -108,4 +127,78 @@ struct UnknownRecord {
   mcap::ByteArray data;
 };
 
+struct McapWriterOptions {
+  bool indexed;
```
jtbandes, line 131:
```suggestion
  bool indexed = false;
```
I generally put initializers on all primitives to help avoid accidental use of uninitialized data

cpp/mcap/include/mcap/errors.hpp
jhurliman: `error_condition` is used for matching`, `error_code` is used for storage/transport. APIs should return error codes and user code should match against those codes using categories.

cpp/mcap/include/mcap/mcap.hpp
```diff
@@ -108,4 +127,78 @@ struct UnknownRecord {
...
+  std::string library;
+  mcap::KeyValueMap metadata;
+
+  McapWriterOptions(const std::string_view profile)
+      : indexed(false)
+      , profile(profile)
+      , library("libmcap " LIBRARY_VERSION) {}
+};
+
+class McapWriter {
```
jtbandes, line 142:
```suggestion
class McapWriter final {
  McapWriter(const McapWriter&) = delete;
  McapWriter(McapWriter&&) = delete;
  McapWriter& operator=(const McapWriter&) = delete;
  McapWriter& operator=(McapWriter&&) = delete;
```

cpp/mcap/include/mcap/mcap.hpp
jhurliman: The initializer is a few lines down

cpp/mcap/include/mcap/mcap.hpp
```diff
@@ -108,4 +127,78 @@ struct UnknownRecord {
...
+   */
+  std::optional<std::error_code> write(const mcap::Message& message);
+
+  /**
+   * @brief Write an attachment to the output stream.
+   *
+   * @param attachment Attachment to add.
+   * @return An error code on failure.
+   */
+  std::optional<std::error_code> write(const mcap::Attachment& attachment);
```
jtbandes, line 181: error_code defines `operator bool`; is there any need to wrap it in `optional`?

cpp/mcap/include/mcap/mcap.hpp
jtbandes: I seem to remember some strangeness where it was possible to get uninitialized data even if there was a user defined constructor, but can't come up with it now... maybe the convention was established just to avoid forgetting to initialize members when changing the constructor definitions (which may be far away in a different file)

jtbandes: https://abseil.io/tips/61

> They will help reduce bugs from omissions, especially when someone adds a new constructor or a new data member.

cpp/mcap/include/mcap/errors.hpp
jtbandes: `inline` isn't about optimization, it's about not getting linker errors if this header is included in two object files that are later linked together.

PR #55, opened 2022-01-23
spec: remove crc from Channel Info, Message index, and Chunk Index

### LINKED ISSUE #43: Why does Chunk Index have a crc field when Attachment Index does not?
It seems like they should either both have it or neither have it. My personal leaning is toward only doing CRCs on user data (chunks, attachments) and not on index records.

defunctzombie:
Crc fields provide an additional layer of integrity checking on written data. Rather than add crc fields to every record, we provide crc fields in specific records to allow for opt-in crc checking.

If you want to opt-into crc validation for your messages, write messages into chunks (uncompressed chunks are valid chunks).

If you opt into index records, the footer contains a crc for all the written index data.

Attachment records contain a crc because they cannot appear in a chunk.

Fixes #43

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -128,7 +128,6 @@ Identifies a stream of messages on a particular topic and includes information a
 | 4 + N | schema_name | String | Schema Name | std_msgs/Header |
 | 4 + N | schema | uint32 length-prefixed bytes | Schema |  |
 | N | user_data | KeyValues<string, string> | Metadata about this channel | used to encode protocol-specific details like callerid, latching, QoS profiles... Refer to [supported profiles][profiles]. |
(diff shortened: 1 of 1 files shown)


defunctzombie:
While writing this up I did wonder whether it would be useful to provide a _CRC Record_ which would look something like the following:

```
start_offset - location in the file to start computing the crc
end_offset - location in the file to end computing the crc
crc - the crc value
```

The idea would be to provide a way for writers to crc when not using chunks. Even using chunks does present an overhead that might be undesirable compared to writing messages out directly. These CRC records would provide incremental crc snapshots.

jhurliman (approved):

jtbandes (approved):

Issue #24, opened 2022-01-15
profiles are underspecified


defunctzombie:
The _profile_ field in the [Header](https://github.com/foxglove/mcap/tree/main/docs/specification#header-op0x01) says it specifies "interpretation of channel info user data.". Does this mean that a file with protobuf encoding would still be a valid _ros1_ profile?

It would be useful to expand the scope of what a profile can describe to include any of the open-ended fields (i.e. encoding, schema, schema name, etc).

This would allow creating a _ros1_ profile that would indicate _ros1_ as the required encoding, .msg text as the required schema format, and the schema naming convention. By specifying all of the requirements within the profile, a library author can definitely say they implement support for a _ros1_ profile which will interoperate with other tooling producing _ros1_ profile files.

defunctzombie:
From discussion:

A profile could also document semantic concepts on how to reason about concepts like _tf_ frames. (i.e. in ROS1 a frame_id of `foo` is synonymous with `/foo`, tf frames only have a single parent). By documenting such semantics within the profile, tooling can make the decision on how to support the specific mcap file.

defunctzombie:
Rolled into my branch here: https://github.com/foxglove/mcap/tree/roman/spec-format/docs/specification

PR #57, opened 2022-01-23
spec: cleanup profiles

defunctzombie:
Profiles are for more than user data. They also specify values for other fields within the file. This change updates the language to allow for this and the ros1, and ros2 profiles to specify other required values.

Fixes: #24

### DIFF
--- a/docs/specification/profiles/README.md
+++ b/docs/specification/profiles/README.md
@@ -1,13 +1,6 @@
-## Supported profiles
+## Profiles
 
-[ros1]: ./ros1.md
-[ros2]: ./ros2.md
+Well-known MCAP profiles:
 
-This directory contains supported "profiles" for MCAP channel info user data. Usage of these profiles is not mandatory, but may be helpful to third party tooling in better understanding and displaying your data. For instance, an application that reads a "latching" key from a channel info record will not necessarily know what to do with the value - however if the reader knows the MCAP file is recorded with the "ros1" profile, it can make an inference that this is indicating a "latching topic" and behave accordingly.
-
-To make use of a profile, simply include the name of the profile in the "profile" field in the file header, and include the required keys in the user data section of all channel info records in the file. Additional keys can be added beyond those required by the profile as desired.
-
-Supported profiles are listed below:
-
-- [ros1][ros1]
-- [ros2][ros2]
+- [ros1](./ros1.md)
+- [ros2](./ros2.md)
--- a/docs/specification/profiles/ros1.md
+++ b/docs/specification/profiles/ros1.md
@@ -1,8 +1,19 @@
-## ROS1 Profile
+## ROS1
 
-Profile name: "ros1".
+Profile: **ros1**
 
-Channel info records serialized with the "ros1" profile may include the following fields:
+The `ros1` profile describes how to create mcap files for v1 of the [ROS](https://ros.org/) framework.
+
+### Channel Info
+
+| field       | value            |
+| ----------- | ---------------- |
+| encoding    | ros1             |
+| schema      | .MSG             |
+| schema_name | package/datatype |
+| user_data   | See below        |
+
+#### user_data
 
 - callerid (optional, string)
 - latching (optional, bool stringified as "true" or "false")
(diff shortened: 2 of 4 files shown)


docs/specification/profiles/ros2.md
```diff
@@ -1,7 +1,18 @@
...
-Channel info records serialized with the "ros1" profile may include the following fields:
+The `ros2` profile describes how to create mcap files for v2 of the [ROS](https://ros.org/) framework.
+
+### Channel Info
+
+| field       | value            |
+| ----------- | ---------------- |
+| encoding    | cdr              |
+| schema      | .MSG             |
+| schema_name | package/datatype |
```
jtbandes, line 13:
```suggestion
| encoding    | CDR              |
| schema      | [.msg](https://docs.ros.org/en/galactic/Concepts/About-ROS-Interfaces.html)             |
| schema_name | `package/subtype/datatype`, e.g. `my_msgs/msg/Thing` |
```

docs/specification/profiles/README.md
```diff
@@ -1,13 +1,6 @@
...
 
-[ros1]: ./ros1.md
-[ros2]: ./ros2.md
+Well-known MCAP profiles:
 
-This directory contains supported "profiles" for MCAP channel info user data. Usage of these profiles is not mandatory, but may be helpful to third party tooling in better understanding and displaying your data. For instance, an application that reads a "latching" key from a channel info record will not necessarily know what to do with the value - however if the reader knows the MCAP file is recorded with the "ros1" profile, it can make an inference that this is indicating a "latching topic" and behave accordingly.
-
-To make use of a profile, simply include the name of the profile in the "profile" field in the file header, and include the required keys in the user data section of all channel info records in the file. Additional keys can be added beyond those required by the profile as desired.
-
-Supported profiles are listed below:
```
jtbandes, line 10: why remove this?

defunctzombie: The relevant portion of this is stated in the spec document.

docs/specification/profiles/ros1.md
```diff
@@ -1,8 +1,19 @@
...
-Channel info records serialized with the "ros1" profile may include the following fields:
+The `ros1` profile describes how to create mcap files for v1 of the [ROS](https://ros.org/) framework.
+
+### Channel Info
+
+| field       | value            |
+| ----------- | ---------------- |
+| encoding    | ros1             |
+| schema      | .MSG             |
+| schema_name | package/datatype |
```
jtbandes, line 13:
```suggestion
| encoding    | [ROS 1 binary serialization](http://wiki.ros.org/msg)             |
| schema      | [.msg](http://wiki.ros.org/msg)             |
| schema_name | `package/datatype`, e.g. `my_msgs/Thing` |
```

defunctzombie: You want the encoding string to be `ROS 1 binary serialization` this seems rather verbose...

defunctzombie: These strings should also match https://github.com/foxglove/mcap/pull/60

jtbandes: Ok, I misunderstood this table because it's not clear if this column has exact values that must be used or just descriptions. If it is a literal value you're supposed to put in the encoding field then it should be a code snippet such as `ros1`

jtbandes:
I'm not sure I get the relationship between profile and encoding. Why do we define a `protobuf` encoding but not profile?

defunctzombie:
> I'm not sure I get the relationship between profile and encoding. Why do we define a `protobuf` encoding but not profile?

`encoding` was added to describe the message encoding for the channel. It doesn't tell you anything except that.

`profile` is where you specify what values for encoding are valid and the form of _schema_ and _schema_name.

If I don't specify a profile, and only specify `encoding` as `protobuf` then what does that mean for _schema_ and _schema_name_ fields? Right now that isn't documented or clear how it should be documented.

defunctzombie:
Rolled into my branch here: https://github.com/foxglove/mcap/tree/roman/spec-format/docs/specification

PR #61, opened 2022-01-24
Consolidate chunked and unchunked record types

wkalt:
Consolidates the chunked and unchunked docs in the specification, and
enables more flexible opt-in support for various record types.
Importantly, chunks can be used without message indexes, and unchunked
writers may omit all indexing if desired.

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -33,9 +33,10 @@ Some helpful terms to understand in the following sections are:
 - **Message**: A type of record representing a timestamped message on a channel (and therefore associated with a topic/schema). A message can be parsed by a reader that has also read the channel info for the channel on which the message appears.
 - **Chunk**: A record type that wraps a compressed set of channel info and message records.
 - **Attachment**: Extra data that may be included in the file, outside the chunks. Attachments may be quickly listed and accessed via an index at the end of the file.
-- **Index**: The format contains indexes for both messages and attachments. For messages, there are two levels of indexing - a **Chunk Index** at the end of the file points to chunks by offset, enabling fast location of chunks based on topic and timerange. A second index - the **Message Index** - after each chunk contains, for each channel in the chunk, and offset and timestamp for every message to allow fast location of messages within the uncompressed chunk data.
-
-  The attachment index at the end of the file allows for fast listing and location of attachments based on name, timestamp, or attachment type.
+- **Index**: The format contains indexes for both messages and attachments. For messages, there are two levels of indexing - a **Chunk Index** at the end of the file points to chunks by offset, enabling fast location of chunks based on channel and timerange. A second index - the **Message Index** - after each chunk contains, for each channel in the chunk, and offset and timestamp for every message to allow fast location of messages within the uncompressed chunk data. The attachment index at the end of the file allows for fast listing and location of attachments based on name, timestamp, or attachment type.
+- **Statistics**: A type of record at the end of the file, used to support fast summarization of file contents.
+- **Message Data Section**: Used in this doc to refer to the first portion of the file that contains chunks and message data. To be distinguished from the **Index Data Section**.
+- **Index Data Section**: The last part of the file, containing records used for searching and summarizing the file. The Index Data section is split into a **channel info portion**, **chunk index portion**, and **attachment index portion** each containing contiguous runs of the corresponding record type, followed by a **Statistics** record. All portions of the index data section are optional, subject to constraints and tradeoffs described below. There are no other record types in the index data section.
 
 ## Format Description
 
@@ -51,29 +52,13 @@ These are the magic bytes:
 
 The first record in every file must be a Header (op=0x01) and the last record must be a Footer (op=0x02).
 
-MCAP files may be **"chunked"** or **"unchunked"**. Chunked and unchunked files have different constraints on the layout of record types in the file. In chunked files, messages are grouped into optionally-compressed blocks of data before being written to disk. In an unchunked file, each message is written out uncompressed. See the diagrams below for clarity (the record types shown are described in the following section):
+MCAP files may contain a variety of record types. Specific constraints on valid usage of the record types is explained in the sections below, but in general record types may be used or not depending on the feature requirements of the consumer.
 
-#### Chunked
+The diagrams below show two possible variants - a file that is chunked and indexed, i.e making full use of the features, and one that is unchunked but contains statistics.
 
 ![Chunked][diagram chunked]
 
-#### Unchunked
-
 ![Unchunked][diagram unchunked]
-
-Benefits of chunked files include:
-
-- Support for random access via time- and topic-based indexing.
-- Reduced storage requirements when recording or processing data.
-- Reduced bandwidth requirements when transferring over a network.
-- Possibly higher write performance if the cost of IO outweighs the cost of compression.
-
-Benefits of unchunked files include:
-
-- Higher write performance on CPU-constrained systems.
-- Less potential for data los in case of a recording crash. No "to-be-compressed" buffer is dropped by the recorder -- though the protocol makes no specification on how the process syncs unchunked messages to disk.
-
-Unchunked files are less friendly to readers than chunked files due to their lack of an index and greater size. When unchunked files are in use, they may be converted to chunked files in post-processing to mitigate this.
 
 ### Record Types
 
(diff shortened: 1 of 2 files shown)


docs/specification/README.md
```diff
@@ -33,9 +33,10 @@ Some helpful terms to understand in the following sections are:
...
-
-  The attachment index at the end of the file allows for fast listing and location of attachments based on name, timestamp, or attachment type.
+- **Index**: The format contains indexes for both messages and attachments. For messages, there are two levels of indexing - a **Chunk Index** at the end of the file points to chunks by offset, enabling fast location of chunks based on topic and timerange. A second index - the **Message Index** - after each chunk contains, for each channel in the chunk, and offset and timestamp for every message to allow fast location of messages within the uncompressed chunk data. The attachment index at the end of the file allows for fast listing and location of attachments based on name, timestamp, or attachment type.
```
defunctzombie, line 36:
```suggestion
- **Index**: The format contains indexes for both messages and attachments. For messages, there are two levels of indexing - a **Chunk Index** at the end of the file points to chunks by offset, enabling fast location of chunks based on channel and timerange. A second index - the **Message Index** - after each chunk contains, for each channel in the chunk, and offset and timestamp for every message to allow fast location of messages within the uncompressed chunk data. The attachment index at the end of the file allows for fast listing and location of attachments based on name, timestamp, or attachment type.
```

docs/specification/README.md
```diff
@@ -33,9 +33,10 @@ Some helpful terms to understand in the following sections are:
...
+- **Statistics**: A type of record at the end of the file, used to support fast summarization of file contents.
```
defunctzombie, line 37: Why does statistics record get a call-out here but other records don't? Actually looking at all these terms it seems like there is some overlap between defining records and non-record concepts.

docs/specification/README.md
```diff
@@ -51,30 +52,19 @@ These are the magic bytes:
 
 The first record in every file must be a Header (op=0x01) and the last record must be a Footer (op=0x02).
 
-MCAP files may be **"chunked"** or **"unchunked"**. Chunked and unchunked files have different constraints on the layout of record types in the file. In chunked files, messages are grouped into optionally-compressed blocks of data before being written to disk. In an unchunked file, each message is written out uncompressed. See the diagrams below for clarity (the record types shown are described in the following section):
+MCAP files may contain a variety of record types. Specific constraints on valid usage of the record types is explained in the sections below, but in general record types may be used or not depending on the feature requirements of the consumer:
```
defunctzombie, line 55:
```suggestion
MCAP files are composed of a sequence of records. Writers can use records appropriate for their feature requirements. Different records and sequences of records allow for making tradeoffs between write performance, write durability, read performance, random access, and more. 
```

docs/specification/README.md
```diff
@@ -51,30 +52,19 @@ These are the magic bytes:
...
+- If the consumer requires fast random access to message data in the file based on topic and timestamp, they will want to use Message Index records after their Chunks, and ensure the index data section includes Channel Info and Chunk Index records.
+- If the consumer requires fast summarization of file contents, they should include a Statistics record in their index data section.
+- If the consumer wants to ensure as much data is written as possible in case of crash, they should use Message records at the top level, instead of Chunks.
+- If the consumer needs to save space on disk, or is constrained on disk or network bandwidth, they may benefit from using Chunks so that messages are compressed. This comes at the risk of losing buffered to-be-compressed messages in the case of a crash.
 
-![Chunked][diagram chunked]
+The diagrams below show two possible variants - a file that is chunked and indexed, i.e making full use of the features, and one that is unchunked but contains statistics.
 
-#### Unchunked
+![Chunked][diagram chunked]
```
defunctzombie, line 64: Maybe we should expand on the diagrams and various record combinations in an appendix section of the spec rather than at the start here.

docs/specification/README.md
```diff
@@ -51,30 +52,19 @@ These are the magic bytes:
...
+MCAP files may contain a variety of record types. Specific constraints on valid usage of the record types is explained in the sections below, but in general record types may be used or not depending on the feature requirements of the consumer:
 
-#### Chunked
+- If the consumer requires fast random access to message data in the file based on topic and timestamp, they will want to use Message Index records after their Chunks, and ensure the index data section includes Channel Info and Chunk Index records.
+- If the consumer requires fast summarization of file contents, they should include a Statistics record in their index data section.
+- If the consumer wants to ensure as much data is written as possible in case of crash, they should use Message records at the top level, instead of Chunks.
```
defunctzombie, line 59: These feel too prescriptive (maybe not the right word). The feeling I walk away with when reading some of these is "ok why". If I want to ensure as much data is written in case of a crash why should I not use chunks? Doesn't that depend on how I've implemented my writer? I can definitely implement a writer that is both crash-data-loss-resilient and uses chunks.

docs/specification/README.md
```diff
@@ -156,7 +144,7 @@ A Chunk is a collection of compressed channel info and message records.
 
 #### Message Index (op=0x06)
 
-The Message Index record maps timestamps to message offsets. One message index record is written for each channel in the preceding chunk. All message index records for a chunk must immediately follow the chunk.
+The Message Index record maps timestamps to message offsets. If message indexing is in use, following each chunk, a message index record is written for each channel in the chunk preceding. All message index records for a chunk must immediately follow the chunk in a contiguous run of records.
```
defunctzombie, line 147: Did we decide whether MessageIndex is optional?

docs/specification/README.md
```diff
@@ -181,7 +169,7 @@ The Chunk Index records form a coarse index of timestamps to chunk offsets, alon
 
 #### Attachment (op=0x08)
 
-Attachments can be used to attach artifacts such as calibration data, text, or core dumps. Attachment records must not appear within a chunk.
+Attachments can be used to attach artifacts such as calibration data, text, or core dumps. Attachment records must not appear within a chunk, or inside the index data section.
```
defunctzombie, line 172: Why do we call out the index data section for attachment records? We don't do that for Message records.

docs/specification/README.md
```diff
@@ -205,7 +193,7 @@ The attachment index is an index to named attachments within the file. One recor
 
 #### Statistics (op=0x0A)
 
-The statistics record contains statistics about the recorded data. It is the last record in the file before the footer.
+The statistics record contains statistics about the recorded data. It is the last record in the file before the footer. The record must be preceded in the index data section by Channel Info records for any channels referenced in the `channel_message_counts` field. If this is undesirable but some statistics are still desired, the field may be set to a zero-length map. The statistics record is optional.
```
defunctzombie, line 196: It's not strictly required that statistics must be preceded by Channel Info. Yes statistics might reference channel ids but doesn't really care where they are found in the file?

docs/specification/notes/explanatory-notes.md
```diff
@@ -4,7 +4,7 @@ The following notes may be useful for users of the MCAP format, including implem
 
 ## Feature Explanations
 
-The format is intended to support efficient, indexed reading of messages and generation of summary data in both local and remote contexts. "Seeking" should be imagined to incur either a disk seek or an HTTP range request to an object store -- the latter being significantly more costly.
+The format is intended to support efficient, indexed reading of messages and generation of summary data in both local and remote contexts. "Seeking" should be imagined to incur either a disk seek or an HTTP range request to an object store -- the latter being significantly more costly. In both random access and summarization, features may be unsupported due to choices taken by the writer of the file. For instance, statistics may not include channel message counts, or there may be no message index present. If the index data section is empty, the index_offset in the file footer will be set to zero.
```
defunctzombie, line 7: I don't really follow who the target audience is for this. Some of the comments here seem like they belong in the spec:

```
If the index data section is empty, the index_offset in the file footer will be set to zero.
```

docs/specification/README.md
wkalt: I called it out because I had to reference it elsewhere in the section

docs/specification/README.md
wkalt: yeah - implied I think by "If message indexing is in use"? What do you think?

docs/specification/README.md
wkalt: if the channel infos don't precede the statistics record, there's no value in setting the field nonempty IMO, since you would need to traverse the entire file to get the channel infos and might as well recompute the counts then. So I think better to require it if not having it wouldn't make sense.

docs/specification/notes/explanatory-notes.md
wkalt: that's in the spec, in the index_offset field description.

docs/specification/README.md
wkalt: it's really topic, right? Because the user will supply a topic, not a channel ID as a lookup key.

wkalt: I guess it's the channel in the chunk index record itself, and that location of chunks is enabled by both channel and topic.

wkalt: updated

defunctzombie (approved):

PR #66, opened 2022-01-25
Clarify that channel_message_counts is optional, update variable-length field byte counts to "4 + N"

jhurliman:


defunctzombie:
I wonder if we are doing something weird with the whole 4+N thing. On one hand we've defined the types to have length prefixes so the `N` could be read to include the prefix size?

jhurliman:
@defunctzombie the `4 + N` was an additional clarifier for me that these are length prefixed (and the length prefix is uint32), and I don't think there's a downside to this additional clarity.

defunctzombie:
The length field updates I've included in this branch: https://github.com/foxglove/mcap/tree/roman/spec-format/docs/specification

I don't know about the optional channel counts in statistics. If you set it to 0, should a reader treat that as an indication that there are no messages in the file? Other records where we allow 0 (i.e. Footer with offsets), 0 means there are no records of that type. A 0 here doesn't actually mean there are no channels in the file - just that the statistics record lied to you.

I would be in favor of a separate statistics record for these counts if we want to make this actually optional. Or a field indicating whether this count is used.

defunctzombie:
Closing - I've incorporated the changes to the latest working copy.

PR #70, opened 2022-01-27
Update typescript reader and writer to v0 candidate

defunctzombie:


typescript/src/v0/constants.ts
```diff
@@ -2,20 +2,24 @@
...
-  MESSAGE = 0x04,
-  CHUNK = 0x05,
-  MESSAGE_INDEX = 0x06,
-  CHUNK_INDEX = 0x07,
-  ATTACHMENT = 0x08,
-  ATTACHMENT_INDEX = 0x09,
-  STATISTICS = 0x0a,
+  MIN = 0x00,
+  HEADER = 0x00,
+  FOOTER = 0x7f,
```
jtbandes, line 7: shouldn't be out of order. but I proposed keeping footer=0x02 on the other PR

PR #71, opened 2022-01-27
Add `doctor` command to mcap CLI

defunctzombie:
Introduce the `doctor` command to the mcap cli.


go/mcap/cmd/doctor.go
```diff
@@ -0,0 +1,178 @@
...
+type mcapRecordReader struct {
+	reader       io.Reader
+}
+
+type RawRecord struct {
+	OpCode libmcap.OpCode
+	Bytes []byte
+}
+
+func (this *mcapRecordReader) Warn(format string, v ...interface{}) {
```
wkalt, line 24: "this" isn't typically used in the same way in go as in js. It would be more idiomatic to call this by a short initial, like `r`, or `rr`.

Also, in terms of structuring the code, I think we will probably want to pull the logging functionality into a separate logging module (next to cmd/) so that all the other pieces can make use of the colors and such - so probably wouldn't attach these methods to mcapRecordReader, but write some `func Errorf(string, ...interface{})`, `func Warnf(string, ...interface{})`, etc with whatever colors we want to attach.


Code would then use those like this:

`import (
log "github.com/foxglove/mcap/go/mcap/logging"
)

log.Warnf("Hello this is red")
```

File: go/mcap/cmd/doctor.go [hunk 2]
```diff
@@ -0,0 +1,178 @@
...
+func (this *mcapRecordReader) Fatalf(format string, v ...interface{}) {
+	color.Red(format, v...)
+	os.Exit(1)
+}
+
+
+func (this *mcapRecordReader) ConsumeMagic() {
+	magicBytes := make([]byte, len(libmcap.Magic))
+
+	bytesRead, err := this.reader.Read(magicBytes)
```
wkalt, line 48: It seems like you're handling it here, but one of the subtleties it's important to be aware of is that calling Read on a buffer does not guarantee that the buffer is filled - it might take multiple Read calls to fill a buffer, and depending on the reader or what is backing the reader there can be numerous reasons for this. You can address this by ensuring the bytes read is expected as you're doing here, or another alternative that will do the length check for you, is this:

```
n, err := io.ReadFull(this.reader, magicBytes)
```
which will perform multiple Read calls as necessary to fill the buffer, and return an error if fewer than len(magicBytes) can be read.

go/mcap/cmd/doctor.go
```diff
@@ -0,0 +1,178 @@
...
+
+	opcode := libmcap.OpCode(opcodeAndLength[0])
+	recordLen := binary.LittleEndian.Uint64(opcodeAndLength[1:9])
+
+	recordBytes := make([]byte, recordLen)
+	bytesRead, err = this.reader.Read(recordBytes)
+	if err != nil {
+		this.Fatal(err)
+	}
+	if (uint64(bytesRead) != recordLen) {
```
wkalt, line 79: the parentheses around the inequality are unnecessary - I'll get some linting in here today and then this will be broken. I think gofmt will remove these, which should run if you install the go vscode plugin. Maybe not though.

go/mcap/cmd/doctor.go
```diff
@@ -0,0 +1,178 @@
...
+
+		if (previousOpcode == nil && rawRecord.OpCode != libmcap.OpHeader) {
+			recordReader.Error("First record MUST be a header")
+		}
+		
+		switch rawRecord.OpCode {
+		case libmcap.OpHeader:
+			// warn if header doesn't contain a library field - this is encouraged
+			// error if header field has a profile that is not well known and is not prefixed by "x-"
+			break
```
wkalt, line 126: break in go is not required to break out of a switch. This will actually break out of the for loop. The breaks should just be removed; the code will exit after it finds a case, unless explicitly told to `fallthrough`.

defunctzombie:
@wkalt updated to latest go code - thoughts on what we'd like to see done before merging the first version of doctor?

wkalt (approved):

PR #75, opened 2022-01-28
Add a data end record

defunctzombie:
The data end record indicates when the data section ends. This lets streaming readers know they have reached the end of the data section.


wkalt:
@defunctzombie I think data end is probably the better choice, because it will enable a CRC checksum on a file that otherwise has no summary section. The record type itself can be optional as well - but that probably does imply some indicator in the header that the reader should pay attention to the checksum.

jhurliman (approved):

PR #77, opened 2022-01-28
Add support for conversion of ros2 db3 bagfiles

wkalt:
Adds support to the mcap CLI tool for converting ros2 db3 files to mcap.
This should work without arguments on a system where AMENT_PREFIX_PATH
points to /opt/ros/<distro>, which seems to be the default ROS install.
Otherwise the equivalent search directories must be supplied via
parameter.


defunctzombie:
I don't think these tools should be in `libmcap`. Maybe a separate folder in the go folder for such conversion tools (similar to mcap folder for info, etc).

defunctzombie:
The thinking is that libmcap shouldn't have to depend on more and more things as we add encoding and schema formats.

go/libmcap/db3_to_mcap.go
```diff
@@ -0,0 +1,210 @@
+package libmcap
```
defunctzombie, line 1: nit: probably ros2db3_to_mcap? db3 is a generic extension for any sqlite file?

.gitattributes
```diff
@@ -4,3 +4,4 @@ testdata/ filter=lfs diff=lfs merge=lfs -text
 testdata/bags filter=lfs diff=lfs merge=lfs -text
 testdata/bags/demo.bag filter=lfs diff=lfs merge=lfs -text
 testdata/mcap filter=lfs diff=lfs merge=lfs -text
+testdata/db3/chatter.db3 filter=lfs diff=lfs merge=lfs -text
```
defunctzombie, line 7: I think you can use a ** here to make anything in this folder gitlfs. Thats a safer bet than forgetting to add it to lfs.

PR #79, opened 2022-01-29
Add conformance test scaffold

jtbandes:
TODO:
- [x] move code into separate files
- [x] move to a separate package under `tests/` or `testdata/`
- [x] choose expected output format (maybe not json)
- [ ] write more detailed files and figure out file bucketing
- [ ] add runners for other languages


tests/conformance/data/NoData-pad-st-sum.expected.txt
```diff
@@ -0,0 +1,6 @@
+{"type":"Header","profile":"","library":""}
+{"type":"Statistics","messageCount":"BigInt(0)","channelCount":0,"attachmentCount":0,"chunkCount":0,"channelMessageCounts":{}}
+{"type":"SummaryOffset","groupOpcode":12,"groupStart":"BigInt(64)","groupLength":"BigInt(0)"}
```
defunctzombie, line 3: If we stick with JSON - maybe quoting the number values is enough (no need for BigInt wrapper). We can generate data with values in the expected numeric range to validate that each reader implementation is able to de-serialize.

.gitattributes
```diff
@@ -1,3 +1,4 @@
 *.png filter=lfs diff=lfs merge=lfs -text
+*.mcap filter=lfs diff=lfs merge=lfs -text
 testdata/mcap/demo.mcap filter=lfs diff=lfs merge=lfs -text
 testdata/ filter=lfs diff=lfs merge=lfs -text
```
defunctzombie, line 4: Are we planning to keep `testdata` its confusing to have both `tests/` and `testdata/` at the top level.

jtbandes: I agree and think we should remove it, just haven't looked into what it is used for right now.

Issue #82, opened 2022-01-31
Rename record_time to log_time


defunctzombie:
From team discussion - record_time is ambiguous. `log_time` was considered more appropriate

PR #84, opened 2022-01-31
spec: rename record_time to log_time

defunctzombie:
After discussion, we've formed a consensus that `log_time` is a better term for
the semantics of this field. The `log_time` is the time any writer wants to indicate
a message is logged at.

Fixes: #82


defunctzombie:
non-binary-breaking wording only change

foxymiles (approved):
👍

jtbandes (approved):

jhurliman (approved):

PR #91, opened 2022-02-02
Minor allocation reductions for go bag transcoder

wkalt:
Adds a basic benchmark test for the mcap to bag conversion, and reuses a
message buffer across messages when transcoding.


jhurliman:
BenchmarkBag2MCAP

PR #102, opened 2022-02-03
Add Schema record

defunctzombie:


docs/specification/README.md
```diff
@@ -146,9 +150,7 @@ Channel Info records are uniquely identified within a file by their channel ID.
 | 2 | id | uint16 | A unique identifier for this channel within the file. |
 | 4 + N | topic | String | The channel topic. |
 | 4 + N | message_encoding | String | Encoding for messages on this channel. The value should be one of the [well-known message encodings](./well-known-encodings.md). Custom values should use `x-` prefix. |
-| 4 + N | schema_encoding | String | Format for the schema. The value should be one of the [well-known schema formats](./well-known-schema-formats.md). Custom values should use the `x-` prefix. |
-| 4 + N | schema | uint32 lengh prefixed Bytes | Schema should conform to the schema_encoding. |
-| 4 + N | schema_name | String | An identifier for the schema. The schema name should conform to any schema_encoding requirements. |
+| 2 | datatype | uint16 | The datatype for messages on this channel. |
```
jhurliman, line 153: This should be `datatype_id` to be consistent with `channel_id`

docs/specification/README.md
```diff
@@ -440,17 +461,21 @@ A writer may choose to put messages in Chunks to compress record data. This MCAP
 
 ```
 [Header]
+[Datatype]
```
jhurliman, line 464: The other message examples use ids. It will get a bit confusing to use another sequence of ids for Datatype, but I think it's even more confusing to leave them out.

jtbandes (changes requested):
The Datatype record contains schema_encoding, schema, and schema_name. It is clearly a Schema record. 😄

docs/specification/README.md
```diff
@@ -294,12 +296,27 @@ A Summary Offset record contains the location of records within the summary sect
...
+
+A Datatype record defines an individual datatype.
+
+Datatype records are uniquely identified within a file by their datatype ID. A Datatye record must occur at least once in the file prior to any Channel Info referring to its datatype ID.
+
+| Bytes | Name | Type | Description |
+| --- | --- | --- | --- |
+| 2 | id | uint16 | A unique identifier for this datatype within the file. |
+| 4 + N | schema_encoding | String | Format for the schema. The value should be one of the [well-known schema formats](./well-known-schema-formats.md). Custom values should use the `x-` prefix. |
+| 4 + N | schema | uint32 lengh prefixed Bytes | Schema should conform to the schema_encoding. |
```
wkalt, line 315: length-prefixed

docs/specification/README.md
defunctzombie: updated

docs/specification/README.md
defunctzombie: Added some identifiers here A, B

docs/specification/README.md
defunctzombie: fixed

defunctzombie:
@jtbandes renamed to schema record

docs/specification/README.md
```diff
@@ -28,6 +28,7 @@ MCAP files are designed to work well under various workloads, resource constrain
   - [Metadata Index](#metadata-op0x0C)
   - [Summary Offset](#summary-offset-op0x0D)
   - [Data End](#data-end-op0x0E)
+  - [Schema](#schema-op0x0F)
```
amacneil, line 31: move up and reorder opcodes?

defunctzombie: yea - we will reorder the opcodes once we are happy with the records

docs/specification/README.md
```diff
@@ -294,12 +297,27 @@ A Summary Offset record contains the location of records within the summary sect
...
+A Schema record defines an individual schema.
+
+Schema records are uniquely identified within a file by their schema ID. A Datatye record must occur at least once in the file prior to any Channel Info referring to its schema ID.
+
+| Bytes | Name | Type | Description |
+| --- | --- | --- | --- |
+| 2 | id | uint16 | A unique identifier for this schema within the file. |
+| 4 + N | encoding | String | Format for the schema. The value should be one of the [well-known schema formats](./well-known-schema-formats.md). Custom values should use the `x-` prefix. |
+| 4 + N | schema | uint32 length-prefixed Bytes | Schema should conform to the encoding. |
+| 4 + N | name | String | An identifier for the schema. The schema name should conform to any encoding requirements. |
```
amacneil, line 317: move `name` up to the top (under `id`), and rename `schema` to `data` for consistency with other messages?

defunctzombie:
I've let this bake overnight and still feel like the more natural place for `message_encoding` is alongside the fields in the Schema record. Here's my thinking:

When an end user reader encounters a Message record, it will want to deserialize the message `data` into some object appropriate for language/environment. How's does it do that? It needs to know the message encoding and the schema. Those two bits of information describe how to decode a message. These two things are naturally related and should occur together to simplify reasoning about how one impacts the other.

Some questions:

- Does protobuf message encoding imply a specific schema format? Its hard to imagine general purpose mcap readers where this isn't true? Maybe by mcap design its on the end-reader to enforce the pairing of message encoding+schema that the reader supports. In this case, these fields are still used together.
- For ros1 message encoding, readers would "typically" need schemas in the .MSG format to decode the serialized messages.
- CBOR and JSON message encoding are self-contained. They can be decoded without any schema - tho mcap still encourages (requires) a schema to aid in tooling making sense of the file. If you wanted to write cbor or json without a schema there other options.
- For ros2 systems, the message encoding default is CDR and the schemas are in .MSG format. ROS2 allows for different message encodings, but its not so clear how this influences (or is influenced by the schema format used). i.e. if you wanted to use protobuf encoding with ros2 would you use .msg or re-write all your .msg files into .proto files?

After writing out the above, I'm starting to think `schema_encoding` actually dictates the range of allowed (expected?) message encodings. For example:

- schema encoding `jsonschema` allows for message encoding of `cbor` and `json`
- schema encoding `proto` allows for message encoding of `proto`
- schema encoding `ros1msg` allows for message encoding of `ros1`
- schema encoding `ros2msg` allows for message encoding of `cdr`

Would this prohibit using a `ros1msg` schema encoding and using `proto` or `cbor` message encoding? We don't have to "prohibit" it - but there's also some limit to how readers and writers are expected to inter-operate. Most message serialization formats I'm familiar with (which have an IDL) have a 1:1 relationship with their IDL. The tooling to decode uses the IDL to understand how to decode. `cbor` and `json` have no IDL, so we've flipped around the relationship of the schema encoding and message encoding - saying that `jsonschema` is a way of indicating `cbor` or `json` encoded messages. And for ros2, which claims to support many different encodings, today is practically .msg schema files with CDR encoded messages.

amacneil:
> After writing out the above, I'm starting to think schema_encoding actually dictates the range of allowed (expected?) message encodings.

The relationship of `encoding` to `schema` is many to many: `jsonschema` and `typeschema` can both be used with `cbor` and `json` encoding respectively. Not all combinations are allowed, but I would argue the encoding defines supported schema types, not the other way around.

I think putting `encoding` inside `Schema` record is more confusing semantically than putting it inside `ChannelInfo`, especially when you consider that some encodings (cbor/bson/json) do not require a schema.

Encoding is a property of messages, and messages are contained within channels. Some encodings (ros1/cdr/proto) cannot be decoded without the schema, so we allow the channel to reference a Schema record (extracted to a dedicated record type only because of space savings). Other encodings (cbor/bson/json) do not require a schema, and I would argue that we should allow `schema_id=0` (or a `Schema` with a `name` and no `data`) in that case to indicate no schema available.

Semantically, it is more sensible to ask "what is the encoding of this `ChannelInfo`" than to ask "what is the encoding of this `Schema`".

docs/specification/README.md
defunctzombie: Name talks about needing to conform to the encoding - is it weird it comes before encoding?

PR #110, opened 2022-02-04
Move name earlier in Schema record

amacneil:
**Public-Facing Changes**
- Schema `name` appears earlier in the record, for aesthetic purposes.
- A few other strings in the spec are updated (non-breaking).

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -149,9 +149,9 @@ Schema records are uniquely identified within a file by their schema ID. A Schem
 | Bytes | Name | Type | Description |
 | --- | --- | --- | --- |
 | 2 | id | uint16 | A unique identifier for this schema within the file. |
+| 4 + N | name | String | An identifier for the schema. |
 | 4 + N | encoding | String | Format for the schema. The value should be one of the [well-known schema formats](./well-known-schema-formats.md). Custom values should use the `x-` prefix. |
-| 4 + N | schema | uint32 length-prefixed Bytes | Schema should conform to the encoding. |
-| 4 + N | name | String | An identifier for the schema. The schema name should conform to any encoding requirements. |
+| 4 + N | data | uint32 length-prefixed Bytes | Must conform to the schema encoding. |
 
 Schema records may be duplicated in the summary section.
 
@@ -166,7 +166,7 @@ Channel Info records are uniquely identified within a file by their channel ID. 
 | 2 | id | uint16 | A unique identifier for this channel within the file. |
 | 4 + N | topic | String | The channel topic. |
 | 4 + N | message_encoding | String | Encoding for messages on this channel. The value should be one of the [well-known message encodings](./well-known-encodings.md). Custom values should use `x-` prefix. |
-| 2 | schema_id | uint16 | The schema for messages on this channel. |
+| 2 | schema_id | uint16 | The schema for messages on this channel. Can be `0` if schema is not required for this message_encoding. |
 | 4 + N | metadata | Map<string, string> | Metadata about this channel |
 
 Channel Info records may be duplicated in the summary section.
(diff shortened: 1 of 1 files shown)


docs/specification/README.md
```diff
@@ -166,7 +166,7 @@ Channel Info records are uniquely identified within a file by their channel ID.
 | 2 | id | uint16 | A unique identifier for this channel within the file. |
 | 4 + N | topic | String | The channel topic. |
 | 4 + N | message_encoding | String | Encoding for messages on this channel. The value should be one of the [well-known message encodings](./well-known-encodings.md). Custom values should use `x-` prefix. |
-| 2 | schema_id | uint16 | The schema for messages on this channel. |
```
amacneil, line 169: Is this contentious? Seems helpful for `json` or `h264` encoded channels.

docs/specification/README.md
```diff
@@ -149,9 +149,9 @@ Schema records are uniquely identified within a file by their schema ID. A Schem
 | Bytes | Name | Type | Description |
 | --- | --- | --- | --- |
 | 2 | id | uint16 | A unique identifier for this schema within the file. |
+| 4 + N | name | String | An identifier for the schema. |
 | 4 + N | encoding | String | Format for the schema. The value should be one of the [well-known schema formats](./well-known-schema-formats.md). Custom values should use the `x-` prefix. |
-| 4 + N | schema | uint32 length-prefixed Bytes | Schema should conform to the encoding. |
```
amacneil, line 153: renamed to `Schema.data` so that we don't have `Schema.schema`

docs/specification/README.md
```diff
@@ -183,7 +183,7 @@ The message encoding and schema must match that of the channel info record corre
 | 4 | sequence | uint32 | Optional message counter assigned by publisher. If not assigned by publisher, must be recorded by the recorder. |
 | 8 | publish_time | Timestamp | Time at which the message was published. If not available, must be set to the log time. |
 | 8 | log_time | Timestamp | Time at which the message was recorded. |
-| N | message_data | Bytes | Message data, to be decoded according to the schema of the channel. |
```
amacneil, line 186: Renamed to `Message.data` to avoid stuttering in `Message.message_data`

docs/specification/README.md
```diff
@@ -166,7 +166,7 @@ Channel Info records are uniquely identified within a file by their channel ID.
 | 2 | id | uint16 | A unique identifier for this channel within the file. |
 | 4 + N | topic | String | The channel topic. |
 | 4 + N | message_encoding | String | Encoding for messages on this channel. The value should be one of the [well-known message encodings](./well-known-encodings.md). Custom values should use `x-` prefix. |
```
amacneil, line 168: Keeping this as `ChannelInfo.message_encoding` rather than `ChannelInfo.encoding` because it is a property of the message, not the channel (and to avoid confusion with `Schema.encoding`).

jtbandes (approved):

docs/specification/README.md
defunctzombie: It's nice to provide a name for the datatype even if you don't need a schema. I don't think this change belongs in this PR.

amacneil: Nice if you’re using ros 2 and just don’t have the schema data. But for json or h264 why do I have to make up a schema name? That’s silly.

I can split PR if you preferred.

docs/specification/README.md
```diff
@@ -166,7 +166,7 @@ Channel Info records are uniquely identified within a file by their channel ID.
...
-| 2 | schema_id | uint16 | The schema for messages on this channel. |
+| 2 | schema_id | uint16 | The schema for messages on this channel. Can be `0` if schema is not required for this message_encoding. |
```
defunctzombie, line 169:
```suggestion
| 2 | schema_id | uint16 | The schema for messages on this channel. |
```

defunctzombie (approved):

docs/specification/README.md
defunctzombie: If you are using ros2 and don't have schema data mcap is not for you. I think we are getting a bit far from the plot here. Mcap is about storing schematized messages. If you want to record opaque bytes is mcap really the format for that?

amacneil: I didn't say anything about opaque bytes. If I add an h264 NAL stream to an mcap file (a use case we have previously discussed as valid) there is no schema or schema name, only a message encoding. As far as I can tell the only solutions I can see to this are:

- h264 or json streams with no schema are not use cases we want to support
- Create an empty Schema with no name, schema encoding, or data
- Alllow a "null" (zero byte) schema_id

amacneil: > If you are using ros2 and don't have schema data mcap is not for you.

If we want to support converting .db3 to .mcap, there are absolutely going to be some files generated that have a schema name but no schema data. However, that's unrelated to this change because you would still need a Schema record in that instance to save the schema name. I'm pushing the discussion above about what to do if you have no schema name either, just a stream of self-describing encoded data.

PR #111, opened 2022-02-04
Add length prefix to Chunk `records` field

jtbandes:
**Public-Facing Changes**
Add 8 bytes to the Chunk record: a length prefix on the `records` bytes.

**Description**
Schema, Message, Chunk, and Attachment records all have a `Bytes` field. Of these, Schema and Attachment had a length prefix on their bytes while Message and Chunk did not.

A length prefix allows the record to be extended safely in future spec versions. We do not want to pay the extra cost for adding a length prefix to Message since the record will be so frequent, but since there will normally be far fewer chunks than messages, there doesn't seem to be a strong reason to prevent the Chunk record from being extensible in the future.


jhurliman (dismissed):
:(

defunctzombie:
After discussing we think this is a small price to pay for the v0 to avoid introducing new chunk records if we find something we want to add to the end.

defunctzombie (approved):

PR #115, opened 2022-02-07
Convert conformance output to json.

foxymiles:
**Public-Facing Changes**
This switches the format of the conformance tests outputs to json. This will make it easier to reuse them for writer conformance tests.

**Description**
This just switches from our custom text format to JSON.


tests/conformance/scripts/run-tests/runners/index.ts
```diff
@@ -6,9 +6,9 @@ import TypescriptStreamedTestRunner from "./TypescriptStreamedTestRunner";
 export type { ITestRunner };
 
 const runners: readonly ITestRunner[] = [
-  new TypescriptStreamedTestRunner(),
-  new TypescriptIndexedTestRunner(),
   new PythonStreamedTestRunner(),
+  new TypescriptIndexedTestRunner(),
+  new TypescriptStreamedTestRunner(),
```
jtbandes: we have two streamed runners now, can remove one

tests/conformance/scripts/run-tests/index.ts
```diff
@@ -69,9 +73,11 @@ async function main(options: {
           hadError = true;
           continue;
         }
-        if (output !== expectedOutput) {
+        const outputNorm = normalizeJson(output);
+        const expectedNorm = normalizeJson(expectedOutput);
```
jtbandes: expected should already be normalized, right?

tests/conformance/data/OneMessage/OneMessage.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  {
+    "type": "Footer",
+    "fields": [
+      ["summary_crc", 0],
+      ["summary_offset_start", 0],
+      ["summary_start", 0]
+    ]
+  }
+]
```
jtbandes: can we add a trailing newline to make git/github/editors happy?

tests/conformance/scripts/run-tests/runners/stringifyRecords.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (value instanceof Uint8Array) {
+    return Array.from(value);
+  }
+
+  if (value instanceof Map) {
+    return Object.fromEntries(value);
+  }
+
+  if (typeof value === "bigint") {
+    return Number(value);
```
jtbandes: loss of precision – I guess not a problem today since none of our tests use large uint64 values? but maybe we should have some that do?

typescript/package.json
```diff
@@ -48,6 +49,8 @@
     "eslint-plugin-jest": "25.2.4",
     "eslint-plugin-prettier": "4.0.0",
     "jest": "27.3.1",
+    "json-diff": "^0.7.1",
+    "json-stringify-pretty-compact": "^3.0.0",
```
jtbandes: These shouldn't be added here; there is a separate package.json for the conformance tests. And `lodash` should be added there too because I don't think it's currently listed as an explicit dependency.

jtbandes (approved):

tests/conformance/scripts/run-tests/runners/stringifyRecords.ts
jtbandes: I think we may want to use strings for uint64 to make JS's life easier. Otherwise I guess we would have to pull in an external json serialization lib that can handle bigints.

tests/conformance/scripts/run-tests/index.ts
foxymiles: Yeah although maybe we should leave this here in case we decide to rebuild the output from another runner than ts? Maybe it's not needed.

tests/conformance/scripts/run-tests/runners/index.ts
foxymiles: Oops thanks for catching that.

tests/conformance/scripts/run-tests/runners/stringifyRecords.ts
foxymiles: Yeah I think you're right. I'll change it to a string.

PR #116, opened 2022-02-07
typescript: fix Schema record

jtbandes:
Schema data was being treated as a string instead of bytes in TS code. Also rename the Schema fields and Message.data to match the spec.


defunctzombie:
The spec says schema data should be uint32 prefixed:

https://github.com/foxglove/mcap/tree/main/docs/specification#schema-op0x03

jtbandes:
Oops, will fix. Still needs to be a Uint8Array and not string.

typescript/src/v0/Mcap0RecordBuilder.ts
```diff
@@ -92,9 +92,10 @@ export class Mcap0RecordBuilder {
     this.bufferBuilder
       .uint64(0n) // placeholder
       .uint16(schema.id)
-      .string(schema.schemaName)
-      .string(schema.schemaEncoding)
-      .string(schema.schema);
+      .string(schema.name)
+      .string(schema.encoding)
+      .uint64(BigInt(schema.data.byteLength))
```
defunctzombie: This should be a uint32 prefix. The existing code is correct binary wise.

defunctzombie (changes requested):
The spec says uint32 prefix for the schema data

defunctzombie (approved):

PR #117, opened 2022-02-07
Add type prefix to serialized conformance files.

foxymiles:
**Public-Facing Changes**
This adds a type prefix to fields in the conformance test text files.

**Description**
This adds a single character type prefix to the fields in the conformance test expectation files. This will make it easier to write automated writer tests that use the same files.

It also updates the typescript reader to return `Map` values for metadata as per the current spec.


typescript/src/v0/types.ts
```diff
@@ -23,7 +23,7 @@ export type ChannelInfo = {
   topic: string;
   messageEncoding: string;
   schemaId: number;
-  metadata: [key: string, value: string][];
+  metadata: Record<string, string>;
```
jtbandes, line 26: I think we probably want to use Map for these instead of Record. Especially since we are already using Map for some other fields like messageIndexOffsets.

foxymiles: Yes you're right. I've changed this.

PR #118, opened 2022-02-07
Add python streaming reader.

foxymiles:
**Public-Facing Changes**
This implements a basic python streaming reader.

**Description**
This implements a simple python streaming reader class that passes the conformance test sweet. A streaming writer and indexed readers & writers will be added in subsequent PRs.


cspell.config.yaml
```diff
@@ -21,10 +21,16 @@ words:
   - mcap
   - nsec
   - nsecs
+  - pipenv
   - proto
   - protobuf
+  - pypi
+  - pyright
+  - pytest
+  - rglob
```
jtbandes, line 30: maybe put these in the "overrides" section below under a python specific set of rules

python/LICENSE
```diff
@@ -0,0 +1 @@
+TBD
```
jtbandes, line 1: remove this file or copy the one from the repo root?

python/mcap/mcap0/data_stream.py
```diff
@@ -0,0 +1,99 @@
...
+            return b""
+
+        data = self.__stream.read(length)
+        self.__count += len(data)
+        if data == b"":
+            raise EndOfFile()
+        return data
+
+    def read1(self) -> int:
+        [value] = struct.unpack("<B", self.read(1))
```
jtbandes, line 31: Docs say:

> Creating a `Struct` object once and calling its methods is more efficient than calling the `struct` functions with the same format since the format string only needs to be compiled once.

But also:
> Note: The compiled versions of the most recent format strings passed to Struct and the module-level functions are cached, so programs that use only a few format strings needn’t worry about reusing a single Struct instance.

I'm not sure what qualifies as "a few" though, so maybe it's better to use Struct instances?

python/mcap/mcap0/opcode.py
```diff
@@ -0,0 +1,20 @@
...
+    DATA_END = 0x0F
+    FOOTER = 0x02
+    HEADER = 0x01
+    MESSAGE = 0x05
+    MESSAGE_INDEX = 0x07
+    METADATA = 0x0C
+    METADATA_INDEX = 0x0D
+    SCHEMA = 0x03
+    STATISTICS = 0x0B
+    SUMMARY_OFFSET = 0x0E
```
jtbandes, line 20: order by value?

python/mcap/mcap0/records.py
```diff
@@ -0,0 +1,383 @@
...
+            log_time=log_time,
+            data_size=data_size,
+            name=name,
+            content_type=content_type,
+        )
+
+
+@dataclass
+class ChannelInfo(McapRecord):
+    id: int = field(metadata={"value_type": ["int"]})
```
jtbandes, line 80: why do we use this `metadata` thing for ints but not strings?

python/mcap/mcap0/stream_reader.py
```diff
@@ -0,0 +1,133 @@
...
+            records.append(channel_info)
+        elif opcode == Opcode.MESSAGE:
+            message = Message.read(stream, length)
+            records.append(message)
+        elif opcode == Opcode.SCHEMA:
+            schema = Schema.read(stream)
+            records.append(schema)
+        else:
+            # Unknown chunk record type
+            stream.read(length)
```
jtbandes, line 47: maybe worth exposing some kind of `seek` method to avoid keeping the data in memory?

python/mcap/mcap0/stream_reader.py
```diff
@@ -0,0 +1,133 @@
...
+    return records
+
+
+def get_chunk_data_stream(chunk: Chunk) -> Tuple[ReadDataStream, int]:
+    if chunk.compression == "lz4":
+        data: bytes = lz4.frame.decompress(chunk.data)  # type: ignore
+        return ReadDataStream(BytesIO(data)), len(data)
+    elif chunk.compression == "zstd":
+        data = zlib.decompress(chunk.data)
```
jtbandes, line 56: zlib and zstd are different, this will need to be changed.

python/tests/test_read_conformance.py
```diff
@@ -0,0 +1,30 @@
+import difflib
+import unittest
+from pathlib import Path
+
+from mcap.mcap0.records import DataEnd
+from mcap.mcap0.serialization import stringify_record
+from mcap.mcap0.stream_reader import StreamReader
+
+
+class McapStreamingReaderConformanceTestCase(unittest.TestCase):
```
jtbandes, line 10: does this still need to exist or can it be deleted since you added the test runner in TS?

jtbandes (commented):
Looking good! 🐍

python/mcap/mcap0/opcode.py
foxymiles: Personally I find alpha more useful here.

python/mcap/mcap0/records.py
foxymiles: We only have one type of string to handle but need to handle different ints differently. This will go away in the json version.

python/tests/test_read_conformance.py
foxymiles: It's moderately useful since its much faster than the TS runner.

python/mcap/mcap0/data_stream.py
foxymiles: I would imagine we're in the "few" range here since there are < 10 we use but it might be interesting to do some benchmarking.

PR #126, opened 2022-02-08
schema_id=0 indicates no schema

jhurliman:
Improve support for schemaless message encodings (JSON, CBOR, YAML, CSV, H.26x, text, compressed audio, etc) by using a `schema_id` of 0 to indicate "no schema".


docs/specification/README.md
```diff
@@ -166,7 +166,7 @@ Channel Info records are uniquely identified within a file by their channel ID.
 | 2 | id | uint16 | A unique identifier for this channel within the file. |
 | 4 + N | topic | String | The channel topic. |
 | 4 + N | message_encoding | String | Encoding for messages on this channel. The value should be one of the [well-known message encodings](./well-known-encodings.md). Custom values should use `x-` prefix. |
-| 2 | schema_id | uint16 | The schema for messages on this channel. |
+| 2 | schema_id | uint16 | The schema for messages on this channel. A schema_id of 0 indicates no schema for this channel |
```
defunctzombie, line 169:
```suggestion
| 2 | schema_id | uint16 | The schema for messages on this channel. A schema_id of 0 indicates there is no schema for this channel. |
```

amacneil:
👍 from me

the only alternative I see (to support json/cbor/etc where the schema is optional) is that people have to write an empty schema record with no name and no schema, which seems a bit silly.

defunctzombie (approved):
Please add a note in the Schema Record section indicating that readers should ignore schemas with id 0.

jtbandes (approved):

jtbandes:
Should this be considered a binary breaking change since previously-valid files become invalid or unreadable with this change?

wkalt:
Personally I feel like overloading the "id" this way is a little weird and special-casey. We still need a way for a Schema record to indicate "no schema", probably via an empty schema, because someone might want a name for the empty schema, so this doesn't save us the work of modeling that somehow. Readers and writers will need to introduce conditional logic in a number of cases to account for it in validation and such. Failure to account for it could cause subtle issues in writers, like messages that can still be "read" from the file, but not parsed with a schema, due to having been erroneously assigned a zero-valued id. Readers that fail to account for it could erroneously read a file like that, despite it not being valid. Dunno if it seems worth the overhead.

amacneil:
I think "someone might want a name for the empty schema" is a separate concern: if there is a schema name but we don't have the data (e.g. you converted a ros 2 bag file directly and we couldn't find msg defs), that's different than not having a schema at all.

For a lot of streams (e.g. json) there is no schema required and no name for the schema required, there is just a channel of data. So it seems reasonable to have a concept of "null" schema. Maybe using `0` to represent null is sadness?

I don't fully understand the argument about buggy readers/writers - surely "buggy readers/writers will read/write bad data" is true of almost every aspect of mcap? We need conformance tests to ensure that it is correctly skipped when writing and ignored when reading.

defunctzombie:
I think this comes after the answer to https://github.com/foxglove/mcap/issues/139

defunctzombie:
We are in agreement that schema_id=0 should indicate there is no schema.

PR #128, opened 2022-02-08
Spec: Rename Channel Info record to Channel

amacneil:
Proposal: Rename `Channel Info` to `Channel`.

Rationale:
- Most records are "info" - it's not clear why we treat `Channel` differently than, say, `Schema`.
- This change also makes "primary key" fields such as `Channel.id` more consistent with their "foreign key" relationship fields such as `Message.channel_id`.

Non-binary breaking, although this change would require updates in client libraries.

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -17,7 +17,7 @@ MCAP files are designed to work well under various workloads, resource constrain
   - [Header](#header-op0x01)
   - [Footer](#footer-op0x02)
   - [Schema](#schema-op0x03)
-  - [Channel Info](#channel-info-op0x04)
+  - [Channel](#channel-op0x04)
   - [Message](#message-op0x05)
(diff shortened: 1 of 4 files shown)


jhurliman (approved):

defunctzombie:
`Channel` seems good - As you said, if you were representing this in a database you could easily have a `channels` table with rows corresponding to channel records.

Here's one counterpoint: a `Channel` could also be thought of as the thing that contains the messages (kinda like a chunk) but here it's only used to define information about a channel and does not contain the actual messages on the channel. Schema does contain the actual schema data. I don't think this counterpoint is enough to say we shouldn't call it `Channel`. If it was called `Channel` would we be thinking to change it to `Channel Info`? probably not.

wkalt (approved):

jtbandes (approved):

PR #134, opened 2022-02-09
Add variant key to expected outputs.

foxymiles:
**Public-Facing Changes**
This adds a new `meta` key to the expected output json.

**Description**
This is used to help writer tests more precisely duplicate the mcap output by passing along more detailed information about the options used to generate the mcap reference file.

### DIFF
--- a/tests/conformance/scripts/run-tests/index.ts
+++ b/tests/conformance/scripts/run-tests/index.ts
@@ -8,8 +8,11 @@ import generateTestVariants from "variants/generateTestVariants";
 import runners from "./runners";
 
 function normalizeJson(json: string): string {
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call
-  return stringify(JSON.parse(json));
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const data = JSON.parse(json);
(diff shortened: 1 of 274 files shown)


tests/conformance/scripts/run-tests/index.ts
```diff
@@ -8,8 +8,11 @@ import generateTestVariants from "variants/generateTestVariants";
 import runners from "./runners";
 
 function normalizeJson(json: string): string {
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call
-  return stringify(JSON.parse(json));
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const data = JSON.parse(json);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  delete data.meta;
```
jtbandes: Seems a little unfortunate that this `data.meta` is completely disconnected from the place where meta is added before stringifying. If the two were tied together by some type definition it could aid in debugging and changing the code later.

jtbandes (approved):

tests/conformance/scripts/run-tests/index.ts
foxymiles: Good point. Maybe once I get the writer conformance tests working I'll circle back on this.

PR #135, opened 2022-02-09
ts/py: add metadataCount to Statistics

jtbandes:
TypeScript and Python implementations were missing metadata_count.


tests/conformance/scripts/generate-inputs.ts
```diff
@@ -144,10 +144,11 @@ function generateFile(features: Set<TestFeatures>, records: TestDataRecord[]) {
   const statisticsStart = BigInt(builder.length);
   if (features.has(TestFeatures.UseStatistics)) {
     builder.writeStatistics({
-      attachmentCount,
-      chunkCount,
       messageCount,
       channelCount,
+      attachmentCount,
+      metadataCount: 0,
```
wkalt, line 150: it's not clear to me why this gets a default but not attachment count

wkalt (approved):

tests/conformance/scripts/generate-inputs.ts
jtbandes: though I guess the code does support it. so I should update this and in the future we can add test files with metadata easily

PR #136, opened 2022-02-09
Add schema_count to Statistics

### LINKED ISSUE #137: Add schema_count to Statistics
This was overlooked in #102. schema_count can be uint16 because schema id 0 is reserved (contingent on https://github.com/foxglove/mcap/pull/126).

This could be implemented without binary breakage by appending the field to the end of the Statistics record. Or for aesthetic (and fixed-offset) reasons we could put it earlier in the record with a binary breakage.

jtbandes:
Open for discussion. When the Schema record was added we didn't think to add schema_count to Statistics.

Resolves #137

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -270,7 +270,8 @@ A Statistics record contains summary information about the recorded data. The st
 | Bytes | Name | Type | Description |
 | --- | --- | --- | --- |
 | 8 | message_count | uint64 | Number of Message records in the file. |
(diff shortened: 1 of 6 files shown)


defunctzombie:
We should turn this into an issue and consider all binary breaking changes on one go. If we want to make a PR tracking all the changes thats fine too - but we decided no more binary breaking one-off changes in individual PRs.

wkalt:
I have merged a change that covers the specification-only part of this. We will update the implementations in a separate patch today.

PR #140, opened 2022-02-10
Put fields before type in conformance json.

foxymiles:
**Public-Facing Changes**
This puts the `fields` key before the `type` key in the conformance JSON files.


jtbandes (changes requested):
I think this needs to be a change to the normalization, not the runners. The runners should ideally not be on the hook for outputting keys in a specific order.

PR #141, opened 2022-02-10
[C++] Streamed reader conformance tests

jhurliman:
Implements a conformance test runner wrapping the C++ streaming typed record reader API. Also fixes a bug this caught with KeyValueMap parsing


cpp/test/streamed-reader-conformance.cpp
```diff
@@ -0,0 +1,233 @@
...
+  const std::string inputFile = argv[1];
+  std::ifstream input(inputFile, std::ios::binary);
+  mcap::FileStreamReader dataSource{input};
+  mcap::TypedRecordReader reader{dataSource, 8};
+
+  reader.onHeader = [&](const mcap::Header& header) {
+    recordsJson.push_back(json::object({
+      {"type", "Header"},
+      {"fields", json::array({
+                   json::array({"library", header.library}),
```
jtbandes, line 42: I think you can simplify the syntax by omitting the `json::array` wrapper and the initializer list will be interpreted as an array

tests/conformance/scripts/run-tests/runners/CppStreamedTestRunner.ts
```diff
@@ -0,0 +1,20 @@
...
+import { promisify } from "util";
+import { TestVariant } from "variants/types";
+
+import { ITestRunner } from ".";
+
+export default class CppStreamedTestRunner implements ITestRunner {
+  name = "cpp-streamed";
+  async run(filePath: string): Promise<string> {
+    const { stdout } = await promisify(exec)(`./streamed-reader-conformance ${filePath}`, {
+      cwd: join(__dirname, "../../../../../cpp/test/build/Debug/bin"),
```
jtbandes, line 12: lol

jtbandes (approved):
👏

cpp/test/streamed-reader-conformance.cpp
jhurliman: I tried a few variations, it was getting confused because key/value entries in a map are initialized with {a, b} so it couldn't tell the difference between a key/value pair and a two-entry array

PR #148, opened 2022-02-10
Add writer conformance test support.

foxymiles:
**Public-Facing Changes**
This adds basic support for writer conformance tests.

**Description**
For now this just adapts the harness to support reader & writer tests and adds passthrough writer tests for python & typescript. Implementation of the tests will follow in subsequent PRs.

Writer tests should output the mcap they generate as hex encoded strings of the entire file.

### DIFF
--- a/tests/conformance/scripts/run-tests/index.ts
+++ b/tests/conformance/scripts/run-tests/index.ts
@@ -1,3 +1,4 @@
+import colors from "colors";
 import { program } from "commander";
 import * as Diff from "diff";
 import fs from "fs/promises";
@@ -5,7 +6,14 @@ import stringify from "json-stringify-pretty-compact";
 import path from "path";
 import generateTestVariants from "variants/generateTestVariants";
 
-import runners from "./runners";
+import runners, { ITestRunner } from "./runners";
+
+type TestOptions = {
+  dataDir: string;
+  runner?: string;
+  update: boolean;
+  testRegex?: RegExp;
+};
 
 function normalizeJson(json: string): string {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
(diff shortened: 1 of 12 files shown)


jtbandes:
> Writer tests should output the mcap they generate as hex encoded strings of the entire file.

This seems strange. I think writers should just output binary mcap, like they would normally be expected to do.

tests/conformance/scripts/run-tests/index.ts
```diff
@@ -41,51 +162,13 @@ async function main(options: {
...
-        }
-      }
-    }
+    const testFunction = runner.mode === "read" ? runReaderTest : runWriterTest;
+    const { hadError: newHadError, foundAnyTests: newFoundAnyTests } = await testFunction(
+      runner,
+      options,
+    );
+    hadError ||= newHadError;
+    foundAnyTests ||= newFoundAnyTests;
```
jtbandes, line 171: `foundAnyTests` is going to be the same for all readers/writers test runs. It might make more sense to just move it out of the runReaderTest/runWriterTest functions.

tests/conformance/scripts/run-tests/runners/TypescriptStreamedWriterTestRunner.ts
```diff
@@ -0,0 +1,22 @@
...
+    return true;
+  }
+
+  async run(filePath: string, _variant: TestVariant): Promise<string> {
+    // Passthrough test for now.
+    const mcapPath = filePath.replace(/\.[^.]+$/, ".mcap");
+    const bytes = await fs.readFile(mcapPath);
+    return Array.from(bytes)
+      .map((b) => b.toString(16).padStart(2, "0"))
+      .join("");
```
jtbandes, line 20: I know these passthrough tests will be removed, but I think a cleaner way of disabling the writer runner would be to just return false from supportsVariant.

foxymiles: I wanted to test the whole pipeline including output.

foxymiles:
> > Writer tests should output the mcap they generate as hex encoded strings of the entire file.
> 
> This seems strange. I think writers should just output binary mcap, like they would normally be expected to do.

That would be nicer but would require a different return type from the `run` function. Maybe worth changing?

jtbandes:
> That would be nicer but would require a different return type from the `run` function. Maybe worth changing?

since the reader & writer runners are separate, we can use separate interfaces for them right?

foxymiles:
> > That would be nicer but would require a different return type from the `run` function. Maybe worth changing?
> 
> since the reader & writer runners are separate, we can use separate interfaces for them right?

Yeah. I've been trying to move on this quickly so I can get back to actually implementing a writer and getting the tests in a state where other people can use them but maybe we can do another pass on this after?

PR #149, opened 2022-02-10
Update implementations for breaking changes

wkalt:
Updates all implementations for breaking changes and renames:
- Move Channel.schema_id before Channel.topic
- Add schema_count, message_start_time, message_end_time to Statistics
- Rename Chunk/ChunkIndex start_time/end_time to message_start_time/message_end_time
- Swap Message.log_time and publish_time

Also, improves the summary output of `mcap info` by adding table alignment and incorporating information from Schema records.


cpp/examples/mcapdump.cpp
```diff
@@ -96,9 +96,9 @@ std::string ToString(const mcap::ChunkIndex& chunkIndex) {
     "[ChunkIndex] start_time={}, end_time={}, chunk_start_offset={}, chunk_length={}, "
```
jtbandes, line 96: @jhurliman do we need to update this or can we perhaps remove mcapdump now that we have a c++ conformance test runner?

PR #151, opened 2022-02-10
Change schema_count to uint16

jtbandes:
~If #126 merges,~ it's impossible to have more than 2^16-1 schemas because Schema.id is a uint16 and id 0 is not allowed.


foxymiles (approved):

wkalt (approved):

jhurliman:
I support this if we move schema_count before channel_count in the record layout.

PR #153, opened 2022-02-10
Add motivation document

amacneil:
Import the original white paper (as posted on [ROS Discourse](https://discourse.ros.org/t/evaluation-of-robotics-data-recording-file-formats/22588)) describing motivation for mcap file format, and link from readme.

I think this is an important document, and is currently the best source of information we have about the design principles for mcap. In the future, we may wish to write a longer paper introducing mcap and comparing with alternatives more directly (e.g. performance comparisons now that we have actual libraries available).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ It is primarily intended for use in robotics applications, and works well under 
 ## Documentation
 
 - [File format specification](./docs/specification)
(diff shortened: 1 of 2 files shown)


amacneil:
Oh cool, didn't see that. There was some discussion in slack about preferring markdown. Happy to go with either.

PR #158, opened 2022-02-11
[C++] Improvements, documentation

jhurliman:
- Move StrFormat out of internal:: since it is used in examples
- Iterate over MessageView instead of Message, providing pointers to Channel and Schema
- Catch missing Schema/Channel problems
- Document public interface, rename noIndexing to noSummary


cpp/mcap/include/mcap/mcap.hpp
```diff
@@ -224,57 +327,201 @@ struct MetadataIndex {
...
+   *   pointer to point directly at the source data if possible. The pointer and
+   *   data must remain valid and unmodified until the next call to read().
+   * @param offset The offset in bytes from the beginning of the file to read.
+   * @param size The number of bytes to read.
+   * @return uint64_t Number of bytes actually read. This may be less than the
+   *   requested size if the end of the file is reached. The output pointer must
+   *   be readable from `output` to `output + size`. If the read fails, this
+   *   method should return 0.
+   */
   virtual uint64_t read(std::byte** output, uint64_t offset, uint64_t size) = 0;
```
jtbandes, line 476: `std::byte const** output` ?

jhurliman:
```
mcap.inl:1430:27: error: cannot initialize a parameter of type 'const std::byte **' with an rvalue of type 'std::byte **'
  bytesRead = reader.read(&record->data, offset + 9, record->dataSize);
                          ^~~~~~~~~~~~~
```

jtbandes: Why is record->data a pointer to mutable bytes?

jhurliman: I think there are a few places with const incorrectness that need to be cleaned up.

PR #160, opened 2022-02-11
Represent 'no compression' with 'none' instead of ''

wkalt:
This is useful because consumers of the format may otherwise need to
work around the empty rendering of "" in display settings. A positive
value like "none" allows UIs to reliably display the compression format
directly to the screen.

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -197,7 +197,7 @@ All messages in the chunk must reference channels recorded earlier in the file (
 | 8 | message_end_time | Timestamp | Latest message log_time in the chunk. Zero if the chunk has no messages. |
 | 8 | uncompressed_size | uint64 | Uncompressed size of the `records` field. |
 | 4 | uncompressed_crc | uint32 | CRC32 checksum of uncompressed `records` field. A value of zero indicates that CRC validation should not be performed. |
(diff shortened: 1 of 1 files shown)


amacneil (approved):
+1

jhurliman:
-1

amacneil:
I guess to play devil's advocate: What would a reader do if the compression was empty string? Probably still assume no compression right? In that case, maybe this change isn't a good idea?

defunctzombie:
@amacneil correct - _our_ readers would treat as no compression. I don't think this needs to be binary breaking. IMO its fine to update the spec to say something like _use "none" to indicate no compression. Tho it is different from other fields where we let empty string mean none (i.e. schema encoding). So do we recommend "none" there too?

I don't see as "ui needs to write some code to display values" as enough of a reason. I do think a good reason is that its nice to be explicit rather than implicit, and empty string is a common _default_ and is not explicit about your intent.

I am generally 👎 on this for now.

amacneil:
Changing vote to 👎 because roman pointed out there are other areas where "" means none. Seems weird to be inconsistent.

defunctzombie:
We are going to stick with empty string. Additional reasoning is that a non-zero value here indicates that there is compression used.

PR #162, opened 2022-02-11
Improve TS writer, add TS writer test runner

jtbandes:


typescript/src/v0/Mcap0Writer.ts
```diff
@@ -0,0 +1,415 @@
...
+  private useSummaryOffsets: boolean;
+  private repeatSchemas: boolean;
+  private repeatChannels: boolean;
+
+  // indices
+  private chunkIndices: ChunkIndex[] | undefined;
+  private attachmentIndices: AttachmentIndex[] | undefined;
+  private metadataIndices: MetadataIndex[] | undefined;
+
+  constructor({
```
defunctzombie: I am not a fan of de-structuring inline with function declaration. I find it hard to read. If it doesn't make much difference to you I'd prefer if this was split into the function body.

defunctzombie:
```suggestion
  constructor(options: Mcap0WriterOptions) {
    const {
    writable,
    useStatistics = true,
    useSummaryOffsets = true,
    useChunks = true,
    repeatSchemas = true,
    repeatChannels = true,
    useAttachmentIndex = true,
    useMetadataIndex = true,
    useMessageIndex = true,
    useChunkIndex = true,
  } = options;
```

PR #163, opened 2022-02-11
Separate test reader & writer interfaces.

foxymiles:
**Public-Facing Changes**
This splits the `ITestRunner` interface into two separate reader & writer interfaces.

**Description**
This allows us to return binary from writer tests which is more natural for comparison.

### DIFF
--- a/tests/conformance/scripts/run-tests/runners/TestRunner.ts
+++ b/tests/conformance/scripts/run-tests/runners/TestRunner.ts
@@ -1,0 +1,13 @@
+import { TestVariant } from "../../../variants/types";
+
+export abstract class ReadTestRunner {
+  abstract readonly name: string;
+  abstract supportsVariant(variant: TestVariant): boolean;
+  abstract runReadTest(filePath: string, variant: TestVariant): Promise<string>;
+}
+
+export abstract class WriteTestRunner {
+  abstract readonly name: string;
+  abstract supportsVariant(variant: TestVariant): boolean;
+  abstract runWriteTest(filePath: string, variant: TestVariant): Promise<Uint8Array>;
+}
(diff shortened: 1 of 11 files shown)


tests/conformance/scripts/run-tests/runners/TestRunner.ts
```diff
@@ -0,0 +1,13 @@
+import { TestVariant } from "../../../variants/types";
+
+export abstract class ReadTestRunner {
```
jtbandes, line 3: Interesting, why the choice of `abstract class` instead of `interface`? I haven't used abstract classes in TS before and they seem quite similar.

edit: oh, I see, it's so you can do `instanceof`.

tests/conformance/scripts/run-tests/runners/TypescriptStreamedWriterTestRunner.ts
```diff
@@ -1,22 +1,19 @@
...
 
-  async run(filePath: string, _variant: TestVariant): Promise<string> {
+  async runWriteTest(filePath: string, _variant: TestVariant): Promise<Uint8Array> {
     // Passthrough test for now.
     const mcapPath = filePath.replace(/\.[^.]+$/, ".mcap");
     const bytes = await fs.readFile(mcapPath);
-    return Array.from(bytes)
-      .map((b) => b.toString(16).padStart(2, "0"))
-      .join("");
+    return bytes as Uint8Array;
```
jtbandes, line 17: This cast shouldn't be necessary I think? Buffer inherits from Uint8Array.

tests/conformance/scripts/run-tests/runners/index.ts
```diff
@@ -1,21 +1,19 @@
...
+const runners = [
   new CppStreamedTestRunner(),
   new GoStreamedTestRunner(),
   new PythonStreamedReaderTestRunner(),
   new PythonStreamedWriterTestRunner(),
   new TypescriptIndexedTestRunner(),
   new TypescriptStreamedReaderTestRunner(),
   new TypescriptStreamedWriterTestRunner(),
-];
+] as const;
```
jtbandes, line 17: Rather than as const I think it would be better to annotate `const runners: readonly (ReadTestRunner|WriteTestRunner)[]` so that the specific types are not exposed to anyone importing this file.

jtbandes (approved):

PR #157, opened 2022-02-11
Move log_time earlier in Attachment, rename created_at to creation_time, add creation_time to Attachment Index, describe missing creation_time handling

jhurliman:
**Public-Facing Changes**

Move log_time earlier in Attachment, rename created_at to creation_time, describe missing creation_time handling

PR #174, opened 2022-02-11
Move log_time earlier in Attachment, rename created_at to creation_time, add create_time to Attachment Index

jtbandes:
Implement spec changes from #157.


jhurliman (approved):
C++ lgtm

PR #175, opened 2022-02-11
Clarify lengths include opcode+length prefix

jtbandes:
Add clarifying wording to a few length fields.


jhurliman (approved):

defunctzombie (approved):
Records are already defined as these things - guess it doesn't hurt to say it again.

PR #178, opened 2022-02-11
Move attachment name

wkalt:
Moves attachment name under attachment creation time in attachment
records.


wkalt:
Oops - it's here: https://github.com/foxglove/mcap/pull/179.

PR #187, opened 2022-02-12
Go: Only build static binary on linux

### LINKED ISSUE #89: "make build" for go cli tool fails on mac
Building with `go build` succeeds. This is probably resulting from choice of build flags for sqlite.

### LINKED ISSUE #183: Go command line tool build flags on MacOS
**Description**
On mac the static flag is not possible. Had to remove the flag from the makefile and then it works flawlessly.

Line: https://github.com/foxglove/mcap/blob/1c4f695ad04a94ef3d3ffec124a7dd9693807ca4/go/mcap/Makefile#L3

- Platform: MacOS 12.1

amacneil:
**Public-Facing Changes**
None

**Description**
Don't build static binaries on macOS (not possible). Debatable whether we need this on linux, but it's nice if we plan to generate binaries that people want to run under busybox/alpine.

Closes #89
Closes #183

### DIFF
--- a/go/.gitignore
+++ b/go/.gitignore
@@ -1,0 +1,1 @@
+mcap/mcap
--- a/go/mcap/Makefile
+++ b/go/mcap/Makefile
@@ -1,6 +1,15 @@
+# stripe binaries
+FLAGS := -tags sqlite_omit_load_extension -ldflags '-s -w'
+
+# platform-specific settings
+GOOS := $(shell go env GOOS)
+ifeq ($(GOOS),linux)
+	# statically link binaries to support alpine linux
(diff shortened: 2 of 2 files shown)


go/mcap/Makefile
```diff
@@ -1,6 +1,15 @@
+# stripe binaries
```
wkalt, line 1: strip?

wkalt (approved):
:+1: pending rebase + wording

go/.gitignore
```diff
@@ -0,0 +1 @@
+mcap/mcap
```
jtbandes, line 1: can we put all binaries in a `bin` dir or something?

wkalt: I am +1 but think we can handle it in a separate patch - I know you have a ticket for it already. I'll merge this one.

PR #190, opened 2022-02-12
Run CI conformance tests in parallel, C++ build deps caching

jhurliman:


wkalt:
I think it's nice to have them split out rather than failing on first fail, but the timing difference is going to end up being negligible because the 99% long pole is the C++ build. Still in favor though.

wkalt (approved):

jtbandes:
Can we use a matrix for this instead of duplicating steps?

jhurliman:
@wkalt C++ tests are about half the time of Python tests now. I'm not sure how to make the Python tests run faster.

PR #195, opened 2022-02-14
Don't include padding in attachnent index length.

foxymiles:
**Public-Facing Changes**
This excludes the length of the attachment record's padding from the length in the corresponding `AttachmentIndexField`.

For now this just fixes the logic in the generate-inputs script pending a more comprehensive API for users to pad data themselves.

### DIFF
--- a/tests/conformance/scripts/generate-inputs.ts
+++ b/tests/conformance/scripts/generate-inputs.ts
@@ -82,7 +82,7 @@ function generateFile(features: Set<TestFeatures>, records: TestDataRecord[]) {
         if (features.has(TestFeatures.UseAttachmentIndex)) {
           attachmentIndexes.push({
             name: record.name,
-            length,
+            length: length - (features.has(TestFeatures.AddExtraDataToRecords) ? 3n : 0n),
             offset,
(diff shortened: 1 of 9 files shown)


jtbandes (changes requested):
Based on my reading of the spec, the `length` of the attachment record is supposed to be the length of the _whole_ record. A padded attachment will have extra length.

defunctzombie (changes requested):
The `length` record is suppose to be the entire TLV length including padding. `data_size` should not include any padding.

tests/conformance/scripts/generate-inputs.ts
```diff
@@ -82,7 +82,7 @@ function generateFile(features: Set<TestFeatures>, records: TestDataRecord[]) {
         if (features.has(TestFeatures.UseAttachmentIndex)) {
           attachmentIndexes.push({
             name: record.name,
-            length,
+            length: length - (features.has(TestFeatures.AddExtraDataToRecords) ? 3n : 0n),
```
defunctzombie, line 85: I don't understand why we are subtracting from the length here. The length is the length of the attachment record (including padding).

PR #197, opened 2022-02-14
Implement python writer.

foxymiles:
**Public-Facing Changes**
This implements a basic python writer that passes all conformance tests.


python/tests/run_writer_test.py
```diff
@@ -35,7 +22,7 @@ def deserialize_value(klass: McapRecord, field: str, value: Any) -> Any:
 
 
 def deserialize_record(key: str, fields: List[Dict[str, Any]]) -> McapRecord:
-    klass = getattr(sys.modules[__name__], key)
+    klass = [c for c in McapRecord.__subclasses__() if c.__name__ == key][0]
```
jtbandes, line 25: Nit: maybe use `next(c for c in ...)`?

Philosoraptor: is this metaprogramming better than just having a mapping from name to class?

PR #207, opened 2022-02-15
Spec: Consistent proto -> protobuf

amacneil:
So that the [schema encoding](https://github.com/foxglove/mcap/blob/main/docs/specification/well-known-schema-encodings.md) matches the [message encoding](https://github.com/foxglove/mcap/blob/main/docs/specification/well-known-message-encodings.md).


defunctzombie:
/shrug - the idea was to support string schemas in addition to binary `FileDescriptorSet` in some future. The string schema files have the `.proto` extension: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto

amacneil:
If we want to support both text and filedescriptorset, that's fine. Unclear if anyone has a need for that. But if we do, we should make it clear - calling one of them `proto` doesn't help.

jtbandes (approved):

amacneil:
@defunctzombie @wkalt @jhurliman votes?

defunctzombie (approved):
protobuf is more namespaced than the more generic "proto" so 👍 from me

PR #210, opened 2022-02-15
Add mcap-examples package and revive validate script

### LINKED ISSUE #181: Revive TypeScript validate script
Having some script that can be run on a .mcap file in each language seems useful. For TypeScript we had this but it was removed in #112.

jtbandes:
Move main TS package under mcap subdirectory; add mcap-example-validate package: `yarn workspace @foxglove/mcap-example-validate validate file.mcap`

Closes #181


defunctzombie (approved):
I'd have a slight preference for an `examples` folder within `typescript` and folders within that with individual examples. I find that easier to copy/paste and start new projects.

jtbandes:
> an examples folder within typescript and folders within that with individual examples

This would mean we have to include wasm-lz4, zstd-codec, rosmsg, etc as devDependencies in the typescript package. The published package would also contain the example scripts by default. Isn't it nicer to make them separate? In ws-protocol making them separate was a hard requirement because we wanted people to be able to `npx @foxglove/ws-protocol-examples ...` which meant that we needed real `dependencies` that we wouldn't have wanted in the base ws-protocol package. If we ever want to do the same thing with this examples package then we would have the same issue.

defunctzombie:
> This would mean we have to include wasm-lz4, zstd-codec, rosmsg, etc as devDependencies in the typescript package

I don't follow? rather than your `mcap-examples` package, I would make an examples folder. In that folder there are folders with `package.json` files. Each one is a stand-alone package.

defunctzombie:
> In ws-protocol making them separate was a hard requirement because we wanted people to be able to `npx @foxglove/ws-protocol-examples ...`

I actually find this a bit annoying with the websocket protocol examples repo. Rather than individual examples I can copy/paste the project, all the examples are one project.

PR #215, opened 2022-02-16
TypeScript indexed reader: handle overlapping & out-of-order chunks

### LINKED ISSUE #212: TypeScript support for overlapping / out of order chunks

jtbandes:
**Public-Facing Changes**
The TypeScript Mcap0IndexedReader now handles reading out-of-order chunks, overlapping chunks, and out-of-order messages within chunks, and emits all messages in order of log time.

**Description**
Closes #212


typescript/mcap/src/v0/ChunkBuilder.ts
```diff
@@ -69,6 +71,14 @@ class ChunkBuilder {
     this.recordWriter.writeMessage(message);
   }
 
+  finish(): void {
+    if (this.messageIndices) {
+      for (const index of this.messageIndices.values()) {
+        index.records.sort(([logTimeA], [logTimeB]) => Number(logTimeA - logTimeB));
```
defunctzombie, line 77: Do message index records need to be sorted? This seems like extra work that doesn't need to happen? The reader side should be doing the sorting.

typescript/mcap/src/v0/Mcap0IndexedReader.ts
```diff
@@ -1,11 +1,216 @@
...
 import { DecompressHandlers, IReadable, TypedMcapRecords } from "./types";
 
+type ChunkCursorParams = {
+  chunkIndex: TypedMcapRecords["ChunkIndex"];
+  relevantChannels: Set<number> | undefined;
+  startTime: bigint | undefined;
+  endTime: bigint | undefined;
+};
+
+class ChunkCursor {
```
defunctzombie, line 16: Please move this to a separate file. I am a fan of keeping files smaller by keeping 1 "main" class per file. This class has grown sizable enough that it is easier to maintain as a standalone file.

defunctzombie: A comment for this class would be useful to future maintainers.

typescript/mcap/src/v0/Mcap0IndexedReader.ts
```diff
@@ -1,11 +1,216 @@
...
+    for (
+      let result;
+      (result = parseRecord({ view: messageIndexesView, startOffset: offset, validateCrcs: true })),
+        result.record;
+      offset += result.usedBytes
+    ) {
+      if (result.record.type !== "MessageIndex") {
+        continue;
+      }
+      if (
```
defunctzombie, line 168: Maybe less nesting if you flip this condition around and continue in it rather than put the main logic in it?

typescript/mcap/src/v0/Mcap0IndexedReader.ts
jtbandes: yea, was planning to do that, this PR is still a draft

jtbandes:
> Keeping any active chunks in memory? Or keeping at most N chunks in memory?

That is up to us, we can change how `loadChunkData` works. Currently there is no eviction at all. We could evict chunks after a certain number, a certain size, or only after the chunk is no longer 'active'.

typescript/mcap/src/v0/ChunkBuilder.ts
jtbandes: I thought this was part of the spec but you're right, it is not

defunctzombie (approved):

typescript/mcap/src/v0/Mcap0IndexedReader/ChunkCursor.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * ChunkCursor represents the reader's position in a particular chunk. The indexed reader holds
+ * ChunkCursors in a heap in order to merge multiple chunks together.
+ *
+ * Each chunk can contain multiple channels, and so a ChunkCursor itself contains a heap of cursors
+ * pointing into the message index for each channel of interest.
+ */
+export class ChunkCursor {
+  chunkIndex: TypedMcapRecords["ChunkIndex"];
```
defunctzombie, line 22: readonly?

jtbandes: sure, but true readonly is kinda hopeless, we would need to make/import a DeepReadonly util and sprinkle it everywhere.

Issue #209, opened 2022-02-15
enable attachment CRCs in writers


wkalt:
Currently the go and C++ writers are zeroing attachment CRCs in order to match the conformance tests. We need to update the typescript writer to add these to the conformance test files (or else add a parameter for them) so that they can be used, and then update the go and C++ implementations as well.

PR #219, opened 2022-02-16
Calculate Chunk, Attachment, and Summary CRC in writers, update conformance tests

jtbandes:
**Public-Facing Changes**
Writers now calculate CRC fields for Chunks, Attachment records, and the summary section.

**Description**
Closes https://github.com/foxglove/mcap/issues/209


go/mcap/writer.go
```diff
@@ -607,10 +607,21 @@ func (w *Writer) Close() error {
...
+	if summarySectionStart != 0 && w.opts.IncludeCRC {
+		// add partial footer record to end of summary section CRC
+		footerPrefix := make([]byte, 1+8+8+8)
+		footerPrefix[0] = byte(OpFooter)
+		putUint64(footerPrefix[1:], 8+8+4)
+		putUint64(footerPrefix[1+8:], summarySectionStart)
+		putUint64(footerPrefix[1+8+8:], summaryOffsetStart)
+		w.w.w.crc.Write(footerPrefix)
+		summaryCrc = w.w.Checksum()
+	}
```
wkalt, line 620: I think this can be replaced with something like,
```
err = w.WriteFooter(&Footer{
  SummarySectionStart:...,
  SummarySectionOffset: ...,
  CRC: ...,
})
```

wkalt: oh, sorry - my mistake. You are making this call below. I see what you are doing..

wkalt: what do you think of the idea of WriterFooter computing this CRC, while making an assumption that the CRC has been cleared after completion of the data section (after DataEnd, which I think is required?) Then we wouldn't need to do this partial footer record business, I believe

wkalt: a CRC provided to WriterFooter could be ignored in favor of the computation

go/mcap/writer.go
```diff
@@ -589,6 +588,7 @@ func (w *Writer) Close() error {
 	}
 
 	// summary section
+	w.w.Reset() // reset CRC to begin computing summaryCrc
```
wkalt, line 591: we should probably rename this to ResetCRC

wkalt: also, preexisting but it's confusing that WriteSizer knows anything about CRCs I guess. Maybe this is a WriteCRCSizer, or we can compose somehow.. dunno, maybe for another time.

go/mcap/writer.go
```diff
@@ -607,10 +607,21 @@ func (w *Writer) Close() error {
...
 	}
+	var summaryCrc uint32
+	if summarySectionStart != 0 && w.opts.IncludeCRC {
+		// add partial footer record to end of summary section CRC
+		footerPrefix := make([]byte, 1+8+8+8)
+		footerPrefix[0] = byte(OpFooter)
+		putUint64(footerPrefix[1:], 8+8+4)
+		putUint64(footerPrefix[1+8:], summarySectionStart)
+		putUint64(footerPrefix[1+8+8:], summaryOffsetStart)
+		w.w.w.crc.Write(footerPrefix)
```
wkalt, line 618: this is accessible because the module has a flat structure, but it shouldn't really be public to internal consumers. It would make sense to further split up the module to make this kind of access illegal.

go/mcap/writer.go
```diff
@@ -607,10 +607,21 @@ func (w *Writer) Close() error {
 			}
 		}
 	}
+	var summaryCrc uint32
+	if summarySectionStart != 0 && w.opts.IncludeCRC {
+		// add partial footer record to end of summary section CRC
+		footerPrefix := make([]byte, 1+8+8+8)
```
wkalt, line 613: there's a buffer for this, `w.msg` -- maybe should be `w.record`. You can ensure you have enough room in it with `w.ensureSized(1+8+8+8)`.

foxymiles (approved):
👍

Issue #223, opened 2022-02-17
ros2 converted mcap files are missing some type definitions in schemas


wkalt:
@bertaveira reports that converted MCAP files produce an error in studio: `Error initializing player: Expected 1 top level definition for 'geometry_msgs/TransformStamped' but found 0`.

I think what is most likely happening is the code that is crawling the schema directories to assemble message definitions is not correctly following complex types and concatenating schemas.

PR #225, opened 2022-02-17
Resolve schema dependencies in ros2 bag conversion

wkalt:
Prior to this commit, only the message definition for the top-level type
was included in the mcap. With this change the type dependencies are
resolved.


wkalt:
closes #223

PR #228, opened 2022-02-18
[C++] Skip Chunk compression unless data is >=1KB and compression saves at least 2% or forceCompression=true

jhurliman:
C++ McapWriter
- Skip Chunk compression entirely if the total uncompressed data is <1KB unless forceCompression=true.
- If compression saves <2%, discard the compression result and write an uncompressed Chunk unless forceCompression=true
- Change `data()` and `size()` of `IChunkWriter` to refer to uncompressed data, and add `compressedData()` and `compressedSize()` methods that are only valid after calling `end()`.

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -48,6 +48,6 @@ ci:
 	docker run --rm --name mcap_cpp -v $(HOME):$(HOME) mcap_cpp
 
 .PHONY: ci-format-check
-format-check:
+ci-format-check:
 	docker build -t mcap_cpp -f ci.Dockerfile .
 	docker run --rm --name mcap_cpp -v $(HOME):$(HOME) mcap_cpp python3 scripts/format.py .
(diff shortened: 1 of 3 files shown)


jhurliman:
@wkalt the primary purpose of this change is to avoid useless compression. The default chunk size (taken from rosbag) is 768KB, which means every `sensor_msgs/CompressedImage` message from a 4K camera could be written into its own chunk. Applying ZSTD compression on top of JPEG is a waste of CPU time and will likely only make the chunk larger. A user setting can't always fix this with heterogeneous channels since some chunks might be highly compressible while others are not at all.

The 2% is rather arbitrary, it could be 0% too but I wanted to include cases where there was a small amount of savings but not enough to justify the cycles it would take to decompress. As far as adapting this to other implementations, I think it would make sense with a 0% threshold. If you have 500KB of uncompressed data and 501KB of compressed data sitting in side by side buffers, why would you write the 501KB of compressed data?

PR #239, opened 2022-02-22
[C++] Switch from conan to CMake, build static libmcap, vendor runtime dependencies

jhurliman:
This switches the C++ build from conan to pure CMake and switches from a header-only library to a static libmcap library that vendors its own dependencies. Additional dependencies for unit tests, conformance tests, and benchmarking and fetched using the CMake FetchContent directive. The Makefile switches from using Docker by default to the host machine by default.


defunctzombie:
How is a developer that wants to use libmcap suppose to consume these files now? Do they need to copy the vendor folder over to their tree as well? I think we need to think this through a bit more before we bring this into the mcap repo like this.

Issue #252, opened 2022-02-28
TypeScript writer starts a new chunk every 10 messages


jtbandes:
For historical reasons the TS writer always rolls over to a new chunk at 10 messages. This should be changed to be more in line with the chunk sizing params of other writers.

https://github.com/foxglove/mcap/blob/6249b93459752b6ca7a5ba0a87bef97b8d2d7776/typescript/mcap/src/v0/Mcap0Writer.ts#L353

PR #254, opened 2022-02-28
TS writer: add chunkSize param with 1MB default

jtbandes:
**Public-Facing Changes**
The TypeScript `Mcap0Writer` now accepts a `chunkSize` option rather than creating a new chunk every 10 messages.

**Description**
Fixes https://github.com/foxglove/mcap/issues/252


defunctzombie (approved):
👍 

1MB feels small in this day and age (especially if images are involved)

jtbandes:
1MB is in line with defaults for Go and Python writers currently (C++ defaults to 768k)

PR #262, opened 2022-03-02
TS writer: fix duplicate data when using multiple chunks

jtbandes:
**Public-Facing Changes**
Fixed a bug in the TypeScript `Mcap0Writer` where message data would be duplicated between chunks, resulting in quadratic growth.


defunctzombie:
😮

PR #270, opened 2022-03-04
typescript: Improve error message for missmatched channels

defunctzombie:
Also fix the error message for message records missing a channel.

### DIFF
--- a/typescript/mcap/src/v0/Mcap0StreamReader.test.ts
+++ b/typescript/mcap/src/v0/Mcap0StreamReader.test.ts
@@ -228,7 +228,7 @@ describe("Mcap0StreamReader", () => {
       ]),
     );
     expect(() => reader.nextRecord()).toThrow(
-      "Encountered message on channel 42 without prior channel",
+      "Encountered message on channel 42 without prior channel record",
     );
   });
 
@@ -508,7 +508,9 @@ describe("Mcap0StreamReader", () => {
           messageEncoding: "utf12",
           metadata: new Map([["foo", "bar"]]),
         } as TypedMcapRecords["Channel"]);
-        expect(() => reader.nextRecord()).toThrow("differing channels for 42");
+        expect(() => reader.nextRecord()).toThrow(
+          /Channel record for id 42 \(topic: (myTopic|XXXXXXXX)\) differs from previous for the same id./,
+        );
       });
     },
   );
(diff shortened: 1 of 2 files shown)


defunctzombie:
I would actually recommend we remove the following checks from the reader _nextRecord_ call altogether:

```typescript
if (result.value?.type === "Channel") {
      const existing = this.channelsById.get(result.value.id);
      this.channelsById.set(result.value.id, result.value);
      if (existing && !isEqual(existing, result.value)) {
        throw new Error(
          `Channel record for id ${result.value.id} (topic: ${result.value.topic}) differs from previous for the same id.`,
        );
      }
    } else if (result.value?.type === "Message") {
      const channelId = result.value.channelId;
      const existing = this.channelsById.get(channelId);
      if (!existing) {
        throw new Error(`Encountered message on channel ${channelId} without prior channel record`);
      }
    }
```

Let the consumer handle this since they need to do the lookup anyway. They will be in a better position to ignore or error appropriately.

typescript/mcap/src/v0/Mcap0StreamReader.ts
```diff
@@ -100,12 +100,15 @@ export default class Mcap0StreamReader implements McapStreamReader {
       const existing = this.channelsById.get(result.value.id);
       this.channelsById.set(result.value.id, result.value);
       if (existing && !isEqual(existing, result.value)) {
-        throw new Error(`differing channels for ${result.value.id}`);
+        throw new Error(
+          `Channel record for id ${result.value.id} (topic: ${result.value.topic}) differs from previous for the same id.`,
```
wkalt, line 104: "from previous channel record..."?

defunctzombie, line 104:
```suggestion
          `Channel record for id ${result.value.id} (topic: ${result.value.topic}) differs from previous channel record of the same id.`,
```

typescript/mcap/src/v0/Mcap0StreamReader.test.ts
```diff
@@ -508,7 +508,9 @@ describe("Mcap0StreamReader", () => {
           messageEncoding: "utf12",
           metadata: new Map([["foo", "bar"]]),
         } as TypedMcapRecords["Channel"]);
-        expect(() => reader.nextRecord()).toThrow("differing channels for 42");
+        expect(() => reader.nextRecord()).toThrow(
+          /Channel record for id 42 \(topic: (myTopic|XXXXXXXX)\) differs from previous for the same id./,
```
defunctzombie, line 512:
```suggestion
          /Channel record for id 42 \(topic: (myTopic|XXXXXXXX)\) differs from previous channel record of the same id./,
```

PR #276, opened 2022-03-07
Decode raw records if possible in python stream reader

foxymiles:
**Public-Facing Changes**
This improves the ergonomics of the stream reader class.

**Description**
This adds a `decoded_records` method to the stream reader class that returns only decoded messages instead of all the lower level mcap records if the required ros1 libraries are available. We can expand on this later to support encodings other than ros1.

### DIFF
--- a/python/mcap/mcap0/stream_reader.py
+++ b/python/mcap/mcap0/stream_reader.py
@@ -1,8 +1,14 @@
 import struct
 from io import BufferedReader, BytesIO, RawIOBase
-from typing import Iterator, List, Optional, Tuple, Union
-from zstd import ZSTD_uncompress  # type: ignore
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
+
 import lz4.frame  # type: ignore
+from zstd import ZSTD_uncompress  # type: ignore
+
+try:
+    from genpy import dynamic  # type: ignore
+except ImportError:
+    dynamic = None
 
 from .data_stream import ReadDataStream
 from .exceptions import McapError
@@ -73,7 +79,7 @@ class StreamReader:
     @property
     def records(self) -> Iterator[McapRecord]:
         """
-        Returns records encountered in the MCAP in order.
+        Returns raw records encountered in the MCAP in order.
         """
         if not self.__magic:
             self.__magic = read_magic(self.__stream)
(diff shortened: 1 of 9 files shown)


jtbandes:
I don't think we should structure our Python MCAP lib with an (even optional) dependency on `genpy` or other ROS stuff. Could we make it a separate package instead? Like change the iterator to accept a "decoder" function that the user can configure?

foxymiles:
> I don't think we should structure our Python MCAP lib with an (even optional) dependency on `genpy` or other ROS stuff. Could we make it a separate package instead? Like change the iterator to accept a "decoder" function that the user can configure?

The goal here is make this as easy to use for casual users as possible. I guess if we also supplied the ros1 decoder that might be an ok solution.

jtbandes:
Sometimes explicit composition is easier in the long run than things that work "magically" :)

jhurliman:
I don't understand the terminology here. What is a "raw record"? ROS doesn't use the term record, only MCAP, and we decode all the MCAP record types already. The PR description says:

> This adds a decoded_records method to the stream reader class that returns only decoded messages

It sounds like messages (or message payloads) and records are being used as interchangeable terms here.

python/mcap/mcap0/stream_reader.py
```diff
@@ -96,6 +102,37 @@ def records(self) -> Iterator[McapRecord]:
...
+                    msg_type = dynamic.generate_dynamic(  # type: ignore
+                        record.name, record.data.decode()
+                    )
+                    msg_types[record.name] = msg_type[record.name]
+            if isinstance(record, Channel):
+                channels[record.id] = record
+            if isinstance(record, Message):
+                channel = channels[record.channel_id]
+                schema = schemas[channel.schema_id]
+                message = msg_types[schema.name]().deserialize(record.data)
```
jhurliman, line 133: What happens if this fails?

foxymiles:
> I don't understand the terminology here. What is a "raw record"? ROS doesn't use the term record, only MCAP, and we decode all the MCAP record types already. The PR description says:
> 
> > This adds a decoded_records method to the stream reader class that returns only decoded messages
> 
> It sounds like messages (or message payloads) and records are being used as interchangeable terms here.

Yeah something like `decoded_messages` would be a better name here.

python/mcap/mcap0/stream_reader.py
foxymiles: Should probably wrap this in a try/catch that throws an `McapError` right?

python/mcap/mcap0/writer.py
```diff
@@ -48,9 +48,9 @@ class Writer:
 
     def __init__(
         self,
-        output: Union[str, BytesIO, BufferedWriter],
+        output: Union[str, IO[Any]],
         chunk_size: int = 1024 * 1024,
-        compression: CompressionType = CompressionType.NONE,
+        compression: CompressionType = CompressionType.ZSTD,
```
wkalt, line 53: I think this change should definitely be made, maybe independent of the larger patch

foxymiles: Yeah I made this in separate smaller PR. I'm going to close this PR in favor of another package that provides a separate ROS1 decoder class.

foxymiles:
Closed in favor of https://github.com/foxglove/python-mcap-ros1-support

PR #280, opened 2022-03-08
Enable compression by default in python writer

foxymiles:
**Public-Facing Changes**
This enables zstd compression by default in the python writer.

**Description**
This enables zstd compression by default in the python writer and also adds some simple tests of writing compressed files.

### DIFF
--- a/python/tests/test_writer.py
+++ b/python/tests/test_writer.py
@@ -1,0 +1,56 @@
+import contextlib
+import json
+from tempfile import TemporaryFile
+
+from mcap.mcap0.writer import CompressionType, Writer
+
+
+@contextlib.contextmanager
+def generate_sample_data(compression: CompressionType):
+    file = TemporaryFile("w+b")
+    writer = Writer(file, compression=compression)
+    writer.start(profile="x-json", library="test")
+    schema_id = writer.register_schema(
+        name="sample",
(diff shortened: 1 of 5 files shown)


python/tests/test_writer.py
```diff
@@ -0,0 +1,56 @@
...
+        channel_id=channel_id,
+        log_time=0,
+        data=json.dumps({"sample": "test"}).encode("utf-8"),
+        publish_time=0,
+    )
+
+    writer.finish()
+    file.seek(0)
+
+    yield file
```
jtbandes, line 44: should this be wrapped in `with TemporaryFile(...)`? or is that only for NamedTemporaryFile?

jtbandes (approved):

python/tests/test_writer.py
foxymiles: I think this one will clean itself up automatically:

```Return a [file-like object](https://docs.python.org/3/glossary.html#term-file-like-object) that can be used as a temporary storage area. The file is created securely, using the same rules as [mkstemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp). It will be destroyed as soon as it is closed (including an implicit close when the object is garbage collected).```

PR #281, opened 2022-03-09
Add py.typed stub file

foxymiles:
**Public-Facing Changes**
Add a py.typed file to indicate the python mcap module includes type information

**Description**
This just adds a an empty py.typed file so that other tools recognize the mcap module as typed.

### DIFF
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = mcap
-version = 0.0.5
+version = 0.0.6
 description = MCAP libraries for Python
 long_description = file: README.md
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
Cool, didn't know about this. But do we need to add it to the built package somehow? I don't see it being picked up in the `python -m build` step

foxymiles:
> Cool, didn't know about this. But do we need to add it to the built package somehow? I don't see it being picked up in the `python -m build` step

Yeah it looks like we might need to specify it in the setup.cfg too:

https://docs.python.org/3/distutils/setupscript.html#installing-package-data

PR #289, opened 2022-03-12
CLI: JSON output option for cat

wkalt:
Adds a ros1msg JSON transcoder to the ros utilities package, and calls
it from the CLI's cat command when the --json switch is supplied.

### DIFF
--- a/go/cli/mcap/Makefile
+++ b/go/cli/mcap/Makefile
@@ -24,3 +24,6 @@ lint:
 
 test:
 	go test ./...
+
+bench:
+	make -C cmd bench
--- a/go/cli/mcap/cmd/cat.go
+++ b/go/cli/mcap/cmd/cat.go
@@ -1,28 +1,79 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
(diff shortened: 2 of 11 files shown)


wkalt:
This is a kind of cool feature IMO, but we will need to consider whether all the JSON representations here are ideal, and also whether/when we want to invest in support for protobuf and ros2msg.

The questionable representation choices I think are,
* Durations as nanosecond integers instead of strings "5h2m10s"
* Ignoring constants (not sure what to do with these)
* Time represented as an RFC3339 timestamp; could be a secs/nsecs object.
* []uint8 is rendered as base64, while other arrays are fully transcoded. This seems sensible to me and results in a major performance increase on e.g images.

go/ros/ros1_json_transcoder.go
```diff
@@ -0,0 +1,520 @@
...
+			continue
+		}
+		// comment
+		if strings.HasPrefix(line, "#") {
+			continue
+		}
+		// constant
+		if strings.Contains(line, "=") {
+			continue
+		}
```
wkalt, line 85: this incorrectly skips lines that have = in trailing comments

PR #301, opened 2022-03-15
CLI: Align frequencies in mcap info output

wkalt:


wkalt:
after reviewing a few different files, decided the current left-aligned behavior is preferable

PR #303, opened 2022-03-16
Upgrade to go 1.18

wkalt:


go/mcap/writer_test.go
```diff
@@ -30,7 +30,7 @@ func TestMCAPReadWrite(t *testing.T) {
 		assert.Equal(t, "ros1", profile)
 		library, _, err := readPrefixedString(record, offset)
 		assert.Nil(t, err)
-		assert.Equal(t, "mcap go #unknown", library)
+		assert.Equal(t, "mcap go #", library)
```
wkalt, line 33: this is a behavior change that I can't find a reference for, but the version still comes through correctly when the code is built. There is a new buildinfo package we can switch to using as well, but I haven't changed anything.

PR #333, opened 2022-03-24
Fix chunk start & end time calculation

foxymiles:
**Public-Facing Changes**
This fixes an error in the message start and end time calculation for chunks.

**Description**
The existing logic would overwrite a message start time of zero if any subsequent messages had a later start time. The new logic is to set the chunk start time and end time to the minimum and maximum times of all received messages.

Fixes https://github.com/foxglove/studio/issues/3073

### DIFF
--- a/python/mcap/mcap0/chunk_builder.py
+++ b/python/mcap/mcap0/chunk_builder.py
@@ -29,7 +29,9 @@ class ChunkBuilder:
     def add_message(self, message: Message):
         if self.num_messages == 0:
             self.message_start_time = message.log_time
-        self.message_end_time = message.log_time
+        else:
+            self.message_start_time = min(self.message_start_time, message.log_time)
+        self.message_end_time = max(self.message_end_time, message.log_time)
 
         if not self.message_indices.get(message.channel_id):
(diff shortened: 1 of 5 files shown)


python/mcap/mcap0/chunk_builder.py
```diff
@@ -27,7 +27,7 @@ def add_schema(self, schema: Schema):
         schema.write(self.record_writer)
 
     def add_message(self, message: Message):
-        if self.message_start_time == 0:
+        if self.num_messages == 0:
```
jhurliman, line 30: This is not correct, at least on its own. The Chunk start_time is the minimum timestamp in the Chunk, not the first timestamp (unless those happen to be the same).

jhurliman (changes requested):

python/mcap/mcap0/chunk_builder.py
jhurliman: That’s correct. It’s the range of the log_times contained in the chunk, regardless of their in-chunk ordering.

foxymiles: Updated logic to use min & max message time to set chunk start & end times.

PR #335, opened 2022-03-25
Go: Add support for sorting serialized chunks

wkalt:
Adds a SortChunkMessages option to the go writer, defaulted to false,
that sorts the messages and message index records in chunks prior to
writing to disk.

The sort used for the index records is insertion sort, under the
assumption messages will be mostly/almost-always time-ordered to start.
The sort of the chunk mirrors the swaps executed by the insertion sort,
staging copies through the w.msg buffer. Since the same buffer was used
to stage writes to the chunk, it is guaranteed to have sufficient room
without additional allocations.


go/mcap/writer.go
```diff
@@ -27,23 +43,26 @@ type Writer struct {
...
+	schemas           map[uint16]*Schema
+	messageIndexes    map[uint16]*MessageIndex
+	w                 *writeSizer
+	buf               []byte
+	msg               []byte
+	chunk             []byte
+	uncompressed      *bytes.Buffer
+	compressed        *bytes.Buffer
+	compressedWriter  resettableWriteCloser
+	uncompressedChunk *bytes.Buffer
```
wkalt, line 60: double check we need this and "uncompressed".. not clear why that would be

go/mcap/writer.go
```diff
@@ -374,6 +392,119 @@ func (w *Writer) WriteSummaryOffset(s *SummaryOffset) error {
...
+func u64(b []byte) uint64 {
+	return binary.LittleEndian.Uint64(b)
+}
+
+func max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
```
wkalt, line 404: move to utils

go/mcap/writer_test.go
```diff
@@ -3,13 +3,22 @@ package mcap
...
 	"time"
 
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	kilobyte = 1024
+	megabyte = 1024 * kilobyte
+	gigabyte = 1024 * megabyte
+)
```
wkalt, line 20: move to utils or testutils

go/mcap/writer.go
```diff
@@ -374,6 +392,119 @@ func (w *Writer) WriteSummaryOffset(s *SummaryOffset) error {
...
+
+// swapSlices swaps the intervals of buf defined by leftstart/leftend and
+// rightstart/rightend. The intervals are assumed to be nonoverlapping, and
+// bounds-checked. The argument tmp is a scratch buffer; if set to nil, or if
+// the length is insufficient, one will be allocated. The return value is the
+// length of scratch space used.
```
wkalt, line 410: the return value is the possibly-allocated scratch buffer

go/mcap/writer.go
```diff
@@ -374,6 +392,119 @@ func (w *Writer) WriteSummaryOffset(s *SummaryOffset) error {
...
+		copy(buf[leftstart:], buf[rightstart:rightstart+rightLen])
+		copy(buf[rightstart:rightstart+rightLen], scratch)
+	}
+	return tmp
+}
+
+// sortChunk sorts the input chunk, and the provided index, using the provided
+// index on (timestamp, offset). Uses an insertion sort under the assumption the
+// input chunk is mostly sorted already and disorderings are usually localized.
+func sortChunk(tmp []byte, chunk []byte, index []messageIndexEntry) {
```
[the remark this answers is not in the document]
wkalt: in general the chunk is already sorted, or nearly sorted, in which case a heap is overkill. I can take a look at a heap based approach though.

defunctzombie:
Probably we can close this? If we want it in the future we'll make a new issue and revisit.

PR #344, opened 2022-03-28
Go: reduce seeking in indexed playback

wkalt:
Contains three optimizations to reduce seeking in indexed playback, primarily targeted at improving performance when reading remote files:
* Read message index records in one pass, rather than individual reads for relevant indexes
* Do not "seek" the remote reader if it is already in desired position.
* Read message indexes in the same read as the chunk they correspond to.

This commit also drops support for transparently merging and sorting chunks that are out of order. If this ends up being a requested feature at some point we can reintroduce it, but for now it's more complexity than I think we need. Out of order messages recorded in chunks, are still played back in time-sorted order.


wkalt:
with this change, it's roughly as fast for me to `mcap cat` a file out of GCS as it is to `gsutil cat` it, which is as good as we can hope for.

Issue #354, opened 2022-03-30
go "make test" doesn't pass


jameskuszmaul-brt:
**Description**

No clue if this is even meant to be supported, since the CI seems to run the tests in `go/conformance` indirectly through some typescript scripts, but since there are Makefiles with test targets in `go/`, I would've expected `cd go/conformance; make test` to work, but instead the tests appear to fail.

jtbandes:
Can you share what errors you got? I just tried this:

```
$ make test
make -C test-streamed-write-conformance test
go test ./...
ok      github.com/foxglove/mcap/go/conformance/test-streamed-write-performance 0.154s
make -C test-streamed-read-conformance test
go test ./...
ok      github.com/foxglove/mcap/go/conformance/test-streamed-read-conformance  0.186s
```

jameskuszmaul-brt:
Oh it's `git-lfs` nonsense. Still not used to having repos that use `git-lfs`, so I'm not used to pulling that first....

PR #355, opened 2022-03-30
Add note about git-lfs to README

jameskuszmaul-brt:
Add note about git-lfs to Developer section of README. This might've helped with #354.


defunctzombie (approved):

defunctzombie:
@jtbandes @foxymiles I think CI is failing here because the PR is from an external contributor rather than repository collaborator and this does not have access to repository secrets. The TestPyPi step fails without access to the secret. Do we want to keep that as a part of the CI for python or only run it on main once merged?

jtbandes:
Another option is to do something [like this](https://github.community/t/make-secrets-available-to-builds-of-forks/16166/21) and only run the TestPyPI step if the head repo is the upstream repo (not a fork). That would not catch any breakages to the TestPyPI step on PRs from forks (probably rarely/never something we would need to worry about), but it would otherwise allow us to pass CI on PRs from forks.

defunctzombie:
What do you think we should do here? I'd like to merge this and also make it easy for others to contribute PRs for merging. Right now the CI is preventing that.

foxymiles:
@defunctzombie This should be fixed by https://github.com/foxglove/mcap/pull/358

PR #358, opened 2022-03-31
Don't run PyPi CI steps on forks

foxymiles:
**Public-Facing Changes**
This skips the TestPyPi and PyPi CI steps for forks

**Description**
This skips the PyPI related Python CI steps for forks so that the publishing secrets are not necessary for CI to succeed for forks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,7 +184,8 @@ jobs:
       - run: yarn workspace @mcap/core test
 
       - name: Publish to NPM
-        if: ${{ startsWith(github.ref, 'refs/tags/releases/typescript/core/v') }}
+        if:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -234,7 +235,7 @@ jobs:
           skip_existing: true
 
       - name: Publish to PyPI
-        if: ${{ startsWith(github.ref, 'refs/tags/releases/python/v') }}
+        if: ${{ !github.event.pull_request.head.repo.fork && startsWith(github.ref, 'refs/tags/releases/python/v') }}
```
jtbandes, line 238: Don't think it's necessary here, but I guess it doesn't hurt.

jtbandes (approved):

PR #361, opened 2022-04-04
add more links and version badges to readme for each language

jtbandes:
**Public-Facing Changes**
None

**Description**
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,13 @@ MCAP is a modular container format and logging library for pub/sub messages with
 
 MCAP libraries are provided in the following languages. For guidance on each language, see its corresponding README:
 
-- [C++](./cpp)
-- [Go](./go)
-- [Python](./python)
-- [TypeScript](./typescript)
-- [Swift](./swift)
+| Language              | Readme                 | API docs                                                        | Package name | Version                                                                              |
+| --------------------- | ---------------------- | --------------------------------------------------------------- | ------------ | ------------------------------------------------------------------------------------ |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,11 +12,13 @@ MCAP is a modular container format and logging library for pub/sub messages with
...
 
-- [C++](./cpp)
-- [Go](./go)
-- [Python](./python)
-- [TypeScript](./typescript)
-- [Swift](./swift)
+| Language              | Readme                 | API docs                                                        | Package name | Version                                                                              |
+| --------------------- | ---------------------- | --------------------------------------------------------------- | ------------ | ------------------------------------------------------------------------------------ |
+| C++                   | [readme](./cpp)        | [API docs](https://mcap.dev/docs/cpp)                           | `mcap`       | [![](https://shields.io/conan/v/mcap)](https://conan.io/center/mcap)                 |
+| Go                    | [readme](./go)         | [API docs](https://pkg.go.dev/github.com/foxglove/mcap/go/mcap) |              | see [releases](https://github.com/foxglove/mcap/releases)                            |
```
jtbandes, line 18: @wkalt wasn't sure if I should link to releases here, or is https://pkg.go.dev/github.com/foxglove/mcap/go/mcap?tab=versions a better link?

wkalt: @jtbandes I'm not sure if "Version" is really capturing what's in this column. Seems like if we weren't linking the generated docs under API docs, that would be the thing to stick in here as it's most consistent with the other languages. What if the column were called "Latest" and this linked to the same place as API docs for Go?

No super strong feelings though...

PR #362, opened 2022-04-04
update includes for cstdio, use ftell/fseek instead of ftello/fseeko

jtbandes:
**Public-Facing Changes**
None

**Description**
Attempting to fix build failure on Windows from https://github.com/conan-io/conan-center-index/pull/10104#issuecomment-1088038052

### DIFF
--- a/cpp/mcap/include/mcap/reader.inl
+++ b/cpp/mcap/include/mcap/reader.inl
@@ -42,16 +42,16 @@ Status BufferReader::status() const {
 
 // FileReader //////////////////////////////////////////////////////////////////
 
-FileReader::FileReader(FILE* file)
+FileReader::FileReader(std::FILE* file)
     : file_(file)
     , size_(0)
     , position_(0) {
   assert(file_);
 
   // Determine the size of the file
-  fseeko(file_, 0, SEEK_END);
-  size_ = ftello(file_);
-  fseeko(file_, 0, SEEK_SET);
+  std::fseek(file_, 0, SEEK_END);
+  size_ = std::ftell(file_);
+  std::fseek(file_, 0, SEEK_SET);
 }
 
 uint64_t FileReader::size() const {
@@ -64,8 +64,8 @@ uint64_t FileReader::read(std::byte** output, uint64_t offset, uint64_t size) {
   }
 
   if (offset != position_) {
-    fseeko(file_, offset, SEEK_SET);
-    fflush(file_);
+    std::fseek(file_, offset, SEEK_SET);
+    std::fflush(file_);
     position_ = offset;
   }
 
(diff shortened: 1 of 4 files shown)


cpp/mcap/include/mcap/reader.inl
```diff
@@ -64,16 +64,16 @@ uint64_t FileReader::read(std::byte** output, uint64_t offset, uint64_t size) {
   }
 
   if (offset != position_) {
-    fseeko(file_, offset, SEEK_SET);
-    fflush(file_);
+    std::fseek(file_, offset, SEEK_SET);
```
jhurliman, line 67: Will this limit the file sizes we can handle?

jtbandes: I dunno, it seems like it depends. We weren't defining `_FILE_OFFSET_BITS` so I'm not sure that `fseeko` would have been any better than `fseek`?

Also, I'm not sure that `long` is really a limitation, it seems to be 8 bytes on common platforms: [godbolt link](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEIADMpAAOqAqETgwe3r7%2ByanpAiFhkSwxcYl2mA4ZQgRMxARZPn4BldUCtfUERRHRsQm2dQ1NOa1D3aG9pf3xAJS2qF7EyOwcoQQA1CxMoRCzGyYA7FYaAIIbFxvEmARLDBtpAF6YqFQQtALAsybxJ6dHABEOPNaJwAKy8PwcLSkVCcNzWawPRbLTAHMzxHikAiaYHzADWIDBGn0nEkkNxsM4vAUIBJOOhwNIcFgMEQKFQLCSdFi5EoaC5PLiyAMRikGhJNFoBFitIgUUpUVC9QAnpwsQK2IIAPIMWhqxmkLDbIziQ34a7VABumFphswqiqXhllPWmFBhtoeCixFVHiwlIIxDwLHVTKoBmACgAanhMAB3bVJRhhmSCERidhSNPyJRqSm6Lj6QzGRGWfTe2mQeaoJKOAR2mnuqr1vwQVwjPxF4KTEplPQpNKtzsD/Ktnp9/pFtqtzrDTzNPQzmrjCd9OLT8YjzddNfTDfzBQolZ6IOYVY8EHgimGuEcVQADgAbABaJ%2BSDYiksbKQAOg0/4bBACKWNYpAbLghAkOimLgR4gr0MQMFcLMvAMloswEkSJIeuSpChsSpBQjCd40nS2K4vMLLsosBBJM6fIQAK3KIXomD4EQG45hm4jZvwgiKCo6iGmeF68PGvpJGGV4cBCRGUne2rOvRmyvBsj6vu%2Bn6isAP6SP%2BgHAZyLGxMhqEUYymGkAgmBMFgcR7KSHB4QRJLEbwpG2OR6F4qQhKER68Q3iR1IWRhMlmMFHmhT5Vk2sQaTOJIQA)

I guess if we run into issues with this we can start adding some platform-specific calls: https://stackoverflow.com/a/8708733

PR #369, opened 2022-04-11
cli: fix `mcap cat` on files without summary/statistics

jtbandes:
**Public-Facing Changes**
Fixed `mcap cat` crashing on files without summary/statistics information.

**Description**
Before:
```
$ mcap cat ../../../tests/conformance/data/OneMessage/OneMessage.mcap   
2022/04/11 13:05:15 Failed to read next message: failed to get next token: failed to allocate 74886844961080141 bytes for unrecognized opcode token: length out of int32 range
$ mcap info ../../../tests/conformance/data/OneMessage/OneMessage.mcap
2022/04/11 13:07:57 Failed to read file ../../../tests/conformance/data/OneMessage/OneMessage.mcap: failed to get info: failed to get next token: failed to allocate 74886844961080141 bytes for unrecognized opcode token: length out of int32 range
```

After:
```
$ go run main.go cat ../../../tests/conformance/data/OneMessage/OneMessage.mcap
$ go run main.go info ../../../tests/conformance/data/OneMessage/OneMessage.mcap
library: 
profile: 
channels:
attachments: 0
```

### DIFF
--- a/go/cli/mcap/cmd/info.go
+++ b/go/cli/mcap/cmd/info.go
@@ -31,18 +31,20 @@ func printInfo(w io.Writer, info *mcap.Info) error {
 	buf := &bytes.Buffer{}
 	fmt.Fprintf(buf, "library: %s\n", info.Header.Library)
 	fmt.Fprintf(buf, "profile: %s\n", info.Header.Profile)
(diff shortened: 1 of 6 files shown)


wkalt (approved):
we should add some testing for the output formatting, but it's a preexisting issue and we don't need to tackle it here.

Issue #387, opened 2022-05-11
Install mcap CLI tool on Linux


Dragonking3x:
I could not find an installation description for the [mcap CLI tool ](https://github.com/foxglove/mcap/releases) on Linux.

The installation of the tool goes like this:

**Download the file:**
mcap-linux-amd64
`sudo wget https://github.com/foxglove/mcap/releases/latest/download/mcap-linux-amd64 -O /usr/bin/mcap`

mcap-linux-arm64
`sudo wget https://github.com/foxglove/mcap/releases/latest/download/mcap-linux-arm64 -O /usr/bin/mcap`

**set permission to executable**
sudo chmod +x /usr/bin/mcap

PR #388, opened 2022-05-11
Document CLI installation

wkalt:
Adds some notes about installation of the CLI tool from releases and
source. Also adds a note linking the CLI to the main README, and changes
the link for the go library to point at the library instead of the
parent folder.


wkalt:
Addresses https://github.com/foxglove/mcap/issues/387

jtbandes (commented):
I wonder if there's more we can do to make this easily discoverable from mcap.dev?

PR #391, opened 2022-05-12
Add `reverse` option to Mcap0IndexedReader#readMessages

defunctzombie:
**Public-Facing Changes**

A new `reverse` option for the `readMessages` method of Mcap0IndexedReader. The reverse option emits messages in reverse logTime order (i.e. older messages first).

`startTime` and `endTime` arguments are _not_ reversed when using the `reverse` option. Only the order of the messages produced from the iterator.

**Description**

Providing a reverse iterator allows users to lookup the latest messages on a topic. This is useful when building a file player and wanting to find the latest available message on all requested topics when seeking to arbitrary locations in an indexed file.

### DIFF
--- a/typescript/core/src/v0/ChunkCursor.ts
+++ b/typescript/core/src/v0/ChunkCursor.ts
@@ -60,6 +60,10 @@ export class ChunkCursor {
    */
   compare(other: ChunkCursor): number {
     const reverse = this.reverse;
+    if (this.reverse !== other.reverse) {
+      throw new Error("Cannot compare a reversed ChunkCursor to a non-reversed ChunkCursor");
+    }
+
     if (reverse) {
       // If chunks don't overlap at all, sort later chunk first
       if (this.chunkIndex.messageEndTime < other.chunkIndex.messageStartTime) {
--- a/typescript/core/src/v0/ChunkCursor.ts
+++ b/typescript/core/src/v0/ChunkCursor.ts
@@ -281,7 +281,7 @@ export class ChunkCursor {
 
       let startIndex = 0;
       if (reverse) {
-        if (this.endTime) {
+        if (this.endTime != undefined) {
           startIndex = sortedIndexBy(
             result.record.records,
             [this.endTime],
@@ -289,7 +289,7 @@ export class ChunkCursor {
           );
         }
       } else {
-        if (this.startTime) {
+        if (this.startTime != undefined) {
           startIndex = sortedIndexBy(
             result.record.records,
             [this.startTime],
--- a/typescript/core/src/v0/ChunkCursor.ts
+++ b/typescript/core/src/v0/ChunkCursor.ts
@@ -9,6 +9,16 @@ type ChunkCursorParams = {
   relevantChannels: Set<number> | undefined;
   startTime: bigint | undefined;
   endTime: bigint | undefined;
+  reverse: boolean;
+};
+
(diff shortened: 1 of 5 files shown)


typescript/core/src/v0/ChunkCursor.ts
```diff
@@ -54,12 +59,23 @@ export class ChunkCursor {
    * and re-sort the cursors.
    */
   compare(other: ChunkCursor): number {
-    // If chunks don't overlap at all, sort earlier chunk first
-    if (this.chunkIndex.messageEndTime < other.chunkIndex.messageStartTime) {
-      return -1;
-    }
-    if (this.chunkIndex.messageStartTime > other.chunkIndex.messageEndTime) {
-      return 1;
+    const reverse = this.reverse;
```
jtbandes, line 62: What about `other.reverse`?

typescript/core/src/v0/Mcap0IndexedReader.test.ts
```diff
@@ -245,14 +245,35 @@ describe("Mcap0IndexedReader", () => {
...
+        endTime: 11n,
+        expected: [message2, message1],
+        reverse: true,
+      },
+      { startTime: 10n, endTime: 12n, expected: [message3, message2, message1], reverse: true },
     ])(
-      "fetches chunk data and reads requested messages between $startTime and $endTime",
-      async ({ startTime, endTime, expected }) => {
+      "fetches chunk data and reads requested messages between $startTime and $endTime, reverse: $reverse",
+      async ({ startTime, endTime, expected, reverse }) => {
```
jtbandes, line 276: I would remove reverse from the test params, and just run two `expect`s within each test. The `expected` result when reverse=true should always be reversed from the original expected result.

jtbandes (approved):
Looks pretty good, left some notes for cleanup.

typescript/core/src/v0/ChunkCursor.ts
```diff
@@ -87,12 +103,22 @@ export class ChunkCursor {
...
+      // Break ties by chunk offset in the file
+      return Number(other.chunkIndex.chunkStartOffset - this.chunkIndex.chunkStartOffset);
+    } else {
+      if (logTimeA !== logTimeB) {
+        return Number(logTimeA - logTimeB);
+      }
+
+      // Break ties by chunk offset in the file
+      return Number(this.chunkIndex.chunkStartOffset - other.chunkIndex.chunkStartOffset);
+    }
```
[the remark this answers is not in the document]
defunctzombie: Not quite. There's this bit of code that does not change under the reverse flag:

```
  // If a cursor has not loaded message indexes, sort it first so it can get loaded and re-sorted.
  // If it overlaps with other cursors, we won't know its order relative to other cursors until we
  // have the message indexes.
 if (!this.messageIndexCursors) {
      return -1;
    }
    if (!other.messageIndexCursors) {
      return 1;
    }
```

typescript/core/src/v0/ChunkCursor.ts
defunctzombie: 🤔 Good catch but I have no idea. Maybe throw if the two don't match? It's an anachronism in the current IndexedReader implementation to compare two such cursors.

jtbandes: I agree, throwing seems reasonable

PR #411, opened 2022-06-09
Move well known spec values into appendix

amacneil:
**Public-Facing Changes**
- Correct well known values (many field names were incorrect) and move into single appendix.
- Moved `explanatory notes` up a level since it was the only file in that directory.

**Motivation**

Reduce confusion for new users between well known message encodings and schema encodings, by making it clear in one appendix what the various well-known options are and how they should be used. Reduce complexity for readers with fewer folders and files to read.


docs/specification/appendix.md
```diff
@@ -0,0 +1,111 @@
+# MCAP File Format Specification Appendix
+
+This document describes well known values for MCAP files.
+
+## Well-known compression formats
+
+The Chunk `compression` field may contain the following options:
+
+- [lz4](<https://en.wikipedia.org/wiki/lz4_(compression_algorithm)>): an algorithm that prioritizes compression/decompression speed over compression ratio.
+- [zstd](https://en.wikipedia.org/wiki/zstandard): an algorithm that prioritizes compression ratio over compression/decompression speed.
```
jhurliman, line 10: I realize this text has probably read this way for a while, but I think a better description for zstd is “an algorithm with a tunable compression ratio/speed tradeoff”.

jhurliman (approved):

docs/specification/appendix.md
defunctzombie, line 10:
```suggestion
- [zstd](https://en.wikipedia.org/wiki/zstandard): an algorithm with a tunable compression ratio/speed tradeoff.
```

defunctzombie (approved):

Issue #351, opened 2022-03-29
MCAP filter cli


amacneil:
Ability to filter mcap files, similar to `rosbag filter`:

- include/exclude topics (glob/regex or just static list?)
- start/end time

james-rms:
What if instead we implemented a filter-friendly API in Python, and delivered a `mcap_filter.py` wrapper for simple CLI jobs?

Here's my thought process:
1. robot data manipulation tasks exist on a smooth complexity continuum, from:
  a. simple filtering: select all messages by channel or time range
  b. more complex filtering: filter by a field in a message, like `frame_id` in a sensor msg or `hardware_id` in `/diagnostics`
  c. transformation: reassign all images to one topic, fix timestamps, etc.
  ideally users can use tools from the same ecosystem for all of the above.
2. Python is a super popular language for data wrangling, in the ROS ecosystem especially but also more generally.
3. Building a nice python API for mcap manipulation will probably be really nice for users.
4. Once that's done, it's easy to implement a command-line filter tool in Python, which would also double as an example script for users who want to dip their toes into mcap manipulation in Python.

amacneil:
I agree with the use cases you outlined. However, our CLI is very easy to install (`brew install mcap`), and at the very least there should be an easy way to split out a single topic or time range (since we already have the ability to merge files from the CLI - seems weird if I can't do the reverse).

Landing more advanced filter features in the python library might be a better place to do it (and improving our python library in general is good). But honestly I think we should ignore more advanced filtering for now until there are requests for it - plenty of other things to work on.

james-rms:
This was completed in https://github.com/foxglove/mcap/pull/445

PR #412, opened 2022-06-09
CLI: Add split and merge subcommand

wkalt:
Adds subcommands "split" and "merge". Split splits an mcap file into one
output file per channel. Merge merges a set of mcap files into a single
output. For instance,

    mcap split demo.mcap -o output-dir
    mcap merge output-dir/* > output.mcap

Output of merge can be piped to other tools, e.g

    mcap merge output-dir/* | mcap cat --json

### DIFF
--- a/go/cli/mcap/cmd/merge.go
+++ b/go/cli/mcap/cmd/merge.go
@@ -19,6 +19,7 @@ var (
 	mergeChunkSize   int64
 	mergeIncludeCRC  bool
 	mergeChunked     bool
+	mergeOutputFile  string
 )
 
 type mergeOpts struct {
@@ -239,8 +240,8 @@ var mergeCmd = &cobra.Command{
 	Use:   "merge file1.mcap [file2.mcap] [file3.mcap]...",
 	Short: "Merge a selection of mcap files by record timestamp",
 	Run: func(cmd *cobra.Command, args []string) {
-		if !utils.StdoutRedirected() {
-			die("Binary output can screw up your terminal. Redirect output to a file or pipe.")
+		if mergeOutputFile == "" && !utils.StdoutRedirected() {
+			die("Binary output can screw up your terminal. Supply -o or redirect to a file or pipe.")
 		}
 		var readers []io.Reader
 		for _, arg := range args {
@@ -259,7 +260,18 @@ var mergeCmd = &cobra.Command{
 			chunked:     mergeChunked,
 		}
 		merger := newMCAPMerger(opts)
-		err := merger.mergeInputs(os.Stdout, readers)
+		var writer io.Writer
+		if mergeOutputFile == "" {
+			writer = os.Stdout
+		} else {
+			f, err := os.Create(mergeOutputFile)
+			if err != nil {
+				die("failed to open output file %s: %s\n", mergeOutputFile, err)
+			}
+			defer f.Close()
+			writer = f
+		}
+		err := merger.mergeInputs(writer, readers)
 		if err != nil {
 			die(err.Error())
 		}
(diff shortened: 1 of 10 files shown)


wkalt:
I added split here mostly to easily test the functionality interactively, but it also occurred to me we could wrap split as a CLI --separate-outputs flag on the filter subcommand requested here: https://github.com/foxglove/mcap/issues/351

This would more naturally allow extension to do stuff like split out only certain topics, or all but certain. Once we add that I think split will start looking pretty similar to filter.

go/cli/mcap/cmd/merge.go
```diff
@@ -0,0 +1,302 @@
...
+			return fmt.Errorf("failed to add initial schema for input %d: %w", inputID, err)
+		}
+		newChannelID, err := m.addChannel(writer, inputID, channel)
+		if err != nil {
+			return fmt.Errorf("failed to add initial channel for input %d: %w", inputID, err)
+		}
+		iterators[inputID] = iterator
+
+		// use the mapped output channel ID when pushing to the heap
+		message.ChannelID = newChannelID
```
wkalt, line 179: that can be assigned on 172

go/cli/mcap/cmd/merge.go
```diff
@@ -0,0 +1,302 @@
...
+		// the message to be written. This is numbered with the correct channel
+		// ID for the output, and schemas + channels for it have already been
+		// written, so it can be written straight to the output.
+		msg := heap.Pop(pq).(utils.TaggedMessage)
+		err = writer.WriteMessage(msg.Message)
+		if err != nil {
+			return fmt.Errorf("failed to write message: %w", err)
+		}
+
+		// the next message off the iterator, to replace the one just popped
```
wkalt, line 196: "Pull the next..."

amacneil:
Seems like both commands should support the `-o` flag (if passed to `merge`, it would write to the specified file rather than stdout)

go/cli/mcap/utils/priority_queue.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type PriorityQueue []TaggedMessage
+
+func (pq PriorityQueue) Len() int {
+	return len(pq)
+}
+
+func (pq PriorityQueue) Less(i, j int) bool {
+	return pq[i].Message.LogTime < pq[j].Message.LogTime
```
wkalt, line 16: this should do a secondary comparison on import ID so ties are broken deterministically

wkalt:
the formatting of the `split` output file names should be noted somewhere.. it's `<topic>_channelID.mcap`, where the topic has slashes converted to double underscores. Some other ideas would be,
* group files by schema instead of channel (I think bad, because the schema name will not be what users want to look for)
* group all channels for the same topic name into the same file, and name it `<topic>.mcap` with the same transformation
* prefix the topic names with the channel ID instead of suffixing them. Maybe nicer filenames, but also don't sort on topic name.

go/cli/mcap/cmd/merge.go
```diff
@@ -0,0 +1,302 @@
...
+	// for each input reader, initialize an mcap reader and read the first
+	// message off. Insert the schema and channel into the output with
+	// renumbered IDs, and load the message (with renumbered IDs) into the
+	// priority queue.
+	for inputID, inputReader := range inputs {
+		reader, err := mcap.NewReader(inputReader)
+		if err != nil {
+			return fmt.Errorf("failed to create input reader for input %d: %w", inputID, err)
+		}
+		iterator, err := reader.Messages(0, math.MaxInt64, nil, false)
```
wkalt, line 156: instantiation of the reader/iterator can probably be pulled into a method

amacneil:
What if you let the user specify the file naming. e.g.:

```sh
mcap split demo.mcap -o "outputdir/{channel_topic}_{channel_id}.mcap"
```

wkalt:
closing - work moved/merged in #418

PR #422, opened 2022-06-12
Simplify github pages build

amacneil:
**Public-Facing Changes**
None

**Description**
I wanted to add an image to the homepage, but the included files were hardcoded in a bunch of places. So I simplified the build slightly by just copying the entire website directory into the `gh-pages` branch.


defunctzombie (approved):

amacneil:
wow so fast... still testing but I think it's working ok

defunctzombie:
> wow so fast... still testing but I think it's working ok

Wasn't a draft PR so I assumed it was ready :D

jtbandes:
Which hardcoded places were you talking about? I think this change is fine, but I'm not sure I see what becomes easier — are you talking about the `cp -R docs/gh-pages/*.html docs/gh-pages/dist __pages__` that was removed?

amacneil:
Yeah - now there is an image in that directory too. So I guess I could have added `*.png` to the script, but it seemed easier if the whole directory just got copied.

Issue #419, opened 2022-06-09
CLI: JSON transcoder does not properly escape string data


wkalt:
The ROS to JSON transcoder needs to implement logic equivalent to that of the stringBytes function in https://go.dev/src/encoding/json/encode.go when converting ROS strings to JSON strings. Currently it is just quoting, which results in the conversion failing if the string contains data that must be JSON escaped.

PR #433, opened 2022-06-22
cli: fix json transcoding in cat

jtbandes:
**Public-Facing Changes**
Fixed a crash in `mcap cat --json` when printing string values that had special characters.

**Description**
Fixes https://github.com/foxglove/mcap/issues/419

### DIFF
--- a/go/cli/mcap/utils/ros/json_transcoder.go
+++ b/go/cli/mcap/utils/ros/json_transcoder.go
@@ -3,6 +3,7 @@ package ros
 import (
 	"encoding/base64"
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"io"
 	"math"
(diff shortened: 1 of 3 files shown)


go/cli/mcap/utils/ros/json_transcoder.go
```diff
@@ -202,17 +202,16 @@ func (t *JSONTranscoder) string(w io.Writer, r io.Reader) error {
...
-	n, err := io.ReadFull(r, t.buf[1:length-1])
+	_, err = io.ReadFull(r, t.buf[:length])
 	if err != nil {
 		return err
 	}
-	t.buf[0] = '"'
-	t.buf[n+1] = '"'
-	_, err = w.Write(t.buf[:n+2])
+	enc := json.NewEncoder(w)
+	err = enc.Encode(string(t.buf[:length]))
```
jtbandes, line 214: introduces a copy of the bytes... sounds like there's not really a way to avoid this if converting from `byte[]` to `string`?

wkalt: yeah, that's correct - cause strings are immutable. Looks good to me.

PR #441, opened 2022-06-23
cli: parse fields with tab after name

james-rms:
Previously the parser assumed that any comment after a field name would
come after a space, like this:

```
Type name # the comment
```

However, there is at least one standard msg definition which breaks this
rule, `ns` in ImageMarker:
https://docs.ros.org/en/noetic/api/visualization_msgs/html/msg/ImageMarker.html

```
string ns\t# namespace, used with id to form a unique id
```

This commit uses the field name description in http://wiki.ros.org/msg#Fields to determine which chars contain the field name.

**Public-Facing Changes**
This slightly changes the behaviour of `mcap cat --json`, which attempts to parse and JSON-encode ROS msg definitions.

### DIFF
--- a/go/ros/ros1msg/ros1msg_parser.go
+++ b/go/ros/ros1msg/ros1msg_parser.go
@@ -2,11 +2,16 @@ package ros1msg
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
 	"github.com/foxglove/mcap/go/ros"
 )
+
+// Field names are restricted to "an alphabetical character followed by any mixture of alphanumeric and underscores",
+// per http://wiki.ros.org/msg#Fields
+var fieldMatcher = regexp.MustCompile(`([^ ]+) +([a-zA-Z][a-zA-Z0-9_]+)`)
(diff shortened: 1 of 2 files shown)


go/ros/ros1msg/ros1msg_parser.go
```diff
@@ -2,12 +2,16 @@ package ros1msg
...
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
 	"github.com/foxglove/mcap/go/ros"
 )
 
+// Field names are restricted to [a-zA-Z][a-zA-Z1-9_]+, per http://wiki.ros.org/msg#Fields
+var fieldMatcher = regexp.MustCompile(`([^ ]+) ([a-zA-Z][a-zA-Z1-9_]+)`)
```
wkalt, line 13: 1-9 not 0-9? Can't see a source

james-rms: seems like a typo on the wiki.

wkalt (approved):

PR #442, opened 2022-06-24
CLI: Make release notes grep case insensitive

wkalt:
Changes the release script to use a case-insensitive filter for "CLI"
when compiling commit summaries for the GH release notes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -319,7 +319,7 @@ jobs:
           OUTPUT: mcap-${{ matrix.os }}-${{ matrix.arch }}
       - name: Make release notes
         run: |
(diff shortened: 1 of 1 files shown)


jtbandes:
Would it make sense to run log on just the cli directory, rather than using grep to determine which changes affect the cli?

wkalt:
@jtbandes yeah, I think that's an improvement over the current approach

jhurliman (approved):

defunctzombie:
Do we still want this?

jtbandes:
@wkalt Any interest in pursuing this further?

defunctzombie:
Closing this since there's been no response.

PR #444, opened 2022-06-27
CLI: use platform-dependent path separators

james-rms:
**Public-Facing Changes**
Windows users can use `--ament-prefix-path` with a native path separator, and not worry about drive letters `C:\` getting in the way.

**Description**
Currently `convert` assumes it's running on a unix-like where the path list separator is `:`, but on Windows this doesn't work. This PR uses the platform's path separator.


wkalt (approved):
thanks, looks good.

jtbandes (approved):

PR #445, opened 2022-06-27
CLI: mcap filter subcommand

james-rms:
**Public-Facing Changes**
* New `mcap filter` subcommand added, which you can use to filter MCAPs by topic and time range.
* Bug fixed in the go MCAP library when writing attachment indexes, which would cause the attachment
  content type to be malformed.

**Description**
This PR adds a new subcommand to the MCAP CLI tool that allows people to quickly filter topics into and out-of an MCAP file.

### DIFF
--- a/go/cli/mcap/cmd/filter.go
+++ b/go/cli/mcap/cmd/filter.go
@@ -80,12 +80,12 @@ func buildFilterOptions() (*filterOpts, error) {
 
 // filterCmd represents the filter command
 var filterCmd = &cobra.Command{
-	Use:   "filter",
+	Use:   "filter [file]",
 	Short: "copy some filtered MCAP data to a new file",
 	Long: `This subcommand filters an MCAP by topic and time range to a new file.
 When multiple regexes are used, topics that match any regex are included (or excluded).
 
-usage example:
+usage:
   mcap filter in.mcap -o out.mcap -y /diagnostics -y /tf -y /camera_(front|back)`,
 	Run: func(cmd *cobra.Command, args []string) {
 		filterOptions, err := buildFilterOptions()
@@ -101,7 +101,7 @@ usage example:
 			if stat.Mode()&os.ModeCharDevice == 0 {
 				reader = os.Stdin
 			} else {
-				die("supply a file to read")
+				die("please supply a file. see --help for usage details.")
 			}
 		} else {
 			close, newReader, err := utils.GetReader(context.Background(), args[0])
@@ -224,19 +224,17 @@ func filter(
 			if err != nil {
 				return err
 			}
-			if len(opts.includeTopics) > 0 {
-				for _, matcher := range opts.includeTopics {
-					if matcher.MatchString(channel.Topic) {
-						channels[channel.ID] = markableChannel{channel, false}
-					}
+			for _, matcher := range opts.includeTopics {
+				if matcher.MatchString(channel.Topic) {
+					channels[channel.ID] = markableChannel{channel, false}
 				}
-			} else if len(opts.excludeTopics) > 0 {
-				for _, matcher := range opts.excludeTopics {
-					if !matcher.MatchString(channel.Topic) {
-						channels[channel.ID] = markableChannel{channel, false}
-					}
+			}
+			for _, matcher := range opts.excludeTopics {
+				if !matcher.MatchString(channel.Topic) {
+					channels[channel.ID] = markableChannel{channel, false}
 				}
-			} else {
+			}
+			if len(opts.includeTopics) == 0 && len(opts.excludeTopics) == 0 {
 				channels[channel.ID] = markableChannel{channel, false}
 			}
 		case mcap.TokenMessage:
--- a/go/cli/mcap/cmd/filter.go
+++ b/go/cli/mcap/cmd/filter.go
@@ -1,0 +1,317 @@
+package cmd
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"os"
+	"regexp"
+
+	"github.com/foxglove/mcap/go/cli/mcap/utils"
+	"github.com/foxglove/mcap/go/mcap"
+	"github.com/spf13/cobra"
+)
+
+var (
+	filterOutput             string
+	filterIncludeTopics      []string
+	filterExcludeTopics      []string
+	filterStart              uint64
+	filterEnd                uint64
+	filterIncludeMetadata    bool
+	filterIncludeAttachments bool
+	filterOutputCompression  string
+)
+
+type filterOpts struct {
+	includeTopics      []regexp.Regexp
+	excludeTopics      []regexp.Regexp
+	start              uint64
+	end                uint64
+	includeMetadata    bool
+	includeAttachments bool
+	compressionFormat  mcap.CompressionFormat
+}
+
+func buildFilterOptions() (*filterOpts, error) {
+	opts := &filterOpts{
+		start:              filterStart,
+		end:                filterEnd,
+		includeMetadata:    filterIncludeMetadata,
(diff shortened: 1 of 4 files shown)


defunctzombie:
What is the syntax for filter? Some examples?

go/mcap/writer.go
```diff
@@ -284,7 +284,7 @@ func (w *Writer) WriteAttachmentIndex(idx *AttachmentIndex) error {
 	if w.opts.SkipAttachmentIndex {
 		return nil
 	}
-	msglen := 8 + 8 + 8 + 8 + 4 + len(idx.Name) + 4 + len(idx.ContentType)
+	msglen := 8 + 8 + 8 + 8 + 8 + 4 + len(idx.Name) + 4 + len(idx.ContentType)
```
wkalt, line 287: what caused this change?

wkalt: oh I see - it was just wrong. probably never exercised... :(

james-rms: I tried to explain it in the commit msg: https://github.com/foxglove/mcap/pull/445/commits/a518ee38568cb18c05eebe9dde7c893ace353cde

basically it's missing 8 bytes for one of the uint64s. Previously we didn't notice because `putStringPrefixed` silently truncates when there's not enough space, and all of out attachment content type strings were more than 8 characters.

go/cli/mcap/cmd/filter.go
```diff
@@ -0,0 +1,317 @@
...
+func init() {
+	rootCmd.AddCommand(filterCmd)
+
+	filterCmd.PersistentFlags().StringVarP(&filterOutput, "output", "o", "", "output filename")
+	filterCmd.PersistentFlags().StringArrayVarP(&filterIncludeTopics, "include-topic-regex", "y", []string{}, "messages with topic names matching this regex will be included")
+	filterCmd.PersistentFlags().StringArrayVarP(&filterExcludeTopics, "exclude-topic-regex", "n", []string{}, "messages with topic names matching this regex will be included")
+	filterCmd.PersistentFlags().Uint64VarP(&filterStart, "start-time-ns", "s", 0, "messages with log times after or equal to this timestamp will be included.")
+	filterCmd.PersistentFlags().Uint64VarP(&filterEnd, "end-time-ns", "e", math.MaxUint64, "messages with log times before timestamp will be included.")
+	filterCmd.PersistentFlags().BoolVar(&filterIncludeMetadata, "include-metadata", true, "whether to include metadata in the output bag")
+	filterCmd.PersistentFlags().BoolVar(&filterIncludeAttachments, "include-attachments", true, "whether to include attachments in the output mcap")
```
wkalt, line 315: I think we should probably default both include-metadata and include-attachments to false - since in the common case of filtering for a single topic, there's no logical reason why all your metadata/attachments should relate to that topic

james-rms:
@defunctzombie from the long help:
```
This subcommand filters an MCAP by topic and time range to a new file.
When multiple regexes are used, topics that match any regex are included (or excluded).
usage example:
  mcap filter in.mcap -o out.mcap -y /diagnostics -y /tf -y /camera_(front|back)
 ```

Some more examples:
```
# exclude topic by regex, using pipes
cat in.mcap | mcap filter -n /pose > out.mcap

# time ranges
mcap filter in.mcap -o out.mcap --start-secs 1657000000 --end-secs 1657030667
```

File: go/cli/mcap/cmd/filter.go [hunk 3]
```diff
@@ -0,0 +1,317 @@
...
+			}()
+			reader = newReader
+		}
+
+		var writer io.Writer
+		if filterOutput == "" {
+			if !utils.StdoutRedirected() {
+				die("Binary output can screw up your terminal. Supply -o or redirect to a file or pipe")
+			}
+			writer = os.Stdin
```
wkalt, line 120: os.Stdout

go/cli/mcap/cmd/filter.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+	}
+}
+
+func init() {
+	rootCmd.AddCommand(filterCmd)
+
+	filterCmd.PersistentFlags().StringVarP(&filterOutput, "output", "o", "", "output filename")
+	filterCmd.PersistentFlags().StringArrayVarP(&filterIncludeTopics, "include-topic-regex", "y", []string{}, "messages with topic names matching this regex will be included")
+	filterCmd.PersistentFlags().StringArrayVarP(&filterExcludeTopics, "exclude-topic-regex", "n", []string{}, "messages with topic names matching this regex will be included")
```
wkalt, line 311: maybe we can indicate somehow that multiple of these can be supplied? might need to futz with it to keep the wording short

go/cli/mcap/cmd/filter.go
```diff
@@ -0,0 +1,317 @@
...
+		return err
+	}
+	mcapWriter, err := mcap.NewWriter(w, &mcap.WriterOptions{
+		Compression: opts.compressionFormat,
+	})
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = mcapWriter.Close()
```
wkalt, line 181: I think this probably should either log the error here in the defer, or do a little more work to overwrite the returned error. Currently the error is just getting dropped.

To properly return the close error from the function you would want to change the function to return a named variable `(err error)` and then do some similar business to https://github.com/foxglove/data-platform/blob/main/stream-server/merging/mcap0_to_bag.go#L35,L45.

go/cli/mcap/cmd/filter.go
```diff
@@ -0,0 +1,317 @@
...
+			if err != nil {
+				return err
+			}
+			schemas[schema.ID] = markableSchema{schema, false}
+		case mcap.TokenChannel:
+			channel, err := mcap.ParseChannel(data)
+			if err != nil {
+				return err
+			}
+			if len(opts.includeTopics) > 0 {
```
wkalt, line 218: the length check here and on 224 is unnecessary - the iteration will not run if the length is zero

go/cli/mcap/cmd/filter.go
```diff
@@ -0,0 +1,326 @@
...
+				}
+			} else if len(opts.excludeTopics) > 0 {
+				for _, matcher := range opts.excludeTopics {
+					if !matcher.MatchString(channel.Topic) {
+						channels[channel.ID] = markableChannel{channel, false}
+					}
+				}
+			} else {
+				channels[channel.ID] = markableChannel{channel, false}
+			}
```
wkalt, line 241: if we write the schema+channel after observing the channel, rather than after observing the first message, would that let us drop the "markable" aspect of the channel map, or maybe even drop the channel map altogether?

That would also result in us returning channels with matching topics even if they have no messages on them, which seems more correct on balance to me. What do you think?

go/cli/mcap/cmd/filter.go
```diff
@@ -0,0 +1,326 @@
...
+		var reader io.Reader
+		if len(args) == 0 {
+			stat, err := os.Stdin.Stat()
+			if err != nil {
+				die("failed to check stdin state: %s", err)
+			}
+			if stat.Mode()&os.ModeCharDevice == 0 {
+				reader = os.Stdin
+			} else {
+				die("supply a file to read")
```
wkalt, line 104: maybe we can point them at `mcap filter -h` or say `mcap filter [file]` here somehow

wkalt (approved):
few high-level comments - feel free to address as you like. I pulled it down and tested, seems to work fine with the stdout change.

go/cli/mcap/cmd/filter.go
wkalt: well, a channel in an mcap with no messages is definitely something you can record - so why wouldn't the filter command preserve channels on the supplied topic?

PR #450, opened 2022-06-30
CLI: Don't require newline terminated message defs in db3 conversion

wkalt:
Prior to this commit, we would produce a malformed merged schema in db3
to mcap conversion if a component message definition did not terminate
in a newline. With this change we will insert newlines after message
definitions that do not end in them.

### DIFF
--- a/go/ros/ros2db3_to_mcap.go
+++ b/go/ros/ros2db3_to_mcap.go
@@ -72,6 +72,13 @@ func getSchemas(encoding string, directories []string, types []string) (map[stri
 		for len(subdefinitions) > 0 {
 			subdefinition := subdefinitions[0]
 			if !first {
+				// if the previous write did not end with a newline, add one now
+				if messageDefinition.Bytes()[messageDefinition.Len()-1] != '\n' {
+					err := messageDefinition.WriteByte('\n')
+					if err != nil {
+						return nil, fmt.Errorf("failed to write newline")
+					}
+				}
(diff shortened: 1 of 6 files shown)


go/cli/mcap/cmd/cat.go
```diff
@@ -152,7 +152,7 @@ func printMessages(
 		target.Data = msg.Bytes()
 		err = encoder.Encode(target)
 		if err != nil {
-			return fmt.Errorf("failed to write encoded message: %s", err)
+			return fmt.Errorf("failed to write encoded message %s: %s", string(msg.Bytes()), err)
```
wkalt, line 155: this should be removed, I must have started with a dirty branch

go/ros/ros2db3_to_mcap.go
```diff
@@ -72,6 +72,13 @@ func getSchemas(encoding string, directories []string, types []string) (map[stri
 		for len(subdefinitions) > 0 {
 			subdefinition := subdefinitions[0]
 			if !first {
+				// if the previous write did not end with a newline, add one now
+				if messageDefinition.Bytes()[messageDefinition.Len()-1] != '\n' {
+					_, err := messageDefinition.WriteString("\n")
```
jtbandes, line 77: nit: could use WriteByte since you're reading a byte on the prev line, doesn't really matter though

jtbandes (approved):

PR #451, opened 2022-06-30
Update CLI dependencies

wkalt:


wkalt:
Among other things, this will pick up the latest changes in the mcap and ros packages, which will give us support for non-newline terminated ROS2 schemas in mcap conversion, and for empty field names.

defunctzombie (approved):

PR #452, opened 2022-06-30
mcap: callback-based reader implementation

james-rms:
**Public-Facing Changes**

**Description**

### DIFF
--- a/go/cli/mcap/cmd/attachments.go
+++ b/go/cli/mcap/cmd/attachments.go
@@ -13,25 +13,33 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func printAttachments(w io.Writer, attachmentIndexes []*mcap.AttachmentIndex) {
+func printAttachments(w io.Writer, r io.Reader) error {
 	tw := tablewriter.NewWriter(w)
-	rows := make([][]string, 0, len(attachmentIndexes))
+	rows := make([][]string, 0)
 	rows = append(rows, []string{
 		"log time",
 		"name",
 		"content type",
 		"content length",
-		"offset",
 	})
-	for _, idx := range attachmentIndexes {
-		row := []string{
-			fmt.Sprintf("%d", idx.LogTime),
-			idx.Name,
-			idx.ContentType,
-			fmt.Sprintf("%d", idx.DataSize),
-			fmt.Sprintf("%d", idx.Offset),
-		}
-		rows = append(rows, row)
+
+	cbReader, err := mcap.NewCallbackReader(mcap.ReadOptions{
+		OnAttachment: func(a *mcap.Attachment) error {
+			row := []string{
+				fmt.Sprintf("%d", a.LogTime),
+				a.Name,
+				a.ContentType,
+				fmt.Sprintf("%d", len(a.Data)),
+			}
+			rows = append(rows, row)
+			return nil
+		},
+	})
+	if err != nil {
+		return err
+	}
+	if err := cbReader.Read(r); err != nil {
+		return err
 	}
 	tw.SetBorder(false)
 	tw.SetAutoWrapText(false)
@@ -40,6 +48,7 @@ func printAttachments(w io.Writer, attachmentIndexes []*mcap.AttachmentIndex) {
 	tw.SetColumnSeparator("")
 	tw.AppendBulk(rows)
 	tw.Render()
+	return nil
 }
 
 var attachmentsCmd = &cobra.Command{
@@ -52,16 +61,7 @@ var attachmentsCmd = &cobra.Command{
 		}
 		filename := args[0]
 		err := utils.WithReader(ctx, filename, func(matched bool, rs io.ReadSeeker) error {
-			reader, err := mcap.NewReader(rs)
-			if err != nil {
-				return fmt.Errorf("failed to get reader: %w", err)
-			}
-			info, err := reader.Info()
-			if err != nil {
-				return fmt.Errorf("failed to get info: %w", err)
-			}
-			printChunks(os.Stdout, info.ChunkIndexes)
-			return nil
+			return printAttachments(os.Stdout, rs)
 		})
 		if err != nil {
 			log.Fatal("Failed to list attachments: %w", err)
--- a/go/mcap/new_reader.go
+++ b/go/mcap/new_reader.go
@@ -1,0 +1,510 @@
+package mcap
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"sort"
+
+	"github.com/klauspost/compress/zstd"
+	"github.com/pierrec/lz4/v4"
+)
+
+type ReadOptions struct {
+	// optional callback which returns whether to include a given channel of messages in the results.
+	ChannelFilter func(schema *Schema, channel *Channel) bool
+	// optional callback which returns whether to include a given attachment in the results.
+	AttachmentFilter func(name string) bool
+	// optional callback which returns whether to include a given metadata record in the results.
+	MetadataFilter func(name string) bool
+	// Messages and Attachments with LogTimes in the range [StartTime, EndTime) will be included.
+	StartTime uint64
+	EndTime   uint64
+
+	// Callbacks to handle parsed records of interest. If a non-nil error is returned,
+	// reading will be aborted.
+	OnMessage    func(*Schema, *Channel, *Message) error
+	OnAttachment func(*Attachment) error
+	OnMetadata   func(*Metadata) error
+	OnStatistics func(*Statistics) error
+}
+
+type CallbackReader struct {
+	opts ReadOptions
+}
+
+func NewCallbackReader(opts ReadOptions) (*CallbackReader, error) {
+	if opts.OnMessage == nil && opts.ChannelFilter != nil {
+		return nil, fmt.Errorf("must supply OnMessage if ChannelFilter is supplied")
+	}
+	if opts.OnAttachment == nil && opts.AttachmentFilter != nil {
+		return nil, fmt.Errorf("must supply OnAttachment if AttachmentFilter is supplied")
+	}
+	if opts.OnMetadata == nil && opts.MetadataFilter != nil {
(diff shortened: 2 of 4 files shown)


go/mcap/new_reader.go
```diff
@@ -0,0 +1,503 @@
...
+
+	"github.com/klauspost/compress/zstd"
+	"github.com/pierrec/lz4/v4"
+)
+
+type MCAPReader struct {
+	// optional callback which returns whether to include a given channel of messages in the results.
+	ChannelFilter func(schema *Schema, channel *Channel) bool
+	// optional callback which returns whether to include a given attachment in the results.
+	AttachmentFilter func(name string) bool
```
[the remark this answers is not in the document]
james-rms: The point of the separate callback is that the filters can work before we even seek to the actual content of the record. This lets us decide not to decompress chunks that don't contain messages of interest.

go/mcap/new_reader.go
```diff
@@ -0,0 +1,503 @@
...
+	start() int64
+	len() int64
+}
+
+func (ci *ChunkIndex) start() int64      { return (int64)(ci.ChunkStartOffset) }
+func (ci *ChunkIndex) len() int64        { return (int64)(ci.ChunkLength + ci.MessageIndexLength) }
+func (ai *AttachmentIndex) start() int64 { return (int64)(ai.Offset) }
+func (ai *AttachmentIndex) len() int64   { return (int64)(ai.Length) }
+func (mi *MetadataIndex) start() int64   { return (int64)(mi.Offset) }
+func (mi *MetadataIndex) len() int64     { return (int64)(mi.Length) }
```
wkalt, line 286: I wonder if there's a better way we can handle this that would just result in failure in face of an out of range int64. FWIW uint64 support within the mcap tooling is a general weak spot. Even though the spec admits uint64 in many cases, all the golang stdlib functions for doing IO tend to take int64 and we haven't bothered to handle the accounting properly here, or I think on the TS side either.

Since the spec has uint64 seems right for our structs to contain them, but anywhere these functions lead to a downcast will probably end up breaking stuff.

go/mcap/new_reader.go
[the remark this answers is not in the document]
wkalt: ok, I see.

go/mcap/new_reader.go
```diff
@@ -0,0 +1,510 @@
...
+			}
+			attachment, err := ParseAttachment(record)
+			if err != nil {
+				return fmt.Errorf("failed to parse attachment record: %w", err)
+			}
+			if attachment.LogTime <= cbr.opts.StartTime || attachment.LogTime > cbr.opts.upperTimeBound() {
+				continue
+			}
+
+			if err := cbr.opts.OnAttachment(attachment); err != nil {
```
wkalt, line 149: I think the call to OnAttachment and probably OnMetadata too, needs to get pushed all the way down to the lexer and actually act against the underlying Reader (though the callback can still accept an Attachment type for ergonomics, if the Attachment type is changed to expose a reader instead of a byte slice).

The reason is, the point of the callback-based flow for handling attachments is to avoid materializing them in RAM if not needed, and by the time we are getting tokens out of the lexer the records have already been materialized. If you imagine someone has attached a 20GB file to the mcap, the lexer would probably consumed all the client's ram by materializing token.

I think this matters less/not at all with Message/Statistics.

wkalt: the behavior of the lexer would then be to stream copy the record to io.Discard if the callback is nil. I think that's what happens today for all attachments.

james-rms: Agreed on attachment. Really wrt. metadata? how much metadata are we expecting?

james-rms: Also - i want to do this in a separate PR, i can't figure out how to do that in this PR without totally rewriting parsing, or leaving it in a gross state.

wkalt: I agree metadata is also debatable, I think attachment is the one that definitely will require it.

wkalt: regarding second PR - I think it's fine for it to be in a separate PR, but I'm not sure if this one really stands alone without the change. I think the goal for this change is primarily to support handling of attachments in a streaming manner, with the callback API (for attachments at least) being an anticipated side effect of that, rather than the callback API being a worthy goal in itself. I am assuming in saying that that the end user is still going to end up with something that looks pretty similar to the iterator today, but perhaps implemented on top of these callbacks (btw - how exactly does this work? callbacks that pass messages over channels?)

If we merge the change to cut the reader over to all callbacks prior to streaming attachment support going in, then it seems like we're exposing a version of attachment support that won't work well on large attachments, which I'm not sure is better than just waiting for the streaming support.

I'm not sure about the parsing rewrite, or what counts as gross exactly.. but I think if we wanted to make the change for streaming in the most minimal way, the model to follow would be the way chunks get lexed here: https://github.com/foxglove/mcap/blob/main/go/mcap/lexer.go#L141,L147.

Like attachments, chunk lexing also needs to work against the reader directly unless the EmitChunks option gets specified. So we are checking for a chunk opcode prior to actually reading the body of whatever record we have in hand. Perhaps we need to expand that check for OpChunk into a switch statement that calls l.options.OnAttachment on any attachment record, the similarly falls through to `continue` to actually return the next emitted record from Next. The Attachment case for the lower switch statement would also be removed.

OnAttachment would be passed in as a new element of LexerOptions, with the default value being a callback that writes the record to io.Discard. It would probably get to the lexer through a new options mechanism for readers. https://golang.cafe/blog/golang-functional-options-pattern.html would be a nice pattern to follow for that - some day it would be nice to move the lexer options implementation over to that too.

I think the only piece of parsing this would break is ParseAttachment - which I think would probably either get removed completely or would only be called as an internal piece of the callback mechanism.

That would give us attachment support in unindexed reads, which I think is a good incremental change. It would still leave out,
* Support for attachment callbacks in indexed reads (using the attachment index)
* Support for metadata

Those pieces could be combined into the change or separately addressed; no particularly strong feelings on that.

Anyway, that's just the rough picture I had in my head. Not sure where it breaks down/differs from what you have in mind.

go/cli/mcap/cmd/attachments.go
```diff
@@ -13,25 +13,33 @@ import (
...
+			}
+			rows = append(rows, row)
+			return nil
+		},
+	})
+	if err != nil {
+		return err
+	}
+	if err := cbReader.Read(r); err != nil {
+		return err
```
wkalt, line 42: The CLI function for listing attachments shouldn't actually visit the attachments in the file. It should just read the attachment indexes from the summary section and get this info from there.

We will need support at some point to actually retrieve an attachment from the file, but these summary functions should only interact with the index - that's particularly important when reading a remote file.

go/mcap/new_reader.go
```diff
@@ -0,0 +1,510 @@
...
+type ReadOptions struct {
+	// optional callback which returns whether to include a given channel of messages in the results.
+	ChannelFilter func(schema *Schema, channel *Channel) bool
+	// optional callback which returns whether to include a given attachment in the results.
+	AttachmentFilter func(name string) bool
+	// optional callback which returns whether to include a given metadata record in the results.
+	MetadataFilter func(name string) bool
+	// Messages and Attachments with LogTimes in the range [StartTime, EndTime) will be included.
+	StartTime uint64
+	EndTime   uint64
```
wkalt, line 25: I think that aspirationally anyway, the reader should be designed to be reusable, in that I should be able to instantiate a reader and make multiple queries against it without needing to allocate new buffers. I think there are various complications with this today, but I think this API may add some more. Some considerations -
* Instantiation of the lexer is going to allocate some buffers, which the lexer will grow to size with use.
* Putting StartTime, EndTime, and the topic list  (via channel filter) into the constructor for the reader, rather than in the equivalent to the rosbag bag.Messages call, means that StartTime, EndTime, Topics have the same lifetime as the costly lexer.
* This causes issues for reuse because I need to make a new lexer every time I want to query different data in the file.

Would it make sense to pull StartTime, EndTime, and ChannelFilter (or some manifestation of) out of the ReadOptions, and instead stick those in the call to Read? And also, to attach the lexer to the reader so that it can be reused.

go/mcap/new_reader.go
```diff
@@ -0,0 +1,510 @@
...
+
+func (m *ReadOptions) upperTimeBound() uint64 {
+	if m.EndTime == 0 {
+		return math.MaxUint64
+	}
+	return m.EndTime
+}
+
+//
+func (cbr *CallbackReader) Read(reader io.Reader) error {
```
wkalt, line 60: it seems more logical to me that CallbackReader would be instantiated on an io.Reader, which would get wrapped in a lexer and attached to the returned CallbackReader, and that the userfacing API would then accept a start/end/topics (and options) rather than a reader.

The current implementation allows me to call Read with the same set of (start/stop/topics) against multiple input files. But I can't think when I would want that.  It seems more useful to me to be able to call Read with different sets of start/stop/topics, against the same input file.

go/cli/mcap/cmd/attachments.go
```diff
@@ -52,16 +61,7 @@ var attachmentsCmd = &cobra.Command{
...
-			if err != nil {
-				return fmt.Errorf("failed to get reader: %w", err)
-			}
-			info, err := reader.Info()
-			if err != nil {
-				return fmt.Errorf("failed to get info: %w", err)
-			}
-			printChunks(os.Stdout, info.ChunkIndexes)
-			return nil
+			return printAttachments(os.Stdout, rs)
```
wkalt, line 64: fwiw I think the previous implementation would be correct if we had instead called,

```
printAttachments(os.Stdout, info.AttachmentIndexes)
```

go/cli/mcap/cmd/attachments.go
james-rms: The trouble with this is that mcaps with no summary still exist and need to have accurate information reported. Maybe we should only do the full scan if we encounter no summary.

go/cli/mcap/cmd/attachments.go
james-rms: Agreed, except for the case where an mcap has no summary.

james-rms:
abandoning this approach in favour of #455

PR #455, opened 2022-07-04
go: generalize message iterator to include metadata and attachments

james-rms:
**Public-Facing Changes**
* Users of the go `mcap.Reader` API are affected. The `MessageIterator` interface is removed and replaced with `ContentIterator`, which generalises to reading messages, attachments and metadata.
* Users of the go `mcap.Lexer` API are affected. The lexer no longer fully materializes all tokens, and instead returns a `LimitedReader` for all token types. The user is not required to consume this themselves - if they don't, the Lexer cleans up after them.

**Description**
* This PR removes the `MessageIterator` interface and replaces it with a `ContentReader` interface. This returns `ContentRecord` pointers, which are a wrapper interface to wrap either a message, an attachment, or a metadata record.
* ContentReaders are constructed with the `reader.Content` function, which now takes options using the "functional options" idiom.
* Attachments are returned as `AttachmentReader`s, which are attachments parsed up to the data section, with a reader to read out the attachment data. This is a `LimitedReader`, which stops consumers from reading into the next record.

## TODO
- [ ] Docstrings
- [ ] Update example code, which is now out-of-date
- [ ] check benchmarks. Each message read out now needs two pointer-indirections rather than one, which isn't ideal.


go/cli/mcap/utils/utils.go
```diff
@@ -98,3 +98,15 @@ func WithReader(ctx context.Context, filename string, f func(remote bool, rs io.
 	defer rs.Close()
 	return f(remote, rs)
 }
+
```
james-rms, line 101: give it a docstring

wkalt:
limit reader can return io.EOF. We should make sure that doesn't cause us to interpret anything as an actual EOF.

james-rms:
Abandoning this approach also.

Issue #457, opened 2022-07-04
Python API: indexed reader


james-rms:
Right now we only have low-level MCAP read support in the Python API. The Go library, on the other hand, is able to use the Summary section to quickly find relevant channels and messages. Users of the Python API would appreciate this functionality as well.

PR #458, opened 2022-07-05
python: add indexed reader interface

james-rms:
**Public-Facing Changes**
* Type hints for StreamReader are now slightly more permissive.
* StreamReader now supports reading MCAP data from any record offset in the file, not just from the beginning.
* StreamReader can be configured to emit Chunks directly rather than unpacking them.
* Users can use the `McapReader` interface to easily iterate through messages by topic.

**Description**
Addresses #457. Add a higher-level MCAP reading API that allows users to filter messages by topic and time range, similar to the [Rosbag Python API](http://wiki.ros.org/rosbag/Code%20API#Python_API).

### DIFF
--- a/python/mcap/mcap0/data_stream.py
+++ b/python/mcap/mcap0/data_stream.py
@@ -1,12 +1,13 @@
 import struct
-from io import IOBase, BytesIO
+from io import BytesIO
+from typing import IO
 
 from .exceptions import EndOfFile
 from .opcode import Opcode
 
 
 class ReadDataStream:
-    def __init__(self, stream: IOBase):
+    def __init__(self, stream: IO[bytes]):
         self.__count = 0
         self.__stream = stream
 
--- a/python/mcap/mcap0/data_stream.py
+++ b/python/mcap/mcap0/data_stream.py
@@ -1,17 +1,14 @@
 import struct
-from io import BufferedIOBase, BytesIO
+from io import IOBase, BytesIO
 
 from .exceptions import EndOfFile
 from .opcode import Opcode
 
 
 class ReadDataStream:
-    def __init__(self, stream: BufferedIOBase):
+    def __init__(self, stream: IOBase):
         self.__count = 0
         self.__stream = stream
-
-    def __del__(self):
-        self.__stream.close()
 
     @property
     def count(self) -> int:
--- a/python/mcap/mcap0/reader.py
+++ b/python/mcap/mcap0/reader.py
@@ -18,7 +18,7 @@ from .records import (
     MetadataIndex,
     AttachmentIndex,
 )
-from .stream_reader import StreamReader, breakup_chunk
+from .stream_reader import StreamReader, breakup_chunk, MAGIC_SIZE
 from .data_stream import RecordBuilder
 from .summary import Summary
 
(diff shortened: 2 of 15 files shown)


python/mcap/mcap0/data_stream.py
```diff
@@ -1,18 +1,15 @@
...
 
 
 class ReadDataStream:
-    def __init__(self, stream: BufferedIOBase):
+    def __init__(self, stream: IOBase):
         self.__count = 0
         self.__stream = stream
 
-    def __del__(self):
-        self.__stream.close()
```
james-rms, line 14: note: `__del__` implementations rarely do what you want them to, since when an object gets deleted is entirely up to the garbage collector. Also, the `__stream` implementation will almost certainly close itself on deletion already.

python/mcap/mcap0/data_stream.py
```diff
@@ -1,18 +1,15 @@
...
-from io import BufferedIOBase, BytesIO
+from io import IOBase, BytesIO
 
 from .exceptions import EndOfFile
 from .opcode import Opcode
 
 
 class ReadDataStream:
-    def __init__(self, stream: BufferedIOBase):
+    def __init__(self, stream: IOBase):
```
james-rms, line 9: as far as I could tell nothing in the ReadDataStream implementation depends on buffering.

python/mcap/mcap0/reader.py
```diff
@@ -0,0 +1,238 @@
...
+    def iter_metadata(self) -> Iterator[Metadata]:
+        raise NotImplementedError()
+
+
+class SeekingReader(MCAPReader):
+    """an MCAPReader for reading out of non-seekable data sources."""
+
+    def __init__(self, stream: io.IOBase):
+        self._stream = stream
+        self._schemas = {}
```
james-rms, line 102: remove these

foxymiles (commented):
This looks good. Have you verified that the https://github.com/foxglove/python-mcap-ros1-support and https://github.com/foxglove/python-mcap-protobuf-support continue to work with these changes?

python/mcap/mcap0/reader.py
```diff
@@ -0,0 +1,291 @@
...
+            # no index available, use a non-seeking reader to read linearly through the stream.
+            return NonSeekingReader(self._stream).iter_messages(
+                topics, start_time, end_time
+            )
+        for chunk_index in _chunks_matching_topics(
+            summary, topics, start_time, end_time
+        ):
+            self._stream.seek(chunk_index.chunk_start_offset + 1 + 8, io.SEEK_SET)
+            chunk = Chunk.read(ReadDataStream(self._stream))
+            for record in breakup_chunk(chunk):
```
jtbandes, line 145: So it looks like we don't support out of order messages or merging across chunks, is that right? I guess the conformance tests don't currently test that but it would be nice to eventually provide support for this across all languages.

jtbandes (approved):

python/mcap/mcap0/reader.py
jtbandes: explained offline – TS implementation for reference https://github.com/foxglove/mcap/blob/c766292714102a58073d8420cf57023c11cb4481/typescript/core/src/v0/Mcap0IndexedReader.test.ts#L440

james-rms:
@foxymiles This PR doesn't modify any of the code that those support packages use, so should be good to go. I've made a ticket to update them to use the new reader implementation, so that people using those can do indexed reads as well.

PR #466, opened 2022-07-12
cpp: add protobuf writer example

james-rms:
**Public-Facing Changes**

**Description**
Add some example code for writing protobuf messages to an MCAP.


cpp/examples/protobuf_pointcloud_writer/protobuf_pointcloud_writer.cpp
```diff
@@ -0,0 +1,153 @@
...
+  if (argc != 2) {
+    std::cerr << "Usage: " << argv[0] << " <output.mcap>\n";
+    return 1;
+  }
+  const char* outputFilename = argv[1];
+
+  mcap::McapWriter writer;
+  auto options = mcap::McapWriterOptions("x-protobuf");
+  std::ofstream out(outputFilename, std::ios::binary);
+  writer.open(out, options);
```
jtbandes, line 75: There is a version of open() that takes in a filename

cpp/examples/protobuf_pointcloud_writer/protobuf_pointcloud_writer.cpp
```diff
@@ -0,0 +1,153 @@
...
+  // set up the schema and channel.
+  mcap::Schema schema("foxglove.PointCloud", "protobuf",
+                      std::string_view((char*)(descriptor_pb_bin), descriptor_pb_bin_len));
```
jtbandes, line 78: Not sure where these `descriptor_pb_bin` are supposed to come from right now, but I think we should use @danchia's method to build the descriptor set at runtime like this instead: https://foxglove.dev/blog/recording-robocar-data-with-mcap

It seems like less of a hassle than having to run with `--descriptor_set_out` and ship that data along with the binary.

cpp/examples/protobuf_pointcloud_writer/protobuf_pointcloud_writer.cpp
```diff
@@ -0,0 +1,153 @@
...
+    point.z = float(cos(phi) * scale);
+    return point;
+  }
+};
+
+// WriteLittleEndianFloat writes a little-endian float into the `output` string of bytes at
+// `offset`.
+void WriteLittleEndianFloat(float input, std::string* output, size_t offset) {
+  static_assert(sizeof(uint32_t) == sizeof(float));
+  uint32_t asInt = *reinterpret_cast<uint32_t*>(&input);
```
jtbandes, line 51: Technically I think this is undefined behavior. You are only allowed to reinterpret_cast arbitrary things to char/uint8: https://stackoverflow.com/a/38744467/23649

Our other mcap code only works on little endian systems anyway, so I think you are safe to continue that trend here:

https://github.com/foxglove/mcap/blob/f341c7f770ee0527a857c5fb780c316ec0d8d521/cpp/mcap/include/mcap/writer.inl#L1057-L1075

I thought we had a static_assert about this somewhere, but maybe we don't because c++17 doesn't provide std::endian 🤦

james-rms: This seems bad - if we want to run our MCAP writing code on ARM-based robots, we surely need to handle this properly.

jhurliman: This library works fine on arm32/arm64 platforms. If you can find a real big-endian system that someone might theoretically purchase and use then we can start caring about this.

cpp/dev.Dockerfile
```diff
@@ -12,7 +12,8 @@ RUN apt-get update && \
   make \
   perl \
   python3 \
-  python3-pip 
+  python3-pip \
```
james-rms, line 15: no longer needed

cpp/examples/protobuf_pointcloud_writer/protobuf_pointcloud_writer.cpp
```diff
@@ -0,0 +1,171 @@
...
+// Writes the FileDescriptor of this descriptor and all transitive dependencies
+// to a string, for use as a channel schema.
+std::string SerializeFdSet(const google::protobuf::Descriptor* toplevelDescriptor) {
+  google::protobuf::FileDescriptorSet fdSet;
+  std::queue<const google::protobuf::FileDescriptor*> toAdd;
+  toAdd.push(toplevelDescriptor->file());
+  std::unordered_set<std::string> added;
+  while (!toAdd.empty()) {
+    const auto& next = toAdd.front();
+    toAdd.pop();
```
jtbandes, line 55: I think a reference to the front() is invalidated once you call pop()

PR #470, opened 2022-07-14
C++: add protobuf reader example

james-rms:
**Public-Facing Changes**

Completes the loop and adds a C++ protobuf reader example.

**Description**

### DIFF
--- a/cpp/examples/protobuf/CMakeLists.txt
+++ b/cpp/examples/protobuf/CMakeLists.txt
@@ -33,8 +33,10 @@ target_compile_options(pointcloud_proto_lib PUBLIC -Wno-error=unused-parameter)
 add_executable(example_protobuf_writer writer.cpp)
 target_link_libraries(example_protobuf_writer ${CONAN_LIBS} pointcloud_proto_lib)
 
-# NOTE: reader does not depend on pointcloud_proto_lib because it reads definitions
-# dynamically out of the MCAP file.
-add_executable(example_protobuf_reader reader.cpp)
-target_link_libraries(example_protobuf_reader ${CONAN_LIBS} ${Protobuf_LIBRARIES})
-target_compile_options(example_protobuf_reader PUBLIC -Wno-error=unused-parameter)
+add_executable(example_protobuf_dynamic_reader dynamic_reader.cpp)
+target_link_libraries(example_protobuf_dynamic_reader ${CONAN_LIBS} ${Protobuf_LIBRARIES})
+target_compile_options(example_protobuf_dynamic_reader PUBLIC -Wno-error=unused-parameter)
+
+add_executable(example_protobuf_static_reader static_reader.cpp)
+target_link_libraries(example_protobuf_static_reader ${CONAN_LIBS} pointcloud_proto_lib)
+target_compile_options(example_protobuf_static_reader PUBLIC -Wno-error=unused-parameter)
--- a/cpp/examples/protobuf/CMakeLists.txt
+++ b/cpp/examples/protobuf/CMakeLists.txt
@@ -32,3 +32,9 @@ target_compile_options(pointcloud_proto_lib PUBLIC -Wno-error=unused-parameter)
 
 add_executable(example_protobuf_writer writer.cpp)
 target_link_libraries(example_protobuf_writer ${CONAN_LIBS} pointcloud_proto_lib)
+
+# NOTE: reader does not depend on pointcloud_proto_lib because it reads definitions
+# dynamically out of the MCAP file.
+add_executable(example_protobuf_reader reader.cpp)
+target_link_libraries(example_protobuf_reader ${CONAN_LIBS} ${Protobuf_LIBRARIES})
+target_compile_options(example_protobuf_reader PUBLIC -Wno-error=unused-parameter)
--- a/cpp/examples/protobuf/dynamic_reader.cpp
+++ b/cpp/examples/protobuf/dynamic_reader.cpp
@@ -12,6 +12,27 @@
 
 namespace gp = google::protobuf;
 
+// Loads the FileDescriptorSet from a protobuf schema definition into a SimpleDescriptorDatabase.
+bool LoadSchema(const mcap::SchemaPtr schema, gp::SimpleDescriptorDatabase* protoDb) {
+  gp::FileDescriptorSet fdSet;
+  size_t size = schema->data.size();
+  if (!fdSet.ParseFromArray(static_cast<const void*>(&(schema->data[0])), size)) {
+    std::cerr << "failed to parse schema data" << std::endl;
+    return false;
+  }
+  gp::FileDescriptorProto unused;
+  for (int i = 0; i < fdSet.file_size(); ++i) {
+    const auto& file = fdSet.file(i);
+    if (!protoDb->FindFileByName(file.name(), &unused)) {
+      if (!protoDb->Add(file)) {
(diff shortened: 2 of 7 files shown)


cpp/examples/protobuf/CMakeLists.txt
```diff
@@ -32,3 +32,9 @@ target_compile_options(pointcloud_proto_lib PUBLIC -Wno-error=unused-parameter)
 
 add_executable(example_protobuf_writer writer.cpp)
 target_link_libraries(example_protobuf_writer ${CONAN_LIBS} pointcloud_proto_lib)
+
+# NOTE: reader does not depend on pointcloud_proto_lib because it reads definitions
+# dynamically out of the MCAP file.
```
jtbandes, line 37: This generic dynamic reading is a good example to have. However, I think many users will want to use statically generated cpp proto classes that they are already using in their other code. Can we extend this example or make another example that uses a generated cpp class to deserialize (matching on schema name) instead of the generic gp::Message?

cpp/examples/protobuf/reader.cpp
```diff
@@ -0,0 +1,152 @@
...
+namespace gp = google::protobuf;
+
+bool WriteFileDescriptorsToPool(const gp::FileDescriptorSet* fdSet, gp::DescriptorPool* pool) {
+  // The FileDescriptorSet structure contains a repeated field of FileDescriptorProto instances.
+  // Adding these to the DescriptorPool should be as simple as calling pool->BuildFile,
+  // however BuildFile fails if any of a FileDescriptorProto's dependencies have not already been
+  // added.
+  // Therefore, do a topological sort first to put the FileDescriptorProtos in dependency order.
+  // we use the technique described here:
+  // https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search
```
jtbandes, line 24: I wonder if it would work to Add() each FileDescriptorProto into a SimpleDescriptorDatabase? https://github.com/protocolbuffers/protobuf/blob/520c601c99012101c816b6ccc89e8d6fc28fdbb8/src/google/protobuf/descriptor_database.h#L144-L165

cpp/examples/protobuf/CMakeLists.txt
```diff
@@ -32,3 +32,9 @@ target_compile_options(pointcloud_proto_lib PUBLIC -Wno-error=unused-parameter)
...
+add_executable(example_protobuf_reader reader.cpp)
```
jtbandes, line 38: Is this run in CI? would be nice to make sure it continues to work

cpp/examples/protobuf/dynamic_reader.cpp
```diff
@@ -0,0 +1,90 @@
...
+        if (!protoDb.FindFileByName(file.name(), &unused)) {
+          if (!protoDb.Add(file)) {
+            std::cerr << "failed to add definition " << file.name() << "to protoDB" << std::endl;
+            reader.close();
+            return 1;
+          }
+        }
+      }
+    }
+    auto descriptor = protoPool.FindMessageTypeByName(it->schema->name);
```
jtbandes, line 65: nit: if FindMessageTypeByName returns non-null the first time, we end up discarding that and immediately calling it again

cpp/examples/protobuf/dynamic_reader.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  std::cout << "topic\t\ttype\t\t\ttimestamp\t\tfields" << std::endl;
+
+  for (auto it = messageView.begin(); it != messageView.end(); it++) {
+    // skip any non-protobuf-encoded messages.
+    if (it->schema->encoding != "protobuf") {
+      continue;
+    }
+    // If the proto descriptor is not yet loaded, load it.
+    if (protoPool.FindMessageTypeByName(it->schema->name) == nullptr) {
```
jtbandes, line 45: I think we might want to recommend caching the descriptors by schema id (or channel id?) rather than having to do a find by name for every single message in the file (which I am guessing would be somewhat slower). It would also help readability to break out this FileDescriptorSet parsing into a separate function.

cpp/examples/protobuf/dynamic_reader.cpp
```diff
@@ -0,0 +1,90 @@
...
+      reader.close();
+      return 1;
+    }
+
+    std::vector<const gp::FieldDescriptor*> fields;
+    message->GetReflection()->ListFields(*message, &fields);
+    std::cout << it->channel->topic << "\t(" << it->schema->name << ")\t[" << it->message.logTime
+              << "]:\t{ ";
+    for (const auto field : fields) {
+      std::cout << field->name() << " ";
```
jtbandes, line 84: assuming outputting the field value is a TODO? :)

cpp/examples/protobuf/static_reader.cpp
```diff
@@ -0,0 +1,47 @@
...
+                << std::endl;
+      return 1;
+    }
+  }
+
+  auto messageView = reader.readMessages();
+  std::cout << "topic\t\ttype\t\t\ttimestamp\t\tframe_id" << std::endl;
+  for (auto it = messageView.begin(); it != messageView.end(); it++) {
+    // skip any non-protobuf, non PointCloud messages.
+    if ((it->schema->encoding != "protobuf") || it->schema->name != "foxglove.PointCloud") {
```
jtbandes, line 32: should we be checking `messageEncoding` here as well (and in the other example)?

cpp/examples/protobuf/dynamic_reader.cpp
james-rms: It's not a TODO, but it is a mess - for the example MCAP we generate in the writer (a pointcloud) the pointcloud data value is a massive binary blob, which I didn't like when i printed it.

cpp/examples/protobuf/dynamic_reader.cpp
james-rms: I'm not exactly sure how safe that is - the [DescriptorPool docs](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.descriptor#DescriptorPool) describe each descriptor as a "whole lot of cross-linked bits of data", which makes me unsure if the pointer returned from `FindMessageTypeByName` is safe to retain between `BuildFile` calls.

cpp/examples/protobuf/dynamic_reader.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "mcap/reader.hpp"
+#include <vector>
+
+namespace gp = google::protobuf;
+
+// Loads the FileDescriptorSet from a protobuf schema definition into a SimpleDescriptorDatabase.
+bool LoadSchema(const mcap::SchemaPtr schema, gp::SimpleDescriptorDatabase* protoDb) {
+  gp::FileDescriptorSet fdSet;
+  size_t size = schema->data.size();
+  if (!fdSet.ParseFromArray(static_cast<const void*>(&(schema->data[0])), size)) {
```
jtbandes, line 19:
```suggestion
  if (!fdSet.ParseFromArray(static_cast<const void*>(schema->data), size)) {
```

jtbandes (approved):

cpp/examples/protobuf/dynamic_reader.cpp
jtbandes: I would hope the docs would explain the usage rules if it is not safe to do... but I also don't want to read all the code to figure it out.

PR #472, opened 2022-07-14
[typescript] Fix indexed message reading

defunctzombie:
Under certain situations the Mcap0IndexedReader would produce out of log time order messages from `readMessages`. This change fixes this behavior to produce in order messages.

**Problem Description**

To read data in-order the Mcap0IndexedReader uses a heap of ChunkCursors and reads data from the first cursor in the heap (the "min"). The heap is setup to use the `compare` method exposed on each ChunkCursor instance to compare it to other instances to maintain the properties of a heap and bring the "min" cursor to the "front". The logic within the compare function would try to sort ChunkCursors by which one needed reading next but also which one needed loading next (since the message indexes are read lazily for the cursors). This logic produced an anachronism within the heap which resulted in the front heap element being incorrect and ultimately resulting in producing out of order messages.

Here is how the compare logic failed.

Our file contains 3 chunks, each chunk contains messages denoted here with a number indicating their log time and drawn to show the relative time of each message in the overall time:

```
A: [1       4]
B:   [2]
C:      [3]
```

When the ChunkCursors are inserted into the heap their message indexes are un-initialized and they create an ordering of: [A, B, C] since the start time of A < B < C.

The read loop begins. It looks at the front of the heap and sees chunk cursor A. It sees that this chunk cursor needs initializing, it runs logic to initialize it and re-inserts it into the front of the heap. This causes the heap to re-sort from the front. When comparing to its child (B), B now evaluates before A because B has not been initialized (the compare function considers un-initialized chunk cursors as sorting before initialized ones (In certain situations - this is ultimately the bug).

B is at the front of the heap now. The read loop grabs B, sees that it needs initializing and runs logic to initialize it and re-insert it into the front of the heap. This causes the heap to re-sort from the front. The heap wants to properly sort B down. To do this it finds the "best" child of B by comparing the children of B (A, C) against each other and getting the smallest. When comparing A to C, the cursor compare function returns C as smaller than A because C has not been initialized. It returns C as the "best" child and compares B to C. Unfortunately, the compare function sees that B is completely before C and sorts B before C. Our heap is now: [B, A, C].

The read loop continues, grabs the first element (B), sees it is initialized and reads its first message (2). We've now read message (2) out of order because we never read message (1).

**fix**

The fix is to simplify the cursor compare function so that it no longer produces an anachronism but rather a stable sort. Rather than trying to surface un-loaded chunks before loaded ones in certain situations, the compare function evaluates the "sort time" of the two chunk cursors. For un-initialized chunks the sort time is simply the start time of the chunk and for initialized chunks the sort time is the time of the "read head" within the chunk.

Using the above scenario we now produce a stable sort. As the read loop evaluates the initial front of the heap it finds chunk cursor A. It initializes it and re-inserts it into the heap causing a re-sort. Since cursor A is still before cursor B, it remains at the front. The read loop reads message (1) from it and re-sorts the heap. Cursor A now has a time of (4) and so it sorts after B and C. The read loop sees (B) is un-initialized, initializes it and re-sorts the heap. The heap sort remains stable because A is still at (4), B is at (2), and C at (3): B, C, A. The read loop reads message (2) from B, and moves on to C. The process repeats until C is read and then finally A and we are done.

### DIFF
--- a/typescript/core/src/v0/ChunkCursor.ts
+++ b/typescript/core/src/v0/ChunkCursor.ts
@@ -59,20 +59,15 @@ export class ChunkCursor {
    * and re-sort the cursors.
    */
   compare(other: ChunkCursor): number {
-    const reverse = this.reverse;
     if (this.reverse !== other.reverse) {
       throw new Error("Cannot compare a reversed ChunkCursor to a non-reversed ChunkCursor");
     }
 
-    const diff = Number(this.getSortTime() - other.getSortTime());
+    let diff = Number(this.getSortTime() - other.getSortTime());
+
+    // Break ties by chunk offset in the file
     if (diff === 0) {
-      if (reverse) {
-        // Break ties by chunk offset in the file
-        return Number(other.chunkIndex.chunkStartOffset - this.chunkIndex.chunkStartOffset);
-      } else {
-        // Break ties by chunk offset in the file
-        return Number(this.chunkIndex.chunkStartOffset - other.chunkIndex.chunkStartOffset);
-      }
+      diff = Number(this.chunkIndex.chunkStartOffset - other.chunkIndex.chunkStartOffset);
     }
 
     return this.reverse ? -diff : diff;
--- a/typescript/core/src/v0/ChunkCursor.ts
+++ b/typescript/core/src/v0/ChunkCursor.ts
@@ -64,65 +64,18 @@ export class ChunkCursor {
       throw new Error("Cannot compare a reversed ChunkCursor to a non-reversed ChunkCursor");
     }
 
-    if (reverse) {
-      // If chunks don't overlap at all, sort later chunk first
-      if (this.chunkIndex.messageEndTime < other.chunkIndex.messageStartTime) {
-        return 1;
-      }
-      if (this.chunkIndex.messageStartTime > other.chunkIndex.messageEndTime) {
-        return -1;
-      }
-    } else {
-      // If chunks don't overlap at all, sort earlier chunk first
-      if (this.chunkIndex.messageEndTime < other.chunkIndex.messageStartTime) {
-        return -1;
-      }
-      if (this.chunkIndex.messageStartTime > other.chunkIndex.messageEndTime) {
-        return 1;
+    const diff = Number(this.getSortTime() - other.getSortTime());
+    if (diff === 0) {
+      if (reverse) {
+        // Break ties by chunk offset in the file
+        return Number(other.chunkIndex.chunkStartOffset - this.chunkIndex.chunkStartOffset);
+      } else {
+        // Break ties by chunk offset in the file
+        return Number(this.chunkIndex.chunkStartOffset - other.chunkIndex.chunkStartOffset);
       }
     }
 
-    // If a cursor has not loaded message indexes, sort it first so it can get loaded and re-sorted.
-    // If it overlaps with other cursors, we won't know its order relative to other cursors until we
-    // have the message indexes.
-    if (!this.messageIndexCursors) {
-      return -1;
-    }
-    if (!other.messageIndexCursors) {
-      return 1;
-    }
-
-    // Earlier messages come first
-    const cursorA = this.messageIndexCursors.peek();
-    if (!cursorA) {
-      throw new Error(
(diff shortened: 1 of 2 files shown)


typescript/core/src/v0/Mcap0IndexedReader.test.ts
```diff
@@ -548,4 +548,109 @@ describe("Mcap0IndexedReader", () => {
       message6,
     ]);
   });
+
+  // fixme
+  it.only("reads something", async () => {
```
defunctzombie, line 553: remove only

typescript/core/src/v0/ChunkCursor.ts
```diff
@@ -64,65 +64,18 @@ export class ChunkCursor {
...
-      if (this.chunkIndex.messageEndTime < other.chunkIndex.messageStartTime) {
-        return -1;
-      }
-      if (this.chunkIndex.messageStartTime > other.chunkIndex.messageEndTime) {
-        return 1;
+    const diff = Number(this.getSortTime() - other.getSortTime());
+    if (diff === 0) {
+      if (reverse) {
+        // Break ties by chunk offset in the file
+        return Number(other.chunkIndex.chunkStartOffset - this.chunkIndex.chunkStartOffset);
```
james-rms, line 71: I'm not sure why this isn't simpler:
```javascript
if (diff === 0) {
    diff = Number(this.chunkIndex.chunkStartOffset - other.chunkIndex.chunkStartOffset);
}
```

james-rms: I guess you'd have to make diff not-const.

james-rms:
I think including the diagram into the unit test comments would help debug it in the future:

```
A: [1       4]
B:   [2]
C:      [3]
```

or something similar.

typescript/core/src/v0/Mcap0IndexedReader.test.ts
```diff
@@ -548,4 +548,280 @@ describe("Mcap0IndexedReader", () => {
...
+      expect(readable.readCalls).toEqual(8);
+      collected.push((await messageIter.next()).value);
+      expect(readable.readCalls).toEqual(10);
+      collected.push((await messageIter.next()).value);
+      expect(readable.readCalls).toEqual(10);
+
+      expect(collected).toEqual([message3, message4, message5, message6]);
+    }
+
+    {
```
jtbandes, line 668: Could we dedupe these tests using `it.each([true, false])("", (reverse) => ...` or `for (const reverse of [true, false])` inside the test?

typescript/core/src/v0/Mcap0IndexedReader.test.ts
```diff
@@ -548,4 +548,280 @@ describe("Mcap0IndexedReader", () => {
       message6,
     ]);
   });
+
+  it("correctly reads overlapping chunks", async () => {
```
jtbandes, line 552: making the description a bit more specific. as is, it's basically the same as the previous test description
```suggestion
  it("uses stable sort when loading overlapping chunks", async () => {
```

jtbandes (approved):
👏 what a wild ride

typescript/core/src/v0/ChunkCursor.ts
```diff
@@ -64,65 +64,18 @@ export class ChunkCursor {
...
+      } else {
+        // Break ties by chunk offset in the file
+        return Number(this.chunkIndex.chunkStartOffset - other.chunkIndex.chunkStartOffset);
```
wkalt, line 74: I might be jumping in with too little context here, but I think sort stability is going to depend on fixing,
* message log time
* offset of message within chunk
* offset of chunk within file

AFAICT the tie-breaking is only considering the first and third of those, but not the second. So if you had two messages with the same timestamp in the same chunk, what guarantees stability?

defunctzombie: @wkalt This is comparing two different chunks - if you have two messages with the same timestamp in the same chunk those are in the same chunk and don't impact sorting one chunk against another.

typescript/core/src/v0/ChunkCursor.ts
defunctzombie: simplified per recommendation

PR #473, opened 2022-07-14
python: move support packages into MCAP repo

james-rms:
**Public-Facing Changes**

**Description**
This PR moves our python support packages into the one repo. The reasons for doing this are:
1. All of our python MCAP documentation is discoverable from the same set of generated docs at https://mcap.dev/docs/python.
2. One CI / test / lint configuration for all of our python code will make our code more consistent.
3. Less opportunity for surprising compatibility issues between packages.

## TODO
- [x] Decide on a mechanism for releasing the support packages. I'd be fine extending our existing multi-release mechanism, so that a tag `releases/python-mcap-ros1-support/v1.2.3` would release the ROS1 support package.
- [x] Give our PyPI and TestPyPI user accounts permission to push the support packages from this repo.
- [ ] Once merged, update the README in the https://github.com/foxglove/python-mcap-protobuf-support and https://github.com/foxglove/python-mcap-ros1-support repos to point to these folders.


python/Makefile
```diff
@@ -0,0 +1,36 @@
...
+	find . -name "*.egg-info" -type d -exec rm -rf {} +
+	find docs -name "*-apidoc" -type d -exec rm -rf {} +
+
+docs: pipenv
+	pipenv run sphinx-apidoc --separate --tocfile index --output-dir docs/mcap-apidoc mcap/mcap
+	pipenv run sphinx-apidoc --separate --tocfile index --output-dir docs/mcap-protobuf-apidoc mcap-protobuf-support/mcap_protobuf
+	pipenv run sphinx-apidoc --separate --tocfile index --output-dir docs/mcap-ros1-apidoc mcap-ros1-support/mcap_ros1
+	pipenv run sphinx-build -W --keep-going docs ../__docs__/python
+
+.PHONY: test lint build pipenv clean docs
```
jtbandes: I read about an organizational trick once which I like to use for makefiles: always keep `.PHONY: foo` right above `foo: ...` to avoid forgetting to add/update the PHONY list:

```makefile
.PHONY: test
test: ...

.PHONY: build
build: ...
```

.github/workflows/ci.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: 3.7
-      - run: cd python && pip install .
+      - run: cd python && pip install -e mcap
```
jtbandes, line 92: Maybe this pip install should be replaced by `pipenv install` like we have in the other python workflow?

python/Pipfile
```diff
@@ -3,10 +3,20 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 name = "pypi"
 
+[[source]]
+url = "https://rospypi.github.io/simple"
```
jtbandes, line 7: There's a big warning on this repo that says "THIS IS NOT THE OFFICIAL ROSPY INDEX SERVER", what's the deal with that?

.github/workflows/ci.yml
james-rms: I actually prefer this way because it tests that our `mcap` library properly specifies its dependencies in `setup.cfg`.

python/Pipfile
james-rms: Yes - ROS does not officially support any pip package, their installation story for python code is either through `.deb` packages or source installs via catkin or ament.

defunctzombie (approved):
All hail the monorepo!

PR #475, opened 2022-07-19
Python: move code to `mcap` module

james-rms:
**Public-Facing Changes**
* all child modules in `mcap.mcap0` have been moved to the root `mcap` module.
* Their content is still importable from `mcap0`, but doing so triggers a `DeprecationWarning`. The idea is to ease migration by not breaking scripts.
* the `profile` and `library` parameters to `writer.start()` are no longer mandatory, and we now have a sane default library string.
 * The pointcloud writer tutorial has been updated to match.
* the library version can now be imported from `mcap.__version__` or `mcap.__version__.__version__`.

**Description**


amacneil:
stale pr?

PR #479, opened 2022-07-21
CLI: ensure filter output is chunked

wkalt:
Prior to this commit, the filter subcommand accepted an
output-compression flag but emitted an unchunked mcap file by default.

This change ensures the output file is always chunked, and also adds an
additional command line option for specifying a chunk size, which is
defaulted to 4MB.

### DIFF
--- a/go/cli/mcap/cmd/filter.go
+++ b/go/cli/mcap/cmd/filter.go
@@ -23,6 +23,7 @@ var (
 	filterIncludeMetadata    bool
 	filterIncludeAttachments bool
 	filterOutputCompression  string
+	filterChunkSize          int64
 )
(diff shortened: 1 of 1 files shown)


wkalt:
I think we should maybe follow this up with a change to either/both,
* make it an error to supply a compression without chunked == true, in writer options
* switch the "Chunked" writer option to "Unchunked", so that it defaults to producing a chunked rather than an unchunked file.

I have not made these changes in this patch though; will file a separate ticket.

wkalt:
followup ticket: https://github.com/foxglove/mcap/issues/480

james-rms (approved):

PR #485, opened 2022-07-22
Add Homebrew to CLI installation instructions

jtbandes:


defunctzombie (approved):

go/cli/mcap/README.md
```diff
@@ -7,9 +7,15 @@ for details.
 
 ### Installing:
 
-Either install from [releases
+Either install from [Homebrew](https://formulae.brew.sh/formula/mcap#default), [releases
 binaries](https://github.com/foxglove/mcap/releases) or by using go.
 
+#### From Homebrew
+
+To install using [Homebrew](https://brew.sh) on macOS or Linux, run:
```
wkalt, line 15: I find it a little weird to list this above binaries (and even using go), since in practice it's only applicable to mac users while the other two are general methods.

I know linux users can install homebrew, but I think it's extremely rare for them to do so.

wkalt (approved):
wording seems fine; one comment on the ordering

PR #497, opened 2022-07-28
Jrms/docs site sphiinx

james-rms:
**Public-Facing Changes**

**Description**


docs/Guides/getting-startted/converting-to-mcap.md
```diff
@@ -0,0 +1,25 @@
+# Converting To MCAP
```
wkalt, line 1: filepath of this doc has "getting-startted" instead of getting-started

docs/Guides/getting-startted/converting-to-mcap.md
```diff
@@ -0,0 +1,25 @@
...
+### ROS 1
+
+```
+$ mcap convert ../../testdata/bags/demo.bag demo.mcap
+```
+
+### ROS 2
+
+```
+$ mcap convert multiple_files_1.db3 demo.mcap
```
wkalt, line 16: this filepath is from one of the rosbag2 test files (either directly or via some example in the CLI I guess). Probably doesn't make much sense out of context, so might be clearer to just say `mcap convert demo.db3 demo.mcap`. If there's another readme file with the same issue we should change it.

docs/Guides/getting-startted/converting-to-mcap.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+mcap will search the path stored in your $AMENT_PREFIX_PATH environment variable to locate the ROS message definitions on your hard drive.
```
wkalt, line 19: "The tool will search ..."?

wkalt: or This will search, maybe - mcap the file format isn't searching, so the sentence is a little unclea.r

docs/Guides/getting-startted/converting-to-mcap.md
```diff
@@ -0,0 +1,25 @@
...
+
+Alternatively, you can specify a colon-separated list of directories for the CLI tool to search using the ament-prefix-path flag:
```
wkalt, line 21: this says colon-separated but the example has semicolon-separated. I thought it was a comma-separated list of colon-delimited $PATH-style paths. We should check the code to confirm

docs/Guides/getting-startted/working-with-mcap.md
```diff
@@ -0,0 +1,37 @@
...
+To validate an MCAP file:
+
+```
+$ mcap doctor demo.mcap
+```
+
+To report summary statistics on an MCAP file to get information like number of messages, start and end time, channels, etc.:
+
+```
+$ mcap info demo.mcap
```
wkalt, line 18: We should have outputs after these commands.

This introduces a bit of a sync issue because it will be tough to keep these guides up to date with changes to the tooling. But I think even slightly outdated outputs would be preferable to no outputs for this kind of doc.

docs/Guides/reading-mcap-files/json-2.md
```diff
@@ -0,0 +1,184 @@
...
+channel_id = writer.register_channel(
+  topic="pointcloud",
+  message_encoding=MessageEncoding.JSON,
+  schema_id=schema_id,
+)
+```
+
+### Write messages
+
+Let's write a single `foxglove.PointCloud` message on the channel we just created:
```
wkalt, line 124: this file has a mix of characters for single quote/apostrophe. Here it's `Let's` while a bit higher up it's `We’ll`. Not sure what is causing the difference (different editors maybe?) but the first one (`Let's`) seems more standard to me.

docs/Guides/reading-mcap-files/json-2.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+That’s it! We now have a valid MCAP file with a single point cloud message.
+
+### Inspect your file
+
+To inspect your MCAP file, install the [MCAP CLI tool](https://github.com/foxglove/mcap/tree/main/go/cli/mcap):
+
+```bash
+$ brew install mcap
```
wkalt, line 153: assumes mac - maybe better to link to the install instructions, than duplicate?

docs/Guides/reading-mcap-files/json-2.md
```diff
@@ -0,0 +1,184 @@
...
+  library: my-excellent-library
+  profile: x-jsonschema
+  messages: 1
+  duration: 0s
+  start: 2011-11-04T01:36:05.987339008+11:00 (1320330965.987339008)
+  end: 2011-11-04T01:36:05.987339008+11:00 (1320330965.987339008)
+  compression:
+    zstd: [1/1 chunks] (48.09%)
+  channels:
+    (1) pointcloud  1 msgs (+Inf Hz)   : foxglove.PointCloud [jsonschema]
```
wkalt, line 169: +Inf Hz looks a little strange without explanation. The reason is because the duration of a one-message file is zero, so this is a divide by zero. We could get around this by recording two messages if it seems odd.

docs/Guides/visualizing-mcap-in-foxglove-studio/store-in-foxglove-data-platform.md
```diff
@@ -0,0 +1,9 @@
+# Store in Foxglove Data platform
+
+[Sign up for a Foxglove account ](https://console.foxglove.dev) to access the data management features of [Foxglove Data Platform](https://foxglove.dev/data-platform) .
```
wkalt, line 3: space before .

docs/Guides/visualizing-mcap-in-foxglove-studio/store-in-foxglove-data-platform.md
```diff
@@ -0,0 +1,9 @@
+# Store in Foxglove Data platform
```
wkalt, line 1: why does this go in the mcap repo?

docs/Guides/visualizing-mcap-in-foxglove-studio/store-in-foxglove-data-platform.md
```diff
@@ -0,0 +1,9 @@
...
+
+[Sign up for a Foxglove account ](https://console.foxglove.dev) to access the data management features of [Foxglove Data Platform](https://foxglove.dev/data-platform) .
+
+## Benefits
+
+Instead of having to pass around hard drives every time you want to share data, you can start importing your MCAP data into a central repository for all your teammates to access and collaborate on.
```
wkalt, line 7: I think we can make a stronger pitch if we feel this is the place for a DP pitch, but it doesn't make sense to me to pitch paid foxglove products in the mcap repo.

I think there could be a sense in which that works against us/mcap, by suggesting that mcap is only useful if you want to use foxglove - so I would probably remove this guide and maybe relocate it to the website/DP docs, but if we don't want that I can provide some expanded copy.

docs/Guides/visualizing-mcap-in-foxglove-studio/visualize-in-foxglove-studio.md
```diff
@@ -0,0 +1,23 @@
+# Visualize in Foxglove Studio
+
+## Using Foxglove schemas
```
wkalt, line 3: similar to the previous comment, I wonder if mcap repo is the place for this, or whether we could just give a link here to some docs on the studio side to describe this stuff. We will presumably need to duplicate this info in studio docs.

docs/Home/core-concepts.md
```diff
@@ -0,0 +1,17 @@
+# Messages, Channels & Schemas
```
wkalt, line 1: These terms should be defined in the specification. If they are, should these sections just link to the relevant definition in the spec?

docs/Home/history.md
```diff
@@ -0,0 +1,15 @@
+# History
```
wkalt, line 1: I think we should replace this History page with something more like "Motivation".

The narrative history isn't too interesting here, and may come off a little pretentious describing before/after as if they are industry shifts that have taken place already. What seems interesting is the motivation for creating a new format:
* must support arbitrary message serializations (like ros2, unlike ros1)
* must support efficient access over a slow network (like ros1, unlike ros2)
* must support heterogeneous types (like ros*, unlike avro, parquet [without abuse/giant union])
* must be self-contained

I think this section would be stronger if it were framed as "here's the requirements, here's the solutions that exist, here's mcap & how it solves", rather than "history of the format".

docs/Home/supported-data-serialization-formats.md
```diff
@@ -0,0 +1,13 @@
+# Data Serialization Formats
+
+MCAP files can store multiple streams of heterogeneous data, each serialized in a different data format.
```
wkalt, line 3: "Supported serialization formats for MCAP channels are listed below" - this avoids introducing a new term "streams". In mcap those are channels.

docs/index.rst
```diff
@@ -0,0 +1,76 @@
+****************
+MCAP File Format
+****************
+
+MCAP (pronounced "em-cap") is a modular container file format for heterogeneous timestamped data. It is ideal for robotics applications, as it can record multiple streams of structured and unstructured data (e.g. ROS, Protobuf, JSON Schema, etc.) in a single file.
```
wkalt, line 5: not clear to the reader how the second part of this sentence supports the first part ("it's ideal for robotics applications"). Maybe the context is that robotics applications involve integrating proprietary stuff that outputs things in formats you don't control? Unclear

wkalt:
I took a pass through it - left a few comments.

A few high-level questions are,
* Should foxglove paid products like the data platform be mentioned in the mcap repo docs at all?
* How aggressive should we be in eliminating duplication between these docs and the README files? Can we replace some of the content in this patch with links to the relevant README section, and if not how do we keep things in sync?
* What kind of docs writing style are we going for? These docs use a handful of stylistic keywords - let's, you've, we'll, history, before/after - they are written in a narrative, hand-holding kind of way. Other docs we have in this repo are shooting for dry & descriptive. What style are we targeting, and does this vary between the mcap, studio, website, repositories? I think probably yes, but we are missing a consistent policy.

james-rms:
Closing in favour of #518.

PR #507, opened 2022-08-02
doctor: check statistics start/end time

### LINKED ISSUE #501: `mcap info` shows incorrect start/end timestamps for my file
**Description**

When running `mcap info` on [whattimeisit.mcap.zip](https://github.com/foxglove/mcap/files/9238763/whattimeisit.mcap.zip), the following incorrect start/end/duration info is shown:

```
$ mcap info whattimeisit.mcap
library: python-mcap-ros1-support
profile: ros1
messages: 100
duration: -460947h7m35.322474798s
start: 2022-08-02T11:07:45.222474798+08:00 (1659409665.222474798)
end: 1970-01-01T08:00:09.9+08:00 (9.900000000)
compression:
	zstd: [1/1 chunks] (80.31%)
channels:
  	(1) /tf  100 msgs (0.00 Hz)   : tf2_msgs/TFMessage [ros1msg]
attachments: 0
```

Studio shows correct start/end timestamps for this file.

- Version: v0.0.13
- Platform: MacOS

jtbandes:
**Public-Facing Changes**
Updated `mcap doctor` to check the correctness of Statistics message_start_time/message_end_time.

**Description**
Fixes #501.

Reports new errors on the file from #501:
```
Examining /Users/jacob/Downloads/whattimeisit.mcap
Statistics has message start time 1659409665222474798, but the minimum chunk start time is 0
Statistics has message end time 9900000000, but the maximum chunk end time is 1659409665222474798
```

### DIFF
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -230,7 +230,7 @@ func (doctor *mcapDoctor) Examine() {
 			}
 
 			if header.Library == "" {
-				doctor.warn("Header.library field should be non-empty. Its good to include a library field for reference.")
+				doctor.warn("Header.library field should be non-empty. The library field should be set to a value that identifies the software which produced the file.")
 			}
 
 			var customProfile = strings.HasPrefix(header.Profile, "x-")
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -332,6 +332,9 @@ func (doctor *mcapDoctor) Examine() {
 			if err != nil {
 				doctor.error("Failed to parse statistics:", err)
 			}
+			if doctor.statistics != nil {
+				doctor.error("File contains multiple Statistics records")
+			}
 			doctor.statistics = statistics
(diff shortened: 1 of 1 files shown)


go/cli/mcap/cmd/doctor.go
```diff
@@ -369,6 +381,15 @@ func (doctor *mcapDoctor) Examine() {
 			doctor.error("Chunk at offset %d has uncompressed size %d, but its chunk index has uncompressed size %d", chunkOffset, chunk.UncompressedSize, chunkIndex.UncompressedSize)
 		}
 	}
+
+	if doctor.statistics != nil {
```
wkalt, line 385: is it possible we will hit this on an unchunked file, without erroring in some other way?

wkalt: cause if the file were unchunked but had statistics, I think the logic would be wrong

jtbandes: what would be the right thing to do in that case? Would it work if I switch this to calculate minStartTime/maxEndTime from looking at all the message records rather than the chunk indexes?

PR #508, opened 2022-08-02
python: add version identifier to mcap_ros1

james-rms:
**Public-Facing Changes**

**Description**


python/mcap-ros1-support/mcap_ros1/writer.py
```diff
@@ -1,15 +1,23 @@
...
 from mcap.mcap0.writer import Writer as McapWriter
+import mcap
 import time
 
+from . import __version__
+
+
+def _library_identifier():
+    mcap_version = getattr(mcap, "__version__", "<=0.0.10")
+    return f"mcap-ros1-support {__version__}; mcap {mcap_version}"
```
jhurliman, line 12: Was dropping the `python-` prefix intentional?

PR #515, opened 2022-08-04
Package named incorrectly in python example

emersonknapp:
Just a fix so this builds against latest verson of mcap plugin


jtbandes (approved):

emersonknapp:
@jtbandes some permissions thing going wrong with the CI/cpp check, do you know what's going on there?

PR #518, opened 2022-08-05
docs: generate with mkdocs

james-rms:
**Public-Facing Changes**
Switches `mcap.dev` to being generated with [mkdocs](https://mkdocs.org). Check the preview out in the CI "Pages Preview" details link below.

**Description**

Evaluated [Docusaurus v2](https://docusaurus.io), [Sphinx](https://sphinx-docs.org), **mkdocs** and [readme.com](https://readme.com) for generating the MCAP Github Pages site at `mcap.dev`.

A quick summary of the evaluation results:
### Pros
- simple build and deployment
- sophisticated text markup capabilities:
  - `literalinclude` is unique: you can include chunks of a script as a code block in a tutorial file.
- best-of-class integration with generated API docs for Python
### Cons
- defaults aren't great for presenting a top-level nav header easily
- Setting up the `toctree` is fairly involved. It's not too hard once you get the hang of it, but there's some ramp-up period required.
- No support for tabbed code blocks
- Markdown is supported, but most functionality only works from restructuredText, meaning that users have to ramp up in `.rst` to contribute properly.

### Pros
- Slick interface
- Markdown native
- good support for advanced markup like admonitions, tabbed code blocks
### Cons
- Docusaurus v2 is a SPA rather than a pure static site generator, which makes the build and preview process more complex than other options
  - It's impossible to preview properly from a GCS bucket without a CNAME redirect, or from `file://` url.
  - It's possible to preview with `Vercel` preview, since it's a node app, but this stops us from previewing API docs that are not generated by node apps (which is all of our generated docs except Typescript).

### Pros
- Slick interface
- Markdown native
- Slick in-browser editor and publishing flow
### Cons
- Not a static site generator - have to go through the `readme.com` interface and API to deploy a site.

### Pros
- Slick interface
- markdown native
- good support for advanced markup like admonitions, tabbed code blocks
### Cons
- No built-in support for including code blocks directly from example code. There's good support for building this with an [extension](https://www.mkdocs.org/dev-guide/plugins/) though.

## Conclusion

Mkdocs seems like a good fit.

Fixes: https://github.com/foxglove/private-issues/issues/58


amacneil:
Can you elaborate more on mkdocs vs docusaurus? Are they both SPAs?

Imo docusaurus default theme looked a bit nicer, but we can probably improve mkdocs.

james-rms:
Docusaurus V2 is a SPA - this makes deployment difficult. It needs to be deployed to a fileserver that automatically loads `index.html` when given a URL with a trailing slash. Also, it needs to be built with full knowledge of its `baseURL`. This forces us to deploy it to something more sophisticated than just a bare GCS bucket for a preview, which is annoying.
Previously I used vercel's preview feature to preview the Docusaurus site, however this makes it impossible to include generated API docs from Python or C++ in the deployed preview, since Vercel only knows about `node`.

Mkdocs is a pure static site generator, and lets us build/deploy much more flexibly. Also, since it doesn't use javascript routing, it lets us deploy other HTML artifacts into the URL path tree easily, making previewing generated API docs possible.

Agree the docusaurus default landing page is nicer. I believe we could do some work to polish up the `mkdocs` landing site to a similar level.

.gitignore
```diff
@@ -6,3 +6,4 @@ yarn-error.log
 .build
 __docs__
```
amacneil, line 7: maybe `__docs__` can be removed from here now? not sure if we still use it

amacneil:
How does `docs/CNAME` end up in the published tree?

mkdocs.yml
```diff
@@ -0,0 +1,86 @@
+site_name: "MCAP"
+theme:
+  name: material
+  favicon: img/favicon32.png
+  logo: img/mcap.png
+  palette:
+    # Palette toggle for dark mode
+    - scheme: slate
```
amacneil, line 8: Can we make it light mode by default?

mkdocs.yml
```diff
@@ -0,0 +1,86 @@
+site_name: "MCAP"
```
amacneil, line 1: possibly dumb question: why does this file live in the root of repo rather than in the `docs` folder?

james-rms: I'm not perfectly clear on why this must be, but I can say that it's considered an error if you try to put this file in the `docs` folder.
```
ERROR    -  Config value: 'docs_dir'. Error: The 'docs_dir' should not be the parent directory of the config file. Use a child
            directory instead so that the 'docs_dir' is a sibling of the config file.
Aborted with 1 Configuration Errors!
```
See also https://www.mkdocs.org/user-guide/writing-your-docs/#file-layout

.gitignore
james-rms: We still use it for generated API docs.

james-rms:
> How does docs/CNAME end up in the published tree?

Any file in `docs` that isn't markdown gets copied verbatim into the published tree.

docs/home/messages-channels-and-schemas.md
```diff
@@ -0,0 +1,19 @@
...
+
+MCAP organizes its data via messages, channels, and schemas.
+
+## Message
+
+The unit of communication between nodes in the pub/sub system.
+
+## Channel
+
+A stream of messages which have the same topic and type, or schema. This is often called a "topic" in pub/sub framework terminology.
```
james-rms, line 11: @esthersweon I updated this - previously it said that a channel often represents a "connection between a publisher and a subscriber", but this felt wrong to me. Pub/sub tries to abstract those connections away, and subscribers just "subscribe to a topic" rather than connecting directly to a subscriber.

mkdocs.yml
jtbandes: Time to switch Studio over as well? 😁

docs/home/messages-channels-and-schemas.md
jtbandes: We use `topic` to refer to the string field in the Channel record which specifies the topic that a subscriber would've subscribed to. However, there may be multiple Channels corresponding to the same topic. We inherited/adapted this from ROS where they are called "connections" because they represent a publisher–subscriber connection (in the case of a bag, the connection is between the publisher and the bag recorder)

jtbandes: On a more pedantic note, the proposed new wording is strange because it says that a topic is a stream of messages which have the same topic.

mkdocs.yml
amacneil: That is a debate as old as time 😄

In our community poll people seemed to favor dark mode. Let's at least start by keeping all our marketing sites light mode though.

mkdocs.yml
amacneil: 😭

mkdocs.yml
```diff
@@ -0,0 +1,87 @@
...
+        icon: material/brightness-7
+        name: Switch to dark mode
+      primary: "white"
+    - scheme: slate
+      toggle:
+        icon: material/brightness-4
+        name: Switch to light mode
+      primary: "black"
+  font:
+    text: Open Sans
```
jtbandes, line 19:
```suggestion
    text: IBM Plex Sans
    code: IBM Plex Mono
```

mkdocs.yml
```diff
@@ -0,0 +1,87 @@
...
+    - scheme: default
+      toggle:
+        icon: material/brightness-7
+        name: Switch to dark mode
+      primary: "white"
+    - scheme: slate
+      toggle:
+        icon: material/brightness-4
+        name: Switch to light mode
+      primary: "black"
```
jtbandes, line 17: Docs suggest that with the `media` settings, we can automatically adjust to the user's selected scheme when they first visit the site: https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#system-preference
```suggestion
    - media: "(prefers-color-scheme: light)"
      scheme: default
      toggle:
        icon: material/brightness-7
        name: Switch to dark mode
      primary: "white"
    - media: "(prefers-color-scheme: dark)"
      scheme: slate
      toggle:
        icon: material/brightness-4
        name: Switch to light mode
      primary: "black"
```

mkdocs.yml
```diff
@@ -0,0 +1,87 @@
...
+  - More Resources:
+      - GitHub: "https://github.com/foxglove/mcap"
+      - mcap CLI: "https://github.com/foxglove/mcap/tree/main/go/cli/mcap"
+      - Slack: "https://foxglovedev.slack.com"
+      - Library Support: "support-matrix.md"
+      - ROS 2 Storage Plugin: "https://github.com/ros-tooling/rosbag2_storage_mcap"
+      - Specification:
+          - "specification/README.md"
+          - "specification/explanatory-notes.md"
+          - "specification/appendix.md"
```
jtbandes, line 68: What do you all think about bringing the Specification section out towards the top level so it's not hidden away under "More Resources"?

mkdocs.yml
```diff
@@ -0,0 +1,87 @@
...
+      link: https://twitter.com/foxglovedev
+      name: Twitter
+    - icon: fontawesome/brands/slack
+      link: https://foxglovedev.slack.com
+      name: Slack
+    - icon: fontawesome/brands/github
+      link: https://github.com/foxglove/mcap
+      name: GitHub
+
+copyright: Copyright &copy; 2022 Foxglove
```
jtbandes, line 87: Should we include a note that mcap is MIT licensed?

docs/index.md
```diff
@@ -0,0 +1,43 @@
+# What is MCAP?
```
jtbandes, line 1: I don't think the main homepage at `/index.html` needs to start out with a question, it can just explain what MCAP is. Either:
```suggestion
# MCAP
```

or:
```suggestion
```

mkdocs.yml
```diff
@@ -0,0 +1,87 @@
+# Configuration for all pages
+site_name: "MCAP"
+theme:
+  name: material
+  favicon: img/favicon32.png
+  logo: img/mcap.png
```
jtbandes, line 6:
```suggestion
  logo: img/mcap.png
  icon:
    repo: fontawesome/brands/github
```

mkdocs.yml
```diff
@@ -0,0 +1,87 @@
...
+        name: Switch to dark mode
+      primary: "white"
+    - scheme: slate
+      toggle:
+        icon: material/brightness-4
+        name: Switch to light mode
+      primary: "black"
+  font:
+    text: Open Sans
+markdown_extensions:
```
jtbandes, line 20:
```suggestion
extra_css: extra.css

markdown_extensions:
```

Suggest adding this CSS to fix the cut-off github info:

image

```css
.md-source__fact--version {
  display: none;
}
.md-source__fact--stars::before {
  margin-left: 0;
}
.md-source__fact::before {
  vertical-align: bottom;
}
```

mkdocs.yml
```diff
@@ -0,0 +1,87 @@
...
+          - "guides/python/writing-jsonschema.md"
+          - "guides/python/reading-writing-protobuf.md"
+      - C++:
+          - "guides/cpp/writing-protobuf.md"
+  - API Reference:
+      - Python: docs/python/index.html
+      - C++: docs/cpp/index.html
+      - Typescript: docs/typescript/index.html
+      - Go: https://pkg.go.dev/github.com/foxglove/mcap/go/mcap
+      - Swift: docs/swift/documentation/mcap/
```
jtbandes, line 57:
```suggestion
      - Swift: docs/swift/documentation/mcap/index.html
```

Even with this change the page does not load. I think you will need to change the `--hosting-base-path` setting passed to the swift doc generator in pages.yml.

docs/getting-started/overview.md
```diff
@@ -0,0 +1,49 @@
...
+
+- [Python](https://github.com/foxglove/mcap/tree/main/python)
+- [C++](https://github.com/foxglove/mcap/tree/main/cpp)
+- [Go](https://github.com/foxglove/mcap/tree/main/go)
+- [Swift](https://github.com/foxglove/mcap/tree/main/swift)
+- [TypeScript](https://github.com/foxglove/mcap/tree/main/typescript)
+
+## Inspect and visualize MCAP
+
+Once you have MCAP data to work with, you may want to **use the mcap CLI to inspect and interact with your MCAP files**.
```
jtbandes, line 23: I can't find any explanation of how to install the CLI tool on the main page. All I found was `More Resources > mcap CLI` which takes me away from the documentation site onto GitHub.

mkdocs.yml
james-rms: @amacneil  I hid this per our previous slack thread, do you have an opinion here?

mkdocs.yml
james-rms: Not in the footer, but I think we can put it in the overview.

mkdocs.yml
```diff
@@ -0,0 +1,87 @@
...
+          - "guides/python/reading-writing-ros1.md"
+          - "guides/python/reading-writing-ros2.md"
+          - "guides/python/writing-jsonschema.md"
+          - "guides/python/reading-writing-protobuf.md"
+      - C++:
+          - "guides/cpp/writing-protobuf.md"
+  - API Reference:
+      - Python: docs/python/index.html
+      - C++: docs/cpp/index.html
+      - Typescript: docs/typescript/index.html
```
jtbandes, line 55:
```suggestion
      - TypeScript: docs/typescript/index.html
```

mkdocs.yml
```diff
@@ -0,0 +1,95 @@
...
+
+extra_css: [styles/extra.css]
+
+markdown_extensions:
+  - footnotes
+  - admonition
+  - pymdownx.highlight:
+      use_pygments: true
+  - pymdownx.superfences
+  - mdx_truly_sane_lists
```
jtbandes, line 34:
```suggestion
  - mdx_truly_sane_lists
  - smarty
```

docs/styles/extra.css
```diff
@@ -0,0 +1,9 @@
+.md-source__fact--version {
+    display: none;
+  }
+  .md-source__fact--stars::before {
+    margin-left: 0;
```
jtbandes, line 5: guess it wasn't important enough to work
```suggestion
    margin-left: 0 !important;
```

cspell.config.yaml
```diff
@@ -65,6 +67,10 @@ words:
   - zstandard
   - zstd
 
+languageSettings:
+  - languageId: "markdown"
+  - "ignoreRegExpList": ["/^\\s*```[\\s\\S]*?^\\s*```/gm"]
```
jtbandes, line 72: Does this really work? Some of our ignoreRegExpLists don't have slashes around them... are both syntaxes allowed?

Also, I think the `-` at the beginning of this line is wrong. I assume you wanted the ignoreRegExpList to be part of the settings with `languageId: markdown` and not a separate entry in languageSettings?

james-rms: it does work to ignore code blocks, though not sure if it's ignoring other things as well. Good catch with the extra `-`.

mkdocs.yml
james-rms: Even updating the base path doesn't fully work - in the end it looks like our preview hosting option not supported: https://github.com/apple/swift-docc/issues/323 I'm going to declare swift doc preview support out of scope of this PR.

Issue #370, opened 2022-04-11
x- prefix is recommended for Schemas, at odds with RFC 6648


jhurliman:
There is no justification given for the x- prefix recommendation. The only historical origin I can find for this convention is from HTTP, which deprecated the practice in 2012.

https://www.rfc-editor.org/rfc/rfc6648

alkasm:
The `x-` prefix is required in code by `mcap doctor`: https://github.com/foxglove/mcap/blob/7f8ac8f11c9ea78a057468379ad599da18d01f8a/go/cli/mcap/cmd/doctor.go#L239-L243

And still recommended in Python comments:
* https://github.com/foxglove/mcap/blob/a254fa07a7680856fdaf6136190dc8970c2045bd/python/mcap/mcap/mcap0/well_known.py#L5-L6
* https://github.com/foxglove/mcap/blob/2f51f478e16f219387df192b86925f760230e90e/python/mcap/mcap/mcap0/writer.py#L340-L342
* https://github.com/foxglove/mcap/blob/2f51f478e16f219387df192b86925f760230e90e/python/mcap/mcap/mcap0/writer.py#L367-L368

jtbandes:
https://github.com/foxglove/mcap/pull/521

PR #521, opened 2022-08-08
Remove more references to x- prefixes

jtbandes:
**Public-Facing Changes**
Removed outdated references to `x-` prefixes in examples and documentation.

**Description**
See #370 & #468.


cpp/examples/jsonschema/writer.cpp
```diff
@@ -39,8 +39,8 @@ int main(int argc, char** argv) {
 
   mcap::McapWriter writer;
   {
-    auto options = mcap::McapWriterOptions("x-jsonschema");
-    const auto res = writer.open(outputFilename, options);
+    auto options = mcap::McapWriterOptions("");
+    const auto res = writer.open(outputFilename);
```
wkalt, line 43: is there something that is still illustrative we can put instead of these empty strings? otherwise as a reader it's unclear why the API even needs these strings

jtbandes: What would be a sensical value to put in here?

> The profile is used for indicating requirements for fields throughout the file (encoding, user_data, etc). If the value matches one of the [well-known profiles](https://github.com/foxglove/mcap/blob/main/docs/specification/appendix.md#well-known-profiles), the file should conform to the profile. This field may also be supplied empty or containing a framework that is not one of those recognized.

Since this is just an example file I'm not sure there are any requirements we care to specify. And it's not really being written by a framework (although I thought the framework info is what would go in `library`, so not sure why we mention frameworks here)

python/examples/jsonschema/pointcloud_csv_to_mcap.py
```diff
@@ -79,7 +79,7 @@ def main():
     # tutorial-write-header-start
     with open(args.output, "wb") as f:
         writer = Writer(f)
-        writer.start("x-jsonschema", library="my-excellent-library")
+        writer.start("", library="my-excellent-library")
```
amacneil, line 82: Imo the writer shouldn't be setting their own library string in 99% of cases either. The library they are using to write the mcap file is our python library, so it should be automatically included.

amacneil (approved):

cpp/examples/jsonschema/writer.cpp
```diff
@@ -39,7 +39,7 @@ int main(int argc, char** argv) {
 
   mcap::McapWriter writer;
   {
-    auto options = mcap::McapWriterOptions("x-jsonschema");
+    auto options = mcap::McapWriterOptions("");
```
defunctzombie, line 42: I agree with @wkalt this empty string is weird. IMO this should use the options builder pattern: `mcap::WriterOptions().library("foobar")` or named struct initialization. Right now this example looks silly.

At the least we should put "my-custom-profile" or something meaningful here rather than an empty string.

python/mcap/tests/test_writer.py
```diff
@@ -17,7 +17,7 @@
 def generate_sample_data(compression: CompressionType):
     file = TemporaryFile("w+b")
     writer = Writer(file, compression=compression)
-    writer.start(profile="x-json", library="test")
+    writer.start(profile="", library="test")
```
defunctzombie, line 20: Ideally `profile` could be an optional argument?

defunctzombie (approved):
Not a fan of the c++ examples after this change. I think a placeholder should be used rather than an empty string at minimum. More broadly I don't think the API itself is good.

cpp/examples/jsonschema/writer.cpp
jtbandes: Is `my-custom-profile` meaningful? I invite someone to come up with a meaningful example that we can use in this code.

defunctzombie: In my mind it is one step above a required empty string /shrug :(

amacneil: I don't think we should encourage custom profiles. People will copy paste this code and think it is required. It is an optional field, and should in most cases only be used with existing well-known profiles.

I don't know C++ but surely we can have an optional option.

python/mcap/tests/test_writer.py
amacneil: +1

python/examples/jsonschema/pointcloud_csv_to_mcap.py
james-rms: Yep - this is better IMO.
```suggestion
        writer.start()
```

python/mcap/tests/test_writer.py
james-rms: It already is.

defunctzombie: I see this is a test not an example, so I don't feel as strongly but it would be nice to drop it entirely from here rather than use an empty string. Or leave `x-json` since again this is a test.

PR #524, opened 2022-08-12
cpp: use correct image for CI, remove --name

james-rms:
This commit removes the docker image called `mcap_cpp` from the C++
Makefile. This was being built erroneously in the `hdoc-build` target.

We don't have a need for specific names of containers in this process.
Also, the presence of containers with the same name as images caused
confusion and made this bug harder to spot. Therefore, this commit
removes the `--name` arguments from all `docker run` invocations in
the C++ Makefile.

**Public-Facing Changes**

**Description**

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -6,31 +6,32 @@ dev-image:
 
 .PHONY: build
 build: dev-image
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

emersonknapp (approved):
lgtm

PR #528, opened 2022-08-15
python: add option to order read messages by log time

james-rms:
**Public-Facing Changes**
Adds an option to `iter_messages` to read messages in log time order.

**Description**
On indexed, seekable MCAP sources, this implementation uses a priority queue to order messages in ascending `log_time` order. Chunks are decompressed when their `message_start_time` is the earliest in the queue.

On unindexed or non-seekable MCAP sources, this implementation loads all message records into memory and sort them by log time. This uses a lot of memory, but for non-seekable sources this is not avoidable. For non-indexed but seekable MCAP sources, this could be improved - we could read the log times into memory up-front but only read the content on-demand.

Fixes: #529


defunctzombie:
> Adds an option to `iter_messages` to read messages in log time order.

I find it count-intuitive that this option is _false_ by default. I'd expect this option to be true and the default behavior to produce messages in log time order.

python/mcap/mcap/mcap0/reader.py
```diff
@@ -150,31 +155,40 @@ def iter_messages(
...
-                    if topics is not None and channel.topic not in topics:
-                        continue
-                    if start_time is not None and record.log_time < start_time:
-                        continue
-                    if end_time is not None and record.log_time >= end_time:
-                        continue
-                    schema = summary.schemas[channel.schema_id]
-                    yield (schema, channel, record)
+            message_queue.push(chunk_index)
+        while message_queue.len() > 0:
```
jtbandes, line 174: pythonically this should be `while message_queue:`

or perhaps in python3.8+, `while next_item := message_queue.pop():`

python/mcap/mcap/mcap0/reader.py
```diff
@@ -251,7 +266,29 @@ def iter_messages(
...
+            setting log_time_order to True on a non-seekable stream will cause the entire content
+            of the MCAP to be loaded into memory.
         """
+        if not log_time_order:
+            for t in self._iter_messages_internal(topics, start_time, end_time):
+                yield t
+        else:
+            for t in sorted(
+                self._iter_messages_internal(topics, start_time, end_time),
+                key=lambda tup: tup[2].log_time,
```
jtbandes, line 282:
```suggestion
                key=lambda (schema, channel, message): message.log_time,
```

jtbandes (approved):

python/mcap/mcap/mcap0/reader.py
```diff
@@ -142,6 +146,7 @@ def iter_messages(
         topics: Optional[Iterable[str]] = None,
         start_time: Optional[int] = None,
         end_time: Optional[int] = None,
+        log_time_order: bool = False,
```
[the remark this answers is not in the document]
james-rms: I've changed the default to True, but for posterity: even for seekable streams you need to load all messages into memory if there is no summary section on the MCAP. This might constitute a performance problem, so I wanted the default to cause the least surprises.

python/mcap/mcap/mcap0/reader.py
james-rms: Pyright currently struggles to handle this syntax, not sure what's happening.

james-rms:
```
mcap/tests/test_reader.py:7: in <module>
    from mcap.mcap0.reader import make_reader, SeekingReader, NonSeekingReader, McapReader
E     File "/Users/j/code/mcap/python/mcap/mcap/mcap0/reader.py", line 291
E       key=lambda (schema, channel, message): message.log_time,
E                  ^
E   SyntaxError: invalid syntax
```

PR #530, opened 2022-08-15
Revert to old "Guides" titles and clean up dead page

esthersweon:
**Public-Facing Changes**
- Revert to old "Guides" titles and clean up dead page, as discussed with @james-rms


james-rms:
image
looks like the toplevel nav links in `mkdocs.yaml` need to be updated?

PR #534, opened 2022-08-17
go: read in log time order with overlapping chunks

### LINKED ISSUE #533: Go: Proper time ordering of indexed reads

james-rms:
**Public-Facing Changes**
Adds a `reverse` argument to McapReader.Messages(), letting a user read in reverse order.

Indexed MCAP reads will now return messages in log time order, even if there are chunks with overlapping time ranges.

# Description
This PR changes the approach to reading messages using the indexed reader. 

## Before
The previous implementation had one active chunk, and all messages were read out of it in ascending log time order before moving on to the next chunk. If two chunks are present in an MCAP with overlapping time ranges, this would cause messages to be read out-of-order.

## After
This PR changes to a scheme where there is one heap containing chunk indexes and message index entries that match the search criteria. 

When a chunk index reaches the top of the heap, the corresponding chunk is decompressed and all of its message index entries are pushed to the heap.

Message index entries in the heap are stored alongside a pointer to the underlying decompressed chunk. There is no "owning" pointer to the decompressed chunk data, so we now rely on the garbage collector to clean up the decompressed chunk once all message index entries are consumed.

Additionally, since it's easy enough, this PR adds a flag to iterate through the MCAP in reverse time order. This can be used to search for the first message on a topic before a given timestamp.
Fixes #533.

# TODO
- [x] Add reader test with overlapping chunks
- [x] Add test to read in reverse order

### DIFF
--- a/go/mcap/indexed_message_iterator.go
+++ b/go/mcap/indexed_message_iterator.go
@@ -2,10 +2,10 @@ package mcap
 
 import (
 	"bytes"
+	"container/heap"
 	"encoding/binary"
 	"fmt"
 	"io"
-	"sort"
 
 	"github.com/klauspost/compress/zstd"
 	"github.com/pierrec/lz4/v4"
@@ -15,13 +15,12 @@ import (
 // seeking is required). It makes reads in alternation from the index data
 // section, the message index at the end of a chunk, and the chunk's contents.
 type indexedMessageIterator struct {
-	lexer  *Lexer
-	rs     io.ReadSeeker
-	topics map[string]bool
-	start  uint64
-	end    uint64
-
-	chunk []byte
+	lexer   *Lexer
+	rs      io.ReadSeeker
+	topics  map[string]bool
+	start   uint64
+	end     uint64
+	reverse bool
 
 	channels          map[uint16]*Channel
 	schemas           map[uint16]*Schema
@@ -29,10 +28,7 @@ type indexedMessageIterator struct {
 	chunkIndexes      []*ChunkIndex
 	attachmentIndexes []*AttachmentIndex
 
-	// current location in the index
-	activeChunkIndex int // index of the active chunk within the set
-	messageOffsets   []MessageIndexEntry
-	messageOffsetIdx int
+	indexHeap rangeIndexHeap
 }
 
 // parseIndexSection parses the index section of the file and populates the
@@ -96,11 +92,16 @@ func (it *indexedMessageIterator) parseSummarySection() error {
 			if err != nil {
 				return fmt.Errorf("failed to parse attachment index: %w", err)
 			}
+			it.chunkIndexes = append(it.chunkIndexes, idx)
 			// if the chunk overlaps with the requested parameters, load it
 			for _, channel := range it.channels {
 				if idx.MessageIndexOffsets[channel.ID] > 0 {
 					if (it.end == 0 && it.start == 0) || (idx.MessageStartTime < it.end && idx.MessageEndTime >= it.start) {
-						it.chunkIndexes = append(it.chunkIndexes, idx)
+						rangeIndex := rangeIndex{
+							chunkIndex: idx,
+							reverse:    it.reverse,
+						}
+						heap.Push(&it.indexHeap, rangeIndex)
 					}
 					break
 				}
@@ -117,12 +118,7 @@ func (it *indexedMessageIterator) parseSummarySection() error {
 	}
 }
 
-func (it *indexedMessageIterator) loadNextChunk() error {
-	if it.activeChunkIndex >= len(it.chunkIndexes) {
-		return io.EOF
-	}
-	chunkIndex := it.chunkIndexes[it.activeChunkIndex]
-	it.activeChunkIndex++
+func (it *indexedMessageIterator) loadChunk(chunkIndex *ChunkIndex) error {
 	_, err := it.rs.Seek(int64(chunkIndex.ChunkStartOffset), io.SeekStart)
 	if err != nil {
 		return err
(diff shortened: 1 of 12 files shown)


go/mcap/range_index_heap.go
```diff
@@ -0,0 +1,41 @@
+package mcap
+
+type rangeIndex struct {
+	chunkIndex        *ChunkIndex
+	messageIndexEntry *MessageIndexEntry
+	reverse           bool
```
jtbandes, line 6: This should be the same for all entries in the heap, right? Maybe we can omit it from the heap entries themselves?

go/mcap/range_index_heap.go
```diff
@@ -0,0 +1,41 @@
...
+		return ri.chunkIndex.MessageStartTime
+	}
+	return ri.messageIndexEntry.Timestamp
+}
+
+type rangeIndexHeap []rangeIndex
+
+func (h rangeIndexHeap) Len() int      { return len(h) }
+func (h rangeIndexHeap) Swap(i, j int) { h[i], h[j] = h[j], h[i] }
+func (h *rangeIndexHeap) Push(x interface{}) {
```
jtbandes, line 24: I'm not a Go expert, but why not `(x rangeIndex)`?

Also, what's the point of this function? Shouldn't we be using `heap.Push` always?

go/mcap/range_index_heap.go
```diff
@@ -0,0 +1,41 @@
...
+	*h = append(*h, x.(rangeIndex))
+}
+
+func (h *rangeIndexHeap) Pop() interface{} {
+	old := *h
+	n := len(old)
+	x := old[n-1]
```
jtbandes, line 31: I'm confused by this because the documentation at https://pkg.go.dev/container/heap says:

>The minimum element in the tree is the root, at index 0.

jtbandes (commented):
Can we get some unit tests for this?
edit: I see this is already marked as TODO

go/mcap/range_index_heap.go
james-rms: Will add comments - Push and Pop are required methods of the  [heap.Interface](https://pkg.go.dev/container/heap#Interface) interface.

go/mcap/range_index_heap.go
james-rms: This also confuses me, since this is a verbatim copy from the IntHeap example there.

james-rms: This is confusing because it's the `Pop()` method of `heap.Interface`, which has the comment:  // remove and return element Len() - 1.

wkalt:
Rather than increasing the arity of Messages, would it make sense to switch to a "functional options" pattern as described in here? https://golang.cafe/blog/golang-functional-options-pattern.html

it seems like it would be nice for users not to care about reversing unless the require it

go/mcap/indexed_message_iterator.go
```diff
@@ -136,10 +130,34 @@ func (it *indexedMessageIterator) loadNextChunk() error {
...
+		if err != nil {
+			return fmt.Errorf("failed to read zstd chunk: %w", err)
+		}
+		defer reader.Close()
+		chunkData, err = io.ReadAll(reader)
+		if err != nil {
+			return fmt.Errorf("failed to decompress zstd chunk: %w", err)
+		}
+	case CompressionLZ4:
+		reader := lz4.NewReader(bytes.NewReader(parsedChunk.Records))
```
wkalt, line 149: reusing these readers across chunks can provide a speedup - also does this one need to be closed if not reused?

go/mcap/range_index_heap.go
```diff
@@ -0,0 +1,54 @@
+package mcap
+
+// rangeIndex contains either a ChunkIndex or a MessageIndexEntry to be sorted on LogTime.
```
wkalt, line 3: what's the rationale for this structure containing either chunk indexes or message indexes?

james-rms: To order them against eachother, and decide per-message if we need to read a message out of an existing chunk or decompress a new one.

go/mcap/indexed_message_iterator.go
james-rms: This is just moved code, but I can take a look at this 👍

james-rms: Yep - now reusing decoders for both zstd and lz4. Please note that:
* lz4 has no `Close()` functionality for its [Reader](https://pkg.go.dev/github.com/pierrec/lz4#Reader)
* zstd decoder has `Close()`, but the [documentation states](https://pkg.go.dev/github.com/klauspost/compress/zstd#Decoder.Close) that the decoder cannot be reused after `Close()` is called. I've taken the `defer Close()` out as a result.

go/mcap/indexed_message_iterator.go
```diff
@@ -96,11 +94,17 @@ func (it *indexedMessageIterator) parseSummarySection() error {
 			if err != nil {
 				return fmt.Errorf("failed to parse attachment index: %w", err)
 			}
+			it.chunkIndexes = append(it.chunkIndexes, idx)
```
wkalt, line 97: why move this up here?

wkalt: or if I understand right, is it needed at all? I guess this allows reuse of the reader in a way, since it would then have the full set of chunk indexes - is that the point?

go/mcap/indexed_message_iterator.go
```diff
@@ -212,22 +211,29 @@ func (it *indexedMessageIterator) Next(p []byte) (*Schema, *Channel, *Message, e
...
+	for it.indexHeap.Len() > 0 {
+		ri, err := it.indexHeap.HeapPop()
 		if err != nil {
 			return nil, nil, nil, err
 		}
+		if ri.messageIndexEntry == nil {
+			err := it.loadChunk(ri.chunkIndex)
+			if err != nil {
+				return nil, nil, nil, err
+			}
```
wkalt, line 224: I think we can eliminate a level of nesting here with a `continue`. Then the following `else` can be denested.

go/mcap/reader.go
```diff
@@ -95,38 +103,115 @@ func (r *Reader) unindexedIterator(topics []string, start uint64, end uint64) *u
...
+		end:      math.MaxInt64,
+		topics:   nil,
+		useIndex: true,
+		order:    ReadOrderFile,
+	}
+}
+
+type ReadOpt func(*readOptions) error
+
+func ReadMessagesAfter(start int64) ReadOpt {
```
wkalt, line 149: since these are supplied to a call to Messages, I wonder if `After`, `Before`, `WithTopics`, `WithIndex` (or `UsingIndex` I guess) might be nicer shortenings without any loss of information

wkalt (approved):
This looks good to me, just a couple more comments.

go/mcap/indexed_message_iterator.go
james-rms: I can revert this change. Previously i thought since `Info()` returns the chunkIndexes from here, it was a bug that the chunkIndexes in Info depended on the time / topic filter you passed in. However `Info()` constructs its own message iterator with no filters, so it's not a bug.

go/mcap/reader.go
james-rms: It's just a matter of cluttering the `mcap` module namespace. I'll see what I can do if I make a separate sub-package.

PR #545, opened 2022-08-18
doctor: report missing DataEnd/Footer

### LINKED ISSUE #535: `mcap doctor` reports no errors for corrupt mcap file
**Description**
> NOTE: I _believe_ this file is corrupt based on not being able to open it in studio with a bad magic error. I looked at the file in a hex viewer and the start of the file is ok - it is somewhere later in reading the file that Studio fails (possibly when it thinks there should be a footer record).

The file is our converted nuscenes.bag -> mcap located here: https://drive.google.com/drive/folders/10TFUJgvhOPqpSQ-q8tSbQum1CX33aO3L

When I run `mcap doctor` on this file there are no errors. But opening in Studio shows this error:

image

Studio uses the typescript reader so I imagine the result would be similar outside of studio for direct users of the typescript reader.

jtbandes:
**Public-Facing Changes**
`mcap doctor` now reports files missing their DataEnd or Footer records.

**Description**
Fixes #535 
Also updates the TypeScript `validate` script to report extra bytes remaining after parsing.

Ideally doctor would also validate trailing magic. I believe today it does not.

### DIFF
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -213,15 +213,24 @@ func (doctor *mcapDoctor) Examine() {
 	}
 
 	var lastMessageTime uint64
+	var lastToken mcap.TokenType
+	var dataEnd *mcap.DataEnd
+	var footer *mcap.Footer
 	msg := make([]byte, 1024)
 	for {
 		tokenType, data, err := lexer.Next(msg)
 		if err != nil {
 			if errors.Is(err, io.EOF) {
+				if dataEnd == nil {
+					doctor.error("File does not contain a DataEnd record (last record was %s)", lastToken.String())
+				} else if footer == nil {
+					doctor.error("File does not contain a Footer record (last record was %s)", lastToken.String())
+				}
 				break
 			}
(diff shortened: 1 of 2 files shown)


go/cli/mcap/cmd/doctor.go
```diff
@@ -213,15 +213,24 @@ func (doctor *mcapDoctor) Examine() {
...
+	var dataEnd *mcap.DataEnd
+	var footer *mcap.Footer
 	msg := make([]byte, 1024)
 	for {
 		tokenType, data, err := lexer.Next(msg)
 		if err != nil {
 			if errors.Is(err, io.EOF) {
+				if dataEnd == nil {
+					doctor.error("File does not contain a DataEnd record (last record was %s)", lastToken.String())
+				} else if footer == nil {
```
james-rms, line 226:
```suggestion
				}
				if footer == nil {
```

james-rms: Ideally there would be two errors if the file did not contain either a DataEnd or a Footer

james-rms (approved):

defunctzombie:
👍 awesome

PR #549, opened 2022-08-23
mcap-protobuf-support: fix duplicate FileDescriptors in FileDescriptorSet

jtbandes:
Fixed a bug in mcap-protobuf-support's `build_file_descriptor_set` where the protobuf schema was encoded with duplicate information in the case of a "diamond" import.


python/mcap-protobuf-support/mcap_protobuf/schema.py
```diff
@@ -13,16 +14,16 @@ def register_schema(writer: McapWriter, message_class: Any):
...
-        file_descriptor_set.file.append(proto)
-
-        for dep in fileDescriptor.dependencies:
-            append_file_descriptor(dep)
+    def append_file_descriptor(file_descriptor: FileDescriptor):
+        for dep in file_descriptor.dependencies:
+            if dep.name not in seen_dependencies:
+                seen_dependencies.add(dep.name)
+                append_file_descriptor(dep)
+        file_descriptor.CopyToProto(file_descriptor_set.file.add())
```
james-rms, line 26: why this rather than `append(proto)`?

jtbandes: No strong reason, just seemed like a simpler (potentially less allocation?) way of getting the bytes from one place to another. I didn't do any perf comparison though, just found that there was a handy CopyToProto method (which, now that I look at it, just calls ParseFromString under the hood)

Via https://developers.google.com/protocol-buffers/docs/reference/python-generated#repeated-message-fields:
> the corresponding Python object also has an add() method that creates a new message object, appends it to the list, and returns it for the caller to fill in. Also, the object's append() method makes a copy of the given message and appends that copy to the list

jtbandes: Okay, actually looking at the underlying implementations:

```py
  def add(self, **kwargs: Any) -> _T:
    """Adds a new element at the end of the list and returns it. Keyword
    arguments may be used to initialize the element.
    """
    new_element = self._message_descriptor._concrete_class(**kwargs)
    new_element._SetListener(self._message_listener)
    self._values.append(new_element)
    if not self._message_listener.dirty:
      self._message_listener.Modified()
    return new_element

  def append(self, value: _T) -> None:
    """Appends one element by copying the message."""
    new_element = self._message_descriptor._concrete_class()
    new_element._SetListener(self._message_listener)
    new_element.CopyFrom(value)
    self._values.append(new_element)
    if not self._message_listener.dirty:
      self._message_listener.Modified()
```

They look pretty much identical.

james-rms (approved):

PR #553, opened 2022-08-24
specification: add ros2idl to MCAP spec appendix

james-rms:
**Public-Facing Changes**
Adds `ros2idl` as a well-known schema encoding, as well as a more complete description of the concatenations scheme for `.msg` definitions.

**Description**

### DIFF
--- a/docs/specification/appendix.md
+++ b/docs/specification/appendix.md
@@ -65,7 +65,7 @@ Schema `encoding` may only be omitted for self-describing message encodings such
 
 ### ros1msg
 
-- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/PointCloud2`. The package resource name does not include the file extension.
+- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/PointCloud2`.
 - `encoding`: `ros1msg`
 - `data`: Delimited concatenated ROS1 .msg files.
 
@@ -76,7 +76,7 @@ the `data` field contains the concatenated `.msg` file content that is sent in t
 The top-level message definition is present first, with no delimiter. All dependent `.msg` definitions are preceded by a two-line delimiter:
 
 - One line containing exactly 80 `=` characters
-- One line containing `MSG: <package resource name>` for that type.
+- One line containing `MSG: <package resource name>` for that type. The space between `MSG:` and the package resource name is mandatory. The package resource name does not include a file extension.
 
 This format can be reproduced using [gendeps --cat](http://wiki.ros.org/roslib/gentools).
 
@@ -101,7 +101,7 @@ The `.msg` definition is stored alongside its dependencies in the same format as
 The IDL definition of the type specified by `name` along with all dependent types are stored together. The IDL definitions can be stored in any order. Every definition is preceded by a two-line delimiter:
 
 - a line containing exactly 80 `=` characters, then
-- A line containing only `IDL: <package resource name>` for that definition. The space between `IDL:` and the package resource name is mandatory.
+- A line containing only `IDL: <package resource name>` for that definition. The space between `IDL:` and the package resource name is mandatory. The package resource name does not include a file extension.
 
 ### jsonschema
 
--- a/docs/specification/appendix.md
+++ b/docs/specification/appendix.md
@@ -65,15 +65,43 @@ Schema `encoding` may only be omitted for self-describing message encodings such
 
 ### ros1msg
 
-- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/PointCloud2`
+- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/PointCloud2`. The package resource name does not include the file extension.
 - `encoding`: `ros1msg`
-- `data`: Concatenated ROS1 .msg files
+- `data`: Delimited concatenated ROS1 .msg files.
+
+#### ros1msg Data Format
+
+the `data` field contains the concatenated `.msg` file content that is sent in the ROS subscription connection header for this message type.
+
+The top-level message definition is present first, with no delimiter. All dependent `.msg` definitions are preceded by a two-line delimiter:
+
+- One line containing exactly 80 `=` characters
+- One line containing `MSG: <package resource name>` for that type.
+
+This format can be reproduced using [gendeps --cat](http://wiki.ros.org/roslib/gentools).
 
 ### ros2msg
 
-- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/msg/PointCloud2`
+- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/msg/PointCloud2`.
 - `encoding`: `ros2msg`
-- `data`: Concatenated ROS2 .msg files
+- `data`: Delimited concatenated ROS2 .msg files
+
+#### ros2msg Data Format
+
+The `.msg` definition is stored alongside its dependencies in the same format as [ros1msg](#ros1msg-data-format).
+
+### ros2idl
+
+- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/msg/PointCloud2`
+- `encoding`: `ros2idl`
+- `data`: Delimited concatenated ROS2 .idl files
+
+#### ros2idl Data Format
+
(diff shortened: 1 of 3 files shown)


docs/specification/appendix.md
```diff
@@ -118,4 +124,67 @@ The `ros2` profile describes how to create MCAP files for [ROS 2](https://docs.r
...
+- `encoding`: MUST contain either `ros2msg` or `ros2idl`
+
+## `gendeps`-style message definition delimiting
+
+For ROS1 and ROS2 message types, the message definition is stored in the [Schema record](./README.md#schema-op0x03) along with the definitions of any non-primitive field types. These definitions are concatenated together using a simple delimiting scheme:
+
+- The full content of the top-level message definition file is stored first.
+- Each dependent definition is preceded by:
+  - a line containing exactly 80 `=` characters, then
+  - A line containing only `MSG: <package resource name>` for that definition.
```
amacneil, line 136: for idl dependencies, do we expect the package resource name to be in the format `builtin_interfaces/msg/Time` or `builtin_interfaces/msg/Time.idl`? 

(would be good to specify this if it's not already covered somewhere)

docs/specification/appendix.md
```diff
@@ -118,4 +124,67 @@ The `ros2` profile describes how to create MCAP files for [ROS 2](https://docs.r
...
 
-- `encoding`: MUST contain `ros2msg`
+- `encoding`: MUST contain either `ros2msg` or `ros2idl`
+
+## `gendeps`-style message definition delimiting
+
+For ROS1 and ROS2 message types, the message definition is stored in the [Schema record](./README.md#schema-op0x03) along with the definitions of any non-primitive field types. These definitions are concatenated together using a simple delimiting scheme:
+
+- The full content of the top-level message definition file is stored first.
+- Each dependent definition is preceded by:
```
defunctzombie, line 134: One thing I've always been annoyed with in the ROS1 "combined message" format is that the first message does not have this "separator" information. I think that we should have even the first message (i.e. the entry message) be preceded by a separator with its file information.

An additional benefit is that when you start reading the file you immediately see that its not a regular .idl message but instead a special separate format.

defunctzombie: This comment is for the new `ros2idl` format - not for the `ros2msg` format.

docs/specification/appendix.md
```diff
@@ -118,4 +124,67 @@ The `ros2` profile describes how to create MCAP files for [ROS 2](https://docs.r
 
 #### Schema
 
-- `encoding`: MUST contain `ros2msg`
+- `encoding`: MUST contain either `ros2msg` or `ros2idl`
+
+## `gendeps`-style message definition delimiting
```
defunctzombie, line 129: I think this documentation should move to the respective `ros2msg` and `ros2idl` schema encoding sections. The profiles should just reference them by name.

docs/specification/appendix.md
james-rms: Seems a little hard to justify diverging from the others just for this schema encoding - if we're willing to diverge at all, why not just do something like a JSON list of objects?

docs/specification/appendix.md
james-rms: will document that we don't expect the file extension.

docs/specification/appendix.md
defunctzombie: > why not just do something like a JSON list of objects?

Now I need JSON library to read the schemas. It's nice to avoid that since these are text and what I am proposing is to add the leading separator so we have the same file and name information for that. In ROS1 there is no leading separator so you don't actually know the _datatype_ of the root message.

james-rms: you do know, it's in the schema `name`.

defunctzombie: From a function to parse this joined definitions its nice to be able to do `parseToAst(joinedText)` rather than `parseToAst(joinedText, "root_msg_datatype")` the latter we need to do today for ros1 messages because the root message metadata is not present.

docs/specification/appendix.md
```diff
@@ -65,15 +65,43 @@ Schema `encoding` may only be omitted for self-describing message encodings such
...
+
+#### ros2msg Data Format
+
+The `.msg` definition is stored alongside its dependencies in the same format as [ros1msg](#ros1msg-data-format).
+
+### ros2idl
+
+- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/msg/PointCloud2`
+- `encoding`: `ros2idl`
+- `data`: ROS2 .idl files, concatenated with [`gendeps`-style message definition delimiting](#gendeps-style-message-definition-delimiting)
```
defunctzombie, line 97: Is `#gendeps-style-message-definition-delimiting` suppose to reference something? I don't think that exists in this document anymore?

docs/specification/appendix.md
```diff
@@ -65,15 +65,43 @@ Schema `encoding` may only be omitted for self-describing message encodings such
...
+
+#### ros2idl Data Format
+
+The IDL definition of the type specified by `name` along with all dependent types are stored together. The IDL definitions can be stored in any order. Every definition is preceded by a two-line delimiter:
+
+- a line containing exactly 80 `=` characters, then
+- A line containing only `IDL: <package resource name>` for that definition.
```
defunctzombie, line 104: Might be worth giving an example of what a `package resource name` looks like.

Is the space after the `IDL:` required or optional?

defunctzombie (approved):
👍 from me. I'd get some more thumbs since this is a schema change and will be with us forever.

docs/specification/appendix.md
```diff
@@ -65,15 +65,43 @@ Schema `encoding` may only be omitted for self-describing message encodings such
 
 ### ros1msg
 
-- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/PointCloud2`
+- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/PointCloud2`. The package resource name does not include the file extension.
```
james-rms, line 68: @defunctzombie the package resource name definition is linked here.

docs/specification/appendix.md
james-rms: Out of date, let me remove.

docs/specification/appendix.md
defunctzombie: This is for `ros1msg`. It might be worth being explicit for `ros2idl` as well. I would almost operate like ros2 is a separate system from ros1.

jhurliman (approved):

amacneil (approved):
I was going to write a comment that I thought `ros2idl` diverging from `ros2msg` and `ros1msg` (gendeps style) was a mistake.

However, I read the arguments in this PR and thought more about it, and it seems like a good choice that will simplify parsing. And, you can think of it as still being gendeps-compatible, only the first "anonymous message definition" is blank.

In fact, we could change the spec to support making the first anonymous message definition optional (and recommended to be blank) for `ros2msg` as well as `ros2idl`, and change the behavior in `rosbag2_storage_mcap` to explicitly specify the `===\nMSG: ` header for all schemas.

docs/specification/appendix.md
```diff
@@ -65,15 +65,43 @@ Schema `encoding` may only be omitted for self-describing message encodings such
...
+- `data`: Delimited concatenated ROS1 .msg files.
+
+#### ros1msg Data Format
+
+the `data` field contains the concatenated `.msg` file content that is sent in the ROS subscription connection header for this message type.
+
+The top-level message definition is present first, with no delimiter. All dependent `.msg` definitions are preceded by a two-line delimiter:
+
+- One line containing exactly 80 `=` characters
+- One line containing `MSG: <package resource name>` for that type.
```
amacneil, line 79:
```suggestion
- One line containing `MSG: <package resource name>` for that type. The space between `MSG:` and the package resource name is mandatory.
```

may as well use consistent wording with IDL

docs/specification/appendix.md
```diff
@@ -65,15 +65,43 @@ Schema `encoding` may only be omitted for self-describing message encodings such
...
+- `name`: A valid [package resource name](http://wiki.ros.org/Names#Package_Resource_Names), e.g. `sensor_msgs/msg/PointCloud2`
+- `encoding`: `ros2idl`
+- `data`: Delimited concatenated ROS2 .idl files
+
+#### ros2idl Data Format
+
+The IDL definition of the type specified by `name` along with all dependent types are stored together. The IDL definitions can be stored in any order. Every definition is preceded by a two-line delimiter:
+
+- a line containing exactly 80 `=` characters, then
+- A line containing only `IDL: <package resource name>` for that definition. The space between `IDL:` and the package resource name is mandatory.
```
amacneil, line 104: To be clear, should `<package resource name>` include the file extension or no?

I assume not, since gendeps does not. However, that conflicts with the `#include` syntax in idl files, so I'm just curious.

emersonknapp (approved):

PR #558, opened 2022-08-25
specification: use literal syntax for generic types

james-rms:
Most markdown parsers (like the one used in mkdocs, our documentation generator) treat our "generic" syntax as arbitrary HTML tags, which they are not. GFM seems to treat this case specially, but so that `mcap.dev` can look nice, i'm putting backticks around them all.


jhurliman (approved):

jtbandes (approved):
Another option would be to use `&lt;`

PR #559, opened 2022-08-26
rust library, reading only

james-rms:
**Public-Facing Changes**
Adds an experimental, alpha, not to be used, rust library for parsing MCAP files.

**Description**
This is totally a proof-of-concept at this point. Writing is also the hard bit, and it's not done.

## TODO
- [ ] Come up with a better API for iterating through chunked MCAP files.
- [ ] make lz4 decompression safe
- [ ] add zstd decompression
- [ ] indexed reading
- [ ] writing
- [ ] tests

## Big questions
- why do we not emit MessageIndex records for the purposes of reader conformance tests?
- why is there a CRC in the spec for the attachment record, but none of our parsers look at it?


james-rms:
Closing in favor of #572

Issue #378, opened 2022-04-28
CLI tool outputs a partial file when ros2 conversion fails to find schemas


wkalt:
I think this must mean we're opening the output before reading the schemas. Even if the failure were partway through though, we should clean up any partial output.

PR #570, opened 2022-09-10
CLI: Support eloquent bagfiles

### LINKED ISSUE #567: `mcap convert` should support ROS 2 Eloquent bags
Here is a very simple bag file I created with ROS 2 Eloquent:

[rosbag2_2022_09_07-06_23_46_0.db3.zip](https://github.com/foxglove/mcap/files/9503297/rosbag2_2022_09_07-06_23_46_0.db3.zip)

I tried to convert it using `mcap convert`:

```sh
$ mcap version
v0.0.17
$ mcap convert rosbag2_2022_09_07-06_23_46_0.db3 out.mcap
failed to convert file: no such column: offered_qos_profiles
```

I believe this missing column is the only difference in Eloquent bags, so we should handle it gracefully.

Side note: It's weird that despite this error, `mcap convert` still generated a 176B `out.mcap` file. If there is a fatal error, I would not expect to create an output file.

wkalt:
These have a different schema in the topics table, which caused conversion to fail previously.

Fixes #567 

Fixes https://github.com/foxglove/mcap/issues/378

### DIFF
--- a/testdata/db3/eloquent-twist.db3
+++ b/testdata/db3/eloquent-twist.db3
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:8a7773ce9a299a115c01ad0f8125823d9b45603d75c9bcdfc1c1a611b3fc8bdb
+size 16384


wkalt:
Fixes https://github.com/foxglove/mcap/issues/567

go/ros/ros2db3_to_mcap.go
```diff
@@ -168,27 +168,58 @@ type topicsRecord struct {
...
+		return nil, err
+	}
+	var rows *sql.Rows
+	if hasQOSProfiles {
+		rows, err = db.Query(
+			`select id, name, type, serialization_format, offered_qos_profiles from topics`,
+		)
+	} else {
+		rows, err = db.Query(
+			`select id, name, type, serialization_format from topics`,
```
amacneil, line 197: can't you just `select *`?

wkalt: `select *` brings its own hassles. You need to land the values onto proper types, and the database library can't help you with that if you use `select *`, so you need to manually check for the existence of the columns you want, and verify they are the types you are expecting. Take a look at https://github.com/foxglove/mcap/pull/570/commits/b40cbc12935ccad81afd942951193cf23ef24f81 and see if you prefer it. After writing that I'm probably inclined to back it out and go back to two queries.

amacneil: Ok up to you, I haven't done go database stuff in a long time. If parsing the record is more work with `select *` then just having two queries seems fine.

james-rms (approved):

Issue #491, opened 2022-07-25
Rust library


amacneil:
Please 👍 and comment if you are interested in a native rust implementation of MCAP.

james-rms:
closed by #611 🎉

PR #572, opened 2022-09-12
Rust reader implementation

james-rms:
**Public-Facing Changes**
Merge the Rust implementation with minimal reading capability. The major contributions here are:

1. Creating struct definitions for all MCAP Records, in the `records` module
2. Adding functions to parse those from a byte buffer, in the `parse` module
3. Adding a simple state machine to copy them out of a `std::io::Read` implementation, in the `lexer` module
4. Wrap that in an ergonomic wrapper for iterating through the records of an MCAP in the `record_iterator` module.
5. Testing all of the above using the existing conformance test framework.

Additional infrastructure here includes:
* Setting up lint, build and test in CI
* Setting up `crates.io` publishing
* Setting up documentation generation and hosting.

Relates to: #491


rust/src/lexer.rs
```diff
@@ -0,0 +1,255 @@
...
+    expect_start_magic: bool,
+    attachment_content_handler: Option<AttachmentContentHandler<'a, R>>,
+}
+
+impl<'a, R: std::io::Read> Lexer<'a, R> {
+    /// Creates a new Lexer with default configuration values.
+    pub fn new(reader: R) -> Self {
+        Self {
+            reader: Some(reader),
+            last_opcode: None,
```
james-rms, line 105: remove this member

rust/src/lexer.rs
```diff
@@ -0,0 +1,255 @@
...
+                        // the next record.
+                        let mut limited_reader = self
+                            .reader
+                            .take()
+                            .unwrap()
+                            .take(len - (attachment_header_buf.len() as u64));
+                        // the handler's return value tells us whether to continue iterating through
+                        // the MCAP.
+                        if !cb(attachment_header, &mut limited_reader) {
+                            self.state = LexerState::End;
```
james-rms, line 192: restore reader here, just for completeness

rust/src/parse.rs
```diff
@@ -0,0 +1,383 @@
...
+        ParseError::StringEncoding(err)
+    }
+}
+
+fn parse_u16(data: &[u8]) -> Result<(u16, &[u8]), ParseError> {
+    if data.len() < std::mem::size_of::<u16>() {
+        return Err(ParseError::DataTooShort);
+    }
+    let (int_bytes, data) = data.split_at(std::mem::size_of::<u16>());
+    Ok((u16::from_le_bytes(int_bytes.try_into().unwrap()), data))
```
james-rms, line 45: expect

rust/src/parse.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+fn parse_u32(data: &[u8]) -> Result<(u32, &[u8]), ParseError> {
+    if data.len() < std::mem::size_of::<u32>() {
+        return Err(ParseError::DataTooShort);
+    }
+    let (int_bytes, data) = data.split_at(std::mem::size_of::<u32>());
+    Ok((u32::from_le_bytes(int_bytes.try_into().unwrap()), data))
```
james-rms, line 53: expect

rust/src/parse.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+pub(crate) fn parse_u64(data: &[u8]) -> Result<(u64, &[u8]), ParseError> {
+    if data.len() < std::mem::size_of::<u64>() {
+        return Err(ParseError::DataTooShort);
+    }
+    let (int_bytes, data) = data.split_at(std::mem::size_of::<u64>());
+    Ok((u64::from_le_bytes(int_bytes.try_into().unwrap()), data))
```
james-rms, line 61: expect

rust/src/record_iterator.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl<'a, R: std::io::Read> Iterator for RecordIterator<'a, R> {
+    type Item = Result<Record<'static>, LexError>;
+    fn next(&mut self) -> Option<Self::Item> {
+        // If in the middle of the chunk, continue iterating through that chunk.
+        if self.chunk_lexer.is_some() {
+            return match self
+                .chunk_lexer
+                .as_mut()
+                .unwrap()
```
james-rms, line 51: expect

rust/src/lexer.rs
```diff
@@ -0,0 +1,255 @@
...
+                    if let Some(mut cb) = self.attachment_content_handler.take() {
+                        let attachment_header_buf = read_attachment_header(self.get_reader()?)?;
+                        let attachment_header =
+                            parse_attachment_header(&attachment_header_buf[..])?;
+                        // hand the user a limited reader which reaches EOF at the beginning of
+                        // the next record.
+                        let mut limited_reader = self
+                            .reader
+                            .take()
+                            .unwrap()
```
james-rms, line 187: expect

rust/src/lexer.rs
```diff
@@ -0,0 +1,255 @@
...
+            Some(r) => Ok(r),
+            None => Err(LexError::CalledReadInCallback),
+        }
+    }
+}
+
+fn read_attachment_header<R: std::io::Read>(reader: &mut R) -> Result<Vec<u8>, LexError> {
+    let mut buf: Vec<u8> = vec![0; 8 + 8 + 4];
+    reader.read_exact(&mut buf[..])?;
+    let name_len = u32::from_le_bytes(buf[(buf.len() - 4)..].try_into().unwrap());
```
james-rms, line 246: expect

rust/src/lib.rs
```diff
@@ -0,0 +1,62 @@
+pub mod lexer;
+pub mod parse;
+pub mod record_iterator;
+pub mod records;
+
+#[cfg(test)]
+mod tests {
```
jtbandes, line 7: should this be in a file or directory called `test`/`tests`?

rust/src/parse.rs
```diff
@@ -0,0 +1,392 @@
+//! Functions for parsing byte buffers into their MCAP record contents.
+use crate::records::*;
+use std::borrow::Cow;
+use std::collections::BTreeMap;
```
jtbandes, line 4: Why BTreeMap?

image

rust/src/parse.rs
```diff
@@ -0,0 +1,392 @@
...
+    }
+}
+
+impl From<std::str::Utf8Error> for ParseError {
+    fn from(err: std::str::Utf8Error) -> Self {
+        ParseError::StringEncoding(err)
+    }
+}
+
+fn parse_u16(data: &[u8]) -> Result<(u16, &[u8]), ParseError> {
```
jtbandes, line 40: how did you choose the pattern of returning a new slice instead of wrapping this in some kind of Reader struct and mutating state via `self.bytes` or `self.offset` or somesuch? The constant re-binding of `data` variables below feels awkward, but I haven't read a lot of Rust before.

rust/src/parse.rs
```diff
@@ -0,0 +1,392 @@
...
+    if data.len() < std::mem::size_of::<u16>() {
+        return Err(ParseError::DataTooShort);
+    }
+    let (int_bytes, data) = data.split_at(std::mem::size_of::<u16>());
```
jtbandes, line 44: is this different from `let data = data[std::mem::size_of::<u16>()..];` (assuming some variant of this is valid...maybe it isn't)?

rust/src/parse.rs
```diff
@@ -0,0 +1,392 @@
...
+            let (key, data) = parse_str(unparsed_map_data)?;
+            let (val, data) = parse_str(data)?;
+            unparsed_map_data = data;
+            result.insert(key.into(), val.into());
+        }
+    }
+    Ok((result, remainder))
+}
+
+fn parse_header(data: &[u8]) -> Result<Header, ParseError> {
```
jtbandes, line 110: any hope of reducing boilerplate by doing these with a macro?

Would it make sense to make them static methods on the record types themselves, e.g. something like `Header::try_from(data: &[u8])`? That's what I did in the Swift library

rust/src/parse.rs
```diff
@@ -0,0 +1,392 @@
...
+        OpCode::SummaryOffset => Ok(Record::SummaryOffset(parse_summary_offset(data)?)),
+        OpCode::DataEnd => Ok(Record::DataEnd(parse_data_end(data)?)),
+        OpCode::UserOpcode(_) => Err(ParseError::OpCodeNotImplemented(opcode)),
+    }
+}
+
+/// Parses only the first fields of an Attachment record, before the `data` field. This
+/// allows a reader to read and parse only the header of an attachment before deciding whether
+/// to use the rest of the attachment.
+pub fn parse_attachment_header(data: &'_ [u8]) -> Result<AttachmentHeader<'_>, ParseError> {
```
jtbandes, line 379: does `'_` have a special meaning different from `'a`? I thought I had usually seen these lifetimes given letter names.

rust/src/record_iterator.rs
```diff
@@ -0,0 +1,83 @@
...
+use lifetime::IntoStatic;
+use std::io::Read;
+
+/// A wrapper for a [`std::io::Read`] implementation which parses out
+/// records and returns them from an iterator.
+pub struct RecordIterator<'a, R: std::io::Read> {
+    lexer: Lexer<'a, R>,
+    chunk_lexer: Option<Lexer<'a, std::io::Cursor<Vec<u8>>>>,
+    raw_record: Vec<u8>,
+    expand_chunks: bool,
```
jtbandes, line 16: are these public by default? I notice the struct is marked pub, bu the `impl Iterator` and `fn next` are not, which was a little surprising

rust/src/records.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub channel_message_counts: BTreeMap<u16, u64>,
+}
+
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Clone)]
+pub struct Metadata<'a> {
+    pub name: Cow<'a, str>,
+    pub metadata: BTreeMap<Cow<'a, str>, Cow<'a, str>>,
+}
+
+impl<'a> IntoStatic for Metadata<'a> {
```
jtbandes, line 186: What the heck is IntoStatic and how is it possible to make something static? Is this like constexpr?

rust/src/records.rs
```diff
@@ -0,0 +1,267 @@
...
+            0x07 => Ok(OpCode::MessageIndex),
+            0x08 => Ok(OpCode::ChunkIndex),
+            0x09 => Ok(OpCode::Attachment),
+            0x0A => Ok(OpCode::AttachmentIndex),
+            0x0B => Ok(OpCode::Statistics),
+            0x0C => Ok(OpCode::Metadata),
+            0x0D => Ok(OpCode::MetadataIndex),
+            0x0E => Ok(OpCode::SummaryOffset),
+            0x0F => Ok(OpCode::DataEnd),
+            x if x < 0x80 => Err(InvalidOpcode(x)),
```
jtbandes, line 53: good idea, we should probably add this check and the UserOpcode equivalent in other libraries. I don't think we have that today

rust/src/parse.rs
james-rms: Not a strong preference. I want the order of keys to be stable and well-defined, though I can't really put into words why - seems good for test determinism.

rust/src/parse.rs
james-rms: `'_` is the anonymous lifetime, code for "let lifetime elision figure out what to do with these lifetimes". Using named lifetimes here compiles, I actually only moved to this style when prompted by [clippy](https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes).

rust/src/record_iterator.rs
james-rms: you don't have to mark functions as pub when you're implementing a trait. Struct fields are private by default.

rust/src/records.rs
james-rms: A third-party trait: https://docs.rs/lifetime/latest/lifetime/trait.IntoStatic.html , used to simplify creating a "owned" form of all of our record structs.

It is possible to make some struct `<'a>` into `<'static>` if you can ensure that none of the fields are borrows that use that lifetime parameter. For example for a `Cow<'a, str>`, IntoStatic would be implemented something like:

```rust
  fn into_static(self) -> Cow<'static, str> {
      Cow::Owned(self.to_string())
  }
```

rust/src/parse.rs
jtbandes: Is there also some default behavior where you can omit the lifetimes entirely and it'll do the same thing by default?

rust/src/parse.rs
james-rms: you're right that the re-binding of data is awkward, i could take a look at using a mutable struct here.

rust/src/parse.rs
james-rms: This is just a nice way to bind both slices at once. lines 41-43 are necessary because both split_at and the slice operator will panic on an out-of-bounds index.

rust/src/parse.rs
james-rms: TryFrom isn't a bad idea, not sure why I didn't do that.

rust/src/parse.rs
james-rms: Good point, that was missing here.

README.md
```diff
@@ -20,6 +20,7 @@ MCAP libraries are provided in the following languages. For guidance on each lan
 | Python                | [readme](./python)     | [API docs](https://mcap.dev/docs/python)                        | `mcap`       | [![](https://shields.io/pypi/v/mcap)](https://pypi.org/project/mcap/)                |
 | JavaScript/TypeScript | [readme](./typescript) | [API docs](https://mcap.dev/docs/typescript)                    | `@mcap/core` | [![](https://shields.io/npm/v/@mcap/core)](https://www.npmjs.com/package/@mcap/core) |
 | Swift                 | [readme](./swift)      | [API docs](https://mcap.dev/docs/swift/documentation/mcap)      |              | see [releases](https://github.com/foxglove/mcap/releases)                            |
+| Rust                  |                        | [API docs](https://mcap.dev/docs/rust/doc/mcap)                 | `mcap`       | [![](https://shields.io/crates/v/mcap)](https://crates.io/crate/mcap)                |
```
amacneil, line 23: I suggest adding a rust readme and explaining the current features/limitations in it.

rust/src/records.rs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Clone, IntoStatic)]
+pub struct Chunk<'a> {
+    pub message_start_time: Timestamp,
+    pub message_end_time: Timestamp,
+    pub uncompressed_size: u64,
+    pub uncompressed_crc: u32,
+    pub compression: CowStr<'a>,
+    pub records: Cow<'a, [u8]>,
```
jtbandes, line 191: How does using Cow differ from just `&'a [u8]`?

rust/src/lexer.rs
```diff
@@ -0,0 +1,256 @@
...
+use crate::parse::{ParseError, Parser};
+use crate::records::{AttachmentHeader, InvalidOpcode, OpCode};
+use std::error::Error;
+
+pub const MAGIC: [u8; 8] = [0x89, b'M', b'C', b'A', b'P', 0x30, b'\r', b'\n'];
+
+#[derive(Debug)]
+pub enum LexError {
+    IO(std::io::Error),
+    ParseError(ParseError),
```
jtbandes, line 11: doesn't parsing come after lexing? How can a parse error become a lex error?

jtbandes (approved):

rust/src/records.rs
james-rms: Cow is an enum of `&'a [u8]` and `Vec<u8>` with some fancy sugar around it - using Cow will let someone construct a record by pushing to a `Vec<u8>` in the future. This will become more relevant when writing gets implemented.

rust/src/lexer.rs
james-rms: Yeah, maybe I should differentiate somehow. Here the ParseError occurs if we fail to parse a length field (needed for lexing).

PR #575, opened 2022-09-14
cli: doctor exits nonzero on failure

### LINKED ISSUE #356: Make mcap doctor return non-zero exit code on validation failure
`mcap doctor` currently returns a zero exit code on certain types of failures (e.g., the ones referenced in https://github.com/foxglove/studio/issues/3090). I'm adding an automated test to run `mcap doctor` over MCAP logs generated by my code and am currently going to just check that stderr is empty and stdout is exactly "Examining XYZ.mcap", but that's not the most robust method.

 If there's some other preferred way to test conformance, happy to use that instead. 

This is also probably within my own capability to implement, but just want to make sure there isn't a good reason not to do it first...

james-rms:
**Public-Facing Changes**
`mcap doctor` now exits non-zero if an error is found in the MCAP.

**Description**

Fixes #356

### DIFF
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -20,6 +20,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	doctorFailLevel string
+)
+
 type mcapDoctor struct {
 	reader io.ReadSeeker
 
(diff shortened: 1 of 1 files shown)


go/cli/mcap/cmd/doctor.go
```diff
@@ -55,6 +58,14 @@ func (doctor *mcapDoctor) fatalf(format string, v ...any) {
 	os.Exit(1)
 }
 
+func (doctor *mcapDoctor) exit() {
+	if doctor.errorCount == 0 {
+		os.Exit(0)
+	}
+	color.Red("Encountered %d errors", doctor.errorCount)
+	os.Exit(1)
+}
```
wkalt, line 67: I'm alright with this if it seems like the cleanest way to go for now. A couple notes FYI though -
* There's a function defined called `die` which logs an error to Stderr and then calls exit(1). Not sure if we can sensibly reuse that here given doctor using colors in contrast to the rest of the tool.
* Both this function and `die` are affected by a potential problem - they prevent any deferred functions from getting called. This may or may not cause issues in our usage, but it's something I've been meaning to take a closer look at.

wkalt (approved):
approved, but see comment

james-rms:
@wkalt i tried a different approach with the last commit, mind taking another look?

wkalt:
@james-rms that looks safer. It still could be affected if someone puts a deferred function in main(), outside the call to WithReader, since log.Fatal will also cause deferred functions not to execute, but it is at least clearly not a problem until someone does that. If they defer a function inside Run() it will be fine, and that's where I'd expect any new logic to go.

PR #576, opened 2022-09-14
conformance: enable data section crc

james-rms:
**Public-Facing Changes**
All writers now can compute data_section_crc and include it in the DataEnd record.

C++: separated the writer options for chunk, attachment, data, and summary CRCs.

**Description**
Conformance tests also updated to ensure correct data_section_crc.


cpp/mcap/include/mcap/writer.inl
```diff
@@ -331,6 +331,7 @@ void McapWriter::open(IWritable& writer, const McapWriterOptions& options) {
       chunkWriter->resetCrc();
     }
   }
+  writer.crcEnabled = compression_ == Compression::None && !options.noCRC;
```
jhurliman, line 334: CRC is not compatible with compression?

jtbandes: Didn't want to both with double-CRC when compression is enabled (compressed chunks already do their own crc). I think we've had some discussions in the past about only computing dataSectionCRC if chunking/compression is not enabled? Or maybe we need to add more writer options? Open to suggestions on what to do here.

james-rms: There's stuff outside of the chunks that needs to be covered by a CRC ie. `MessageIndex`, `Metadata` records, and the fields of the `Chunk` record that aren't the `records` field. I think this should just read:
```
writer.crcEnabled =  !options.noCRC;
```

jhurliman (changes requested):
CRCing the full Data section should only be done for unchunked files, at least in the C++ writer. Compression settings have nothing to do with it, it gets toggled on chunked vs unchunked.

jtbandes:
@jhurliman could you update the PR to fix the C++ implementation as you see fit?

james-rms:
>  CRCing the full Data section should only be done for unchunked files, at least in the C++ writer. Compression settings have nothing to do with it, it gets toggled on chunked vs unchunked.

If we want to force this on the user we should raise an error when they configured chunked writing with CRC. The specification says that the CRC covers the entire data section from the beginning to DataEnd, and we should follow the spec. Otherwise the user will be unpleasantly surprised when their indexes or metadata are corrupted but pass CRC validation.

jhurliman:
@james-rms it sounds like you are advocating for either feeding the contents of all `Chunk` records through two CRC passes, or disabling CRCs entirely, but no option for single-CRC coverage of `Chunk` records. Is that right?

jtbandes:
How about this:
- `options.noCRC` - rename to `options.noChunkCRC` - default false, true disables chunk CRCs
- add `options.noAttachmentCRC` - default false, true disables attachment CRCs
- add `options.enableDataCRC` - default false, true enables dataSectionCRC computation

james-rms:
@jtbandes That seems reasonable to me - happy to do the work if you're bored of it.

jhurliman:
> How about this:
> 
> * `options.noCRC` - rename to `options.noChunkCRC` - default false, true disables chunk CRCs
> * add `options.noAttachmentCRC` - default false, true disables attachment CRCs
> * add `options.enableDataCRC` - default false, true enables dataSectionCRC computation

I think this is a reasonable compromise. We should call out in the docs somewhere that if you disable chunking, Schema/Channel/Message records will no longer be CRC'd, but you can preserve that behavior with `enableDataCRC`.

jtbandes (approved):

jhurliman (approved):

cpp/mcap/include/mcap/writer.hpp
```diff
@@ -17,10 +17,21 @@ namespace mcap {
...
+   */
+  bool noAttachmentCRC = false;
+  /**
+   * @brief Enable CRC calculations for all records in the data section.
+   */
+  bool enableDataCRC = false;
+  /**
+   * @brief Disable CRC calculations for the summary section.
+   */
+  bool noSummaryCRC = false;
```
jhurliman, line 34: I assume https://github.com/ros-tooling/rosbag2_storage_mcap/blob/main/rosbag2_storage_mcap/src/mcap_storage.cpp#L117-L132 needs to change as well once this PR lands?

Issue #579, opened 2022-09-14
`mcap info` crashes


defunctzombie:
Run `mcap info` on the attached file. My version crashes.

[output.mcap.zip](https://github.com/foxglove/mcap/files/9568846/output.mcap.zip)

```
 mcap info output.mcap
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x104af5c20]

goroutine 1 [running]:
github.com/foxglove/mcap/go/cli/mcap/cmd.printInfo({0x104ed22e8, 0x14000010018}, 0x140003aac30)
	/private/tmp/mcap-20220802-52309-1h62fv4/mcap-releases-mcap-cli-v0.0.16/go/cli/mcap/cmd/info.go:99 +0xd00
github.com/foxglove/mcap/go/cli/mcap/cmd.glob..func7.1(0xe2?, {0x12c58db78?, 0x1400039e2a0})
	/private/tmp/mcap-20220802-52309-1h62fv4/mcap-releases-mcap-cli-v0.0.16/go/cli/mcap/cmd/info.go:138 +0xd8
github.com/foxglove/mcap/go/cli/mcap/utils.WithReader({0x104ed7278, 0x1400003c0c0}, {0x16b9d77e2?, 0x0?}, 0x104ecad80)
	/private/tmp/mcap-20220802-52309-1h62fv4/mcap-releases-mcap-cli-v0.0.16/go/cli/mcap/utils/utils.go:99 +0x358
github.com/foxglove/mcap/go/cli/mcap/cmd.glob..func7(0x1053de2c0?, {0x1400039d290?, 0x1?, 0x1?})
	/private/tmp/mcap-20220802-52309-1h62fv4/mcap-releases-mcap-cli-v0.0.16/go/cli/mcap/cmd/info.go:129 +0xa8
github.com/spf13/cobra.(*Command).execute(0x1053de2c0, {0x1400039d250, 0x1, 0x1})
	/Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:876 +0x4b8
github.com/spf13/cobra.(*Command).ExecuteC(0x1053dd140)
	/Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:990 +0x354
github.com/spf13/cobra.(*Command).Execute(...)
[...]

PR #584, opened 2022-09-15
cli: info: don't crash when channel has no schema

jtbandes:
**Public-Facing Changes**
Fixed a bug where `mcap info` would crash if a channel had no schema.

**Description**
Fixes https://github.com/foxglove/mcap/issues/579

### DIFF
--- a/go/cli/mcap/cmd/info.go
+++ b/go/cli/mcap/cmd/info.go
@@ -101,6 +101,8 @@ func printInfo(w io.Writer, info *mcap.Info) error {
 		}
 		if schema != nil {
 			row = append(row, fmt.Sprintf(" : %s [%s]", schema.Name, schema.Encoding))
+		} else if channel.SchemaID != 0 {
+			row = append(row, fmt.Sprintf(" : <missing schema %d>", channel.SchemaID))
 		} else {
 			row = append(row, " : <no schema>")
 		}
--- a/go/cli/mcap/cmd/info.go
+++ b/go/cli/mcap/cmd/info.go
@@ -87,10 +87,7 @@ func printInfo(w io.Writer, info *mcap.Info) error {
 	}
 	for _, chanID := range chanIDs {
 		channel := info.Channels[chanID]
-		var schema *mcap.Schema
-		if channel.SchemaID != 0 {
-			schema = info.Schemas[channel.SchemaID]
-		}
+		schema := info.Schemas[channel.SchemaID]
 		channelMessageCount := info.Statistics.ChannelMessageCounts[chanID]
 		frequency := 1e9 * float64(channelMessageCount) / float64(end-start)
 		row := []string{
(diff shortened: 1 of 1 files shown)


go/cli/mcap/cmd/info.go
```diff
@@ -96,7 +99,11 @@ func printInfo(w io.Writer, info *mcap.Info) error {
 		if info.Statistics != nil {
 			row = append(row, fmt.Sprintf("%*d msgs (%.2f Hz)", maxCountWidth, channelMessageCount, frequency))
 		}
-		row = append(row, fmt.Sprintf(" : %s [%s]", schema.Name, schema.Encoding))
+		if schema != nil {
+			row = append(row, fmt.Sprintf(" : %s [%s]", schema.Name, schema.Encoding))
+		} else {
+			row = append(row, " : <no schema>")
+		}
```
wkalt, line 106: You'll want to test to verify, but I think you only need the change from 102-106 to get the behavior you want. If there is no schema for the channel's schema ID, then schema is going to be nil when you look it up in the map, even without the guard on 91.

If for some reason someone put a schema into the file with ID 0, which would be a spec violation, the code as written here would obscure that schema rather than display it. I think it would probably be better to display it in that case, and instead warn in doctor (which I think we do).

jtbandes: >If for some reason someone put a schema into the file with ID 0, which would be a spec violation, the code as written here would obscure that schema rather than display it. I think it would probably be better to display it in that case, and instead warn in doctor (which I think we do).

When I was writing this I thought about it the opposite way: if someone makes a channel with schema id 0, the spec says that's a schemaless channel, so we shouldn't display the schema even if there happens to be one in the map. I can see this argument too though.

wkalt: I see.

IMO, if "doctoring" is printing warnings about spec violations like the doctor tool does, I think the info command shouldn't do it - otherwise we would want to make it robust to as many kinds of badly-formed files as possible, to the point that the "doctor" command would be redundant - it's not really clear where we would want to draw the line.

I think the point of the "info" and "list schemas" commands is mainly introspection of the file contents, so IMO it's not helpful for them to make aspects of the file contents inaccessible to the user. If the user has recorded a file with zero-ID schemas, and the doctor command warns/errors on it, ideally they would be able to use the CLI subcommands to figure out exactly what the problem is, rather than needing to go parse the file manually. Doctor currently will only give them this to go on: https://github.com/foxglove/mcap/blob/main/go/cli/mcap/cmd/doctor.go#L149,L151.

We could make that a bit clearer admittedly, but currently the only way they would be able to get information about their bad schema would be through the `mcap info` or `mcap list schemas` commands. If we eagerly replace the schemas with `<no schema>` in those cases, seems like we are making life more difficult to me.

jtbandes: Ok, I reverted part of this to allow schemas with id 0. I also adjusted the output so it distinguishes id 0 ("no schema") from another nonzero id that was not found ("missing schema").

wkalt (approved):

PR #586, opened 2022-09-18
Remove pre0 code

defunctzombie:
**Public-Facing Changes**

@mcap/core no longer has pre0 readers or writers

**Description**

There are no pre0 files or users anymore, we can get rid of this code.


typescript/core/src/common/detectVersion.ts
```diff
@@ -1,7 +1,6 @@
-import { MCAP_MAGIC } from "../pre0/constants";
 import { MCAP0_MAGIC } from "../v0/constants";
 
-export type McapVersion = "pre0" | "0";
+export type McapVersion = "0";
```
jtbandes, line 3: Should we remove this type and replace detectVersion with something like detectMcap or validateMagic? The "detect version" wording doesn't make much sense if we only acknowledge the existence of one version.

typescript/core/src/index.ts
```diff
@@ -1,7 +1,3 @@
-export { default as McapPre0Reader } from "./pre0/McapPre0Reader";
-export { default as McapPre0Writer } from "./pre0/McapPre0Writer";
-export * as McapPre0Types from "./pre0/types";
-
 export { Mcap0IndexedReader } from "./v0/Mcap0IndexedReader";
 export { default as Mcap0StreamReader } from "./v0/Mcap0StreamReader";
 export { Mcap0Writer } from "./v0/Mcap0Writer";
```
jtbandes, line 3: We should probably also remove the `v0` parts from all these types/paths?

defunctzombie: I think we should remove these parts yea.

typescript/core/src/common/detectVersion.ts
defunctzombie: Yea - I think we should simplify this down to one version. The v0 idea was compelling but its the version we are gonna have for a while. One day we can introduce the subpaths if we get a v1.

defunctzombie:
Closing in favor of https://github.com/foxglove/mcap/pull/602

PR #597, opened 2022-09-22
Remove lodash dependency from @mcap/core

### LINKED ISSUE #588: Typescript mcap library requires lodash but does not depend on it
https://github.com/foxglove/mcap/blob/main/typescript/core/src/v0/ChunkCursor.ts#L2

Installing @mcap/core and trying to use the v0 library results in errors because `lodash` is not listed as a dependency. I should be able to install `@mcap/core` as a dependency without having to install additional dependencies manually.

jhurliman:
**Public-Facing Changes**
- TypeScript library `@mcap/core` can be used without manually installing `lodash`

**Description**
lodash was used in a few places but there was no dependency on it in package.json. This replaces the few uses of lodash generic  methods with more optimized functions tuned for the @mcap/core library. In particular, it avoids heap allocation of new string  objects during binary searches.

Fixes #588

### DIFF
--- a/typescript/core/src/pre0/parse.ts
+++ b/typescript/core/src/pre0/parse.ts
@@ -191,12 +191,24 @@ export function parseRecord(
 }
 
 function isChannelInfoEqual(a: ChannelInfo, b: ChannelInfo): boolean {
-  return (
-    a.data.byteLength === b.data.byteLength &&
-    a.encoding === b.encoding &&
-    a.id === b.id &&
-    a.schema === b.schema &&
-    a.schemaName === b.schemaName &&
-    a.topic === b.topic
-  );
+  if (
+    a.data.byteLength !== b.data.byteLength ||
+    a.encoding !== b.encoding ||
+    a.id !== b.id ||
+    a.schema !== b.schema ||
+    a.schemaName !== b.schemaName ||
+    a.topic !== b.topic
+  ) {
+    return false;
+  }
+
+  const aData = new Uint8Array(a.data);
+  const bData = new Uint8Array(b.data);
+  for (let i = 0; i < aData.length; i++) {
+    if (aData[i] !== bData[i]) {
+      return false;
+    }
+  }
+
+  return true;
 }
--- a/typescript/core/src/pre0/parse.ts
+++ b/typescript/core/src/pre0/parse.ts
@@ -1,5 +1,3 @@
-import { isEqual } from "lodash";
-
 import { getBigUint64 } from "../common/getBigUint64";
 import { MCAP_MAGIC, RecordType } from "./constants";
 import { McapMagic, McapRecord, ChannelInfo } from "./types";
@@ -122,7 +120,7 @@ export function parseRecord(
       channelInfosSeenInThisChunk.add(id);
       const existingInfo = channelInfosById.get(id);
       if (existingInfo) {
-        if (!isEqual(existingInfo, record)) {
+        if (!isChannelInfoEqual(existingInfo, record)) {
           throw new Error(`differing channel infos for ${record.id}`);
         }
         return {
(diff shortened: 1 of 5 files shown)


typescript/core/src/v0/ChunkCursor.ts
```diff
@@ -245,7 +244,7 @@ export class ChunkCursor {
           startIndex = sortedIndexBy(
             result.record.records,
             [this.startTime],
-            ([logTime]) => logTime,
+            ([logTime]) => logTime!,
```
defunctzombie, line 247: You might be able to specialize sortedIndexBy <T> value to expect a tuple rather than an array and that would let you drop the `!` if that's interesting to avoid errors in the future with type changes.

typescript/core/src/v0/ChunkCursor.ts
```diff
@@ -292,3 +291,28 @@ export class ChunkCursor {
...
+  let high = array == null ? 0 : array.length;
+  if (high === 0) {
+    return 0;
+  }
+
+  const valueNumber = iteratee(value);
+  const valIsNaN = valueNumber !== valueNumber;
+
+  while (low < high) {
+    const mid = Math.floor((low + high) / 2);
```
jtbandes, line 307: You might want to read about the adjustment to the math made in: https://www.npmjs.com/package/binary-search

typescript/core/src/pre0/parse.ts
```diff
@@ -191,3 +189,14 @@ export function parseRecord(
       throw new Error("Not yet implemented");
   }
 }
+
+function isChannelInfoEqual(a: ChannelInfo, b: ChannelInfo): boolean {
+  return (
+    a.data.byteLength === b.data.byteLength &&
```
jtbandes, line 195: Don’t we need to compare the data bytes as well?

typescript/core/src/v0/Mcap0StreamReader.ts
```diff
@@ -258,3 +263,24 @@ export default class Mcap0StreamReader implements McapStreamReader {
...
 }
+
+function isChannelEqual(a: Channel, b: Channel): boolean {
+  if (
+    !(
+      a.id === b.id &&
+      a.messageEncoding === b.messageEncoding &&
+      a.schemaId === b.schemaId &&
+      a.topic === b.topic &&
+      a.metadata.size === b.metadata.size
```
defunctzombie, line 274: Is there anything we can do to protect against adding fields but forgetting to define them here for equality comparison?

Random thought: might it be better to store a crc of the original record bytes and compare that?

typescript/core/src/pre0/parse.ts
```diff
@@ -191,3 +189,14 @@ export function parseRecord(
       throw new Error("Not yet implemented");
   }
 }
+
+function isChannelInfoEqual(a: ChannelInfo, b: ChannelInfo): boolean {
```
defunctzombie, line 193: Similar thoughts to the channel info comparison below - maybe a crc on the bytes would be effective and work if we ever add a new field.

defunctzombie (dismissed):
👍 to dropping lodash. Less thrilled about the untested `sortedIndexBy`. This is the type of method we should have tests for and why we use libraries that have these methods tested.

typescript/core/src/v0/sortedIndexBy.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    expect(sortedIndexBy(array, 0n, (x) => x)).toEqual(0);
+    expect(sortedIndexBy(array, 1n, (x) => x)).toEqual(0);
+    expect(sortedIndexBy(array, 2n, (x) => x)).toEqual(1);
+    expect(sortedIndexBy(array, 3n, (x) => x)).toEqual(1);
+    expect(sortedIndexBy(array, 4n, (x) => x)).toEqual(2);
+  });
+
+  it("handles negation", () => {
+    const array: [bigint, bigint][] = [
```
jtbandes, line 39: This test seems invalid? The input array is not sorted by (x)=>-x.

typescript/core/src/v0/sortedIndexBy.ts
```diff
@@ -0,0 +1,29 @@
...
+  let low = 0;
+  let high = array.length;
+  if (high === 0) {
+    return 0;
+  }
+
+  const computedValue = iteratee(value);
+
+  while (low < high) {
+    const mid = (low + high) >>> 1;
```
jtbandes, line 19: The other math adjustment I was referring to was low + (high - low) >>>1, to avoid overflow during addition

PR #599, opened 2022-09-22
cli: add recover & compress commands

### LINKED ISSUE #446: mcap recover command
We want a way for CLI users to fix mcaps that are incomplete or corrupted, but fixable. CLI is fluid at this point - could do one of:

```
mcap fix FILE [--output OUTFILE] [--in-place] [--destructive]
mcap doctor FILE [--fix] [--output OUTFILE] [--in-place] [--destructive]
mcap recover FILE [--output OUTFILE] [--in-place] [--destructive]
```

### LINKED ISSUE #481: CLI: mcap compress subcommand
Currently you can recompress a file using the filter or merge commands, but to a user who wants to recompress these aren't very discoverable. Rosbag provides a "compress" command that does more or less what we would need - although it allows you to compress multiple files in one call, which we could maybe consider dropping (not clear if it's the tool's job or should be done in bash).

jtbandes:
**Public-Facing Changes**
Added command `mcap recover` for recovering data from truncated MCAP files.
Added command `mcap compress` for changing the compression in an MCAP file.

**Description**
Both new commands share core implementation with `mcap filter`, with different implied options.

Resolves #481 
Resolves #446


jhurliman:
Does `mcap recover` reorder messages so `log_time` is monotonic, or does it preserve file order?

jtbandes:
I believe all of these commands currently preserve file order, since they use the Lexer rather than Reader.

james-rms:
I wonder if we could expose a more general-purpose subcommand `mcap rewrite`, which would cover `mcap compress` and any future modifications such as adding message indexes, adding summary data, changing message encoding, you name it.

jhurliman:
> such as adding message indexes, adding summary data

I assumed `mcap recover` is already doing this since the PR description says it is for recovering data from truncated MCAP files. What does the command actually do? Do chunks with invalid CRCs get discarded, and are indexes rebuilt?

jtbandes:
The general-purpose `mcap rewrite` is basically what we have already with `mcap filter` (this PR just adds a recover option to it). But having separate commands/aliases for `mcap recover` and `mcap compress` seems like better SEO (easier to search on the internet, easier to find when reading `mcap help`, etc.)

defunctzombie:
My take is that separating recover from other transformations is nice from a discovery and ux standpoint. `recover` makes it clear the intent and type of file I am expecting to operate on.

go/cli/mcap/cmd/filter.go
```diff
@@ -207,6 +204,17 @@ func filter(
...
 				return nil
 			}
+			if opts.recover && errors.Is(err, io.ErrUnexpectedEOF) {
+				fmt.Println("Input file was truncated.")
+				return nil
+			}
+			if opts.recover && token == mcap.TokenInvalidChunk {
+				fmt.Printf("Invalid chunk encountered, skipping: %s\n", err)
+				token, data, err = lexer.Next(buf)
+				fmt.Printf("NEXT ERR: %s\n", err)
```
wkalt, line 214: this line needs refinement. if these should be reported, maybe collect and report at the end?

wkalt: we should probably also change the printing in this patch to use stderr like line 189 does

jtbandes: oops, left this debug stuff in, thanks

jtbandes: My rationale for printf was that if `mcap recover` is intended to be a recovery tool, it's okay to print errors to stdout. But can change to stderr if you prefer.

jtbandes:
@wkalt are you okay with the lexer changes in this PR?

wkalt:
@jtbandes lexer changes look fine

amacneil:
product review feedback:
- PR good to merge
- add `decompress` command
- add `compress`/`decompress`/`ignore-errors` flags to the catch all command stream processing command (`filter`)

jhurliman (approved):

Issue #569, opened 2022-09-08
Remove mcap0 from typescript lib


amacneil:
The mcap0 stuff is from an earlier time when we thought we might need a major revision to the spec before calling it stable. As things happened we decided to call the current version stable. Only Python and TS libraries refer to `Mcap0`.

PR #602, opened 2022-09-22
Remove pre0 support from typescript library

jhurliman:
**Public-Facing Changes**
- Support for the alpha version of MCAP files, pre0, has been removed from the typescript library

**Description**
This is a prerequisite for #569


typescript/core/src/common/detectVersion.ts
```diff
@@ -13,9 +12,6 @@ export function detectVersion(prefix: DataView): McapVersion | undefined {
   if (prefix.byteLength < DETECT_VERSION_BYTES_REQUIRED) {
     return undefined;
   }
-  if (MCAP_MAGIC.every((val, i) => val === prefix.getUint8(i))) {
-    return "pre0";
-  }
   if (MCAP0_MAGIC.every((val, i) => val === prefix.getUint8(i))) {
     return "0";
   }
```
jtbandes, line 17: Same comment as I made on #586, this McapVersion concept and version detection should be removed or replaced with something that doesn't reference versions if we only support one version.

typescript/core/src/common/hasMcapPrefix.ts
```diff
@@ -0,0 +1,9 @@
+import { MCAP0_MAGIC } from "../v0/constants";
```
jtbandes, line 1: can we also
- rename `MCAP0_MAGIC` to `MCAP_MAGIC`
- rename `DETECT_VERSION_BYTES_REQUIRED` to something else like `HAS_MCAP_PREFIX_BYTES_REQUIRED` (or maybe delete it, and just use `MCAP_MAGIC.length`?)
- remove the `v0` folder, etc.

jtbandes (approved):

typescript/examples/validate/scripts/validate.ts
```diff
@@ -225,94 +224,86 @@ async function validate(
...
-          for (const record of reader.schemasById.values()) {
-            processRecord(record);
-          }
-          for (const record of reader.channelsById.values()) {
-            processRecord(record);
-          }
-          for await (const record of reader.readMessages()) {
-            processRecord(record);
-          }
-          break;
```
jtbandes, line 298: I'm a little confused by the removed `break` — does that not have some effect (skipping over the `await readStream`)?

PR #610, opened 2022-09-26
[TypeScript] Rename all MCAP0 references to MCAP

jhurliman:
**Public-Facing Changes**
- TypeScript API naming overhaul to replace all "Mcap0*" references with "Mcap*"

**Description**
This is an API-breaking change to rename all "Mcap0*" classes to "Mcap*" by changing the folder structure, file names, class names, and constant names.


typescript/core/src/TempBuffer.ts
```diff
@@ -1,4 +1,4 @@
-import { IWritable } from "../common/IWritable";
+import { IWritable } from "./common/IWritable";
```
jtbandes, line 1: I think the `common` directory can go away too. I believe it was meant to be common code between pre0 and v0, so the distinction is not necessary anymore.

typescript/core/src/types.ts
```diff
@@ -129,7 +129,7 @@ type Values<T> = T[keyof T];
 export type TypedMcapRecord = Values<TypedMcapRecords>;
 export type McapRecord = Values<McapRecords>;
 
-export interface McapStreamReader {
+export interface IMcapStreamReader {
```
jtbandes, line 132: I don't recall why we had this interface but I suspect we can remove it, since we are exporting the McapStreamReader class anyway...

PR #611, opened 2022-09-26
Rust: Port anduril/mcap-rs

mrkline:
**Public-Facing Changes**

- Upstreams [Anduril's Rust library](https://github.com/anduril/mcap-rs), 

**Description**

Adds a more fully-featured Rust library that supports indexed reading, CRC validation, record writing, chunking, and automatic summary writing.

**TODOs**

- Spelling (#618)
- We probably want to make `serde` an optional dependency that we only pull in for the conformance reader. (I can't imagine users wanting to serialize random chunks of an MCAP file to JSON/CBOR/Bincode/etc.)
- The round-trip test (`tests/round_trip.rs`) checks the message indexes by individually seeking each message from its respective chunk. This is... slow. Should we axe that part of the test to avoid bogging down CI?


rust/src/records.rs
```diff
@@ -1,507 +1,537 @@
...
+    Channel(Channel),
+    Message {
+        #[serde(flatten)]
+        header: MessageHeader,
+        data: Cow<'a, [u8]>,
+    },
+    Chunk {
+        #[serde(flatten)]
+        header: ChunkHeader,
+        data: &'a [u8],
```
james-rms, line 65: why no cow here?

rust/src/records.rs
```diff
@@ -1,507 +1,537 @@
...
+    #[br(map = |s: McapString| s.inner )]
+    #[bw(write_with = write_string)]
+    pub name: String,
+
+    #[br(map = |s: McapString| s.inner )]
+    #[bw(write_with = write_string)]
+    pub encoding: String,
+
+    #[serde(skip)]
+    pub data_len: u32,
```
james-rms, line 205: why keep `data_len` around when it's observable from the length of the slice itself?

mrkline: The original idea was to have the header fields match the file spec so that `binrw` could handle all the deserialization. But we need to do some length finagling to properly handle padding anyways. I'll remove it.

rust/src/records.rs
mrkline: I think my thought process was to use normal references for types you were unlikely to copy out or pass around. (For example, it wouldn't be unusual to pass Messages, Channels, and Schemas (either from thread-to-thread in a reader, or in the writer API), but you're probably not going to manually assemble a Chunk.

But typing that out makes it sound a bit odd. Want these to be Cow for uniformity's sake?

james-rms: It seems sensible to give the user the option here (and for attachments).

mrkline: Done with 4ded9c7, let me know what you think.

rust/README.md
```diff
@@ -1,31 +1,18 @@
-# Rust MCAP reading library
+# mcap-rs
```
james-rms, line 1: Out of date?

rust/tests/round_trip.rs
```diff
@@ -0,0 +1,169 @@
...
+        summary
+            .chunk_indexes
+            .iter()
+            .flat_map(|ci| summary.stream_chunk(&ours, ci).unwrap()),
+    ) {
+        assert_eq!(whole?, by_chunk?);
+    }
+
+    // Hard mode: randomly access every message in the MCAP.
+    // Yes, this is dumb and O(n^2).
```
james-rms, line 96: 7 minute CI isn't a terrible situation - with this test rust is the long pole, taking the crown from C++ (5 mins). I think we can speed this up a little by writing a custom MCAP with only 10 messages as part of this test instead of using `demo.mcap`. Happy to do that as a separate PR.

james-rms (approved):
Re: `serde` as an optional dependency, I think this is impossible without [duplicating a large amount of record struct declaration](https://serde.rs/remote-derive.html). Probably the right path there is to just write a manual transformation from `serde_json::Value` objects to Records.

rust/README.md
mrkline: Yep! Thanks, fixed in ae22506

amacneil:
👏 🥳

Issue #434, opened 2022-06-23
CLI: support attachments


wkalt:
It should be possible to use attachments with the CLI tool:
* Add an attachment to an existing mcap file. This should be done efficiently, by removing the summary section from the end of the file, putting the attachment at the end of the data section, and putting the summary section back with a new attachment index record.
* List attachments in the CLI tool. The listing should contain any info we have about attachments in the summary section.
* Spit an attachment out of the file, by offset. The offset can be obtained by the listing and serves as an identifier.

To make these changes, we should strongly consider making a change to the Attachment type in the mcap library itself.

Currently the Attachment type represents attachment data as a byte array. Instead, it should represent it as a reader. AddAttachment should thus accept an Attachment with a reader inside, and ParseAttachment should potentially emit one (see below).

The complication with ParseAttachment is that when you are parsing records off of a stream, all the current Parse* functions fully materialize the record content in memory. With attachments this is undesirable because they may be large. However, _not_ doing this with attachments will obligate the caller of ParseAttachment to consume the reader in some way before proceeding to the next record - otherwise, the next call to Read would start somewhere in the attachment content. The caller is not obligated to do this for any other record type, so this should be at a minimum messaged clearly - or alternatively a better way should be found.

The best alternative way that has come to my mind has been to either remove or supplement ParseAttachment with a callback-based mechanism, where a callback (onAttachment) can be passed to NewLexer in LexerOptions. The lexer can then apply the user's logic - which is free to materialize the data or avoid materializing it by writing it out to disk, or whatever.

We have pretty wide latitude to change attachment handling in a breaking manner. Nobody is using it and a mechanism that doesn't materialize the data will be much for the better.

wkalt:
Listing is actually supported already: https://github.com/foxglove/mcap/blob/main/go/cli/mcap/cmd/attachments.go#L1

but adding or retrieving attachments is not.

Issue #372, opened 2022-04-15
[mcap cli] Add metadata to an existing mcap file


jhurliman:
The `ros2 bag record -s mcap` application does not yet support writing metadata. For now, it would be useful to append metadata to an existing `.mcap` file by rewriting it with the new record(s) inserted.

PR #614, opened 2022-09-27
Add support for adding metadata/attachments

wkalt:
Add support for adding metadata/attachments

Adds support for adding and getting metadata and attachments from an
existing mcap file. To add an attachment to a file,

    mcap add attachment demo.mcap -f Makefile

To get an attachment out of the file,

    mcap get attachment demo.mcap -n Makefile -o attachment.txt

To add metadata to a file,

     mcap add metadata demo.mcap -k foo=bar -k bar=baz -n "my metadata"

To get metadata out of the file,

     mcap get metadata demo.mcap -n "my metadata"
     {
       "bar": "baz",
       "foo": "bar"
     }

Also fixes a bug in the list attachments subcommand. Previously this was
listing chunks.

Fixes: https://github.com/foxglove/mcap/issues/434
Fixes: https://github.com/foxglove/mcap/issues/372

### DIFF
--- a/go/cli/mcap/cmd/metadata.go
+++ b/go/cli/mcap/cmd/metadata.go
@@ -1,17 +1,26 @@
 package cmd
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"math"
 	"os"
+	"strings"
 
 	"github.com/foxglove/mcap/go/cli/mcap/utils"
 	"github.com/foxglove/mcap/go/mcap"
 	"github.com/spf13/cobra"
+)
+
+var (
+	addMetadataKeyValues []string
+	addMetadataName      string
+)
+
+var (
+	getMetadataName string
 )
 
 func printMetadata(w io.Writer, r io.ReadSeeker, info *mcap.Info) error {
@@ -45,17 +54,15 @@ func printMetadata(w io.Writer, r io.ReadSeeker, info *mcap.Info) error {
 		if err != nil {
 			return fmt.Errorf("failed to marshal metadata to JSON: %w", err)
 		}
-
-		indented := &bytes.Buffer{}
-		err = json.Indent(indented, jsonSerialized, "", "  ")
+		prettyJSON, err := utils.PrettyJSON(jsonSerialized)
 		if err != nil {
-			return fmt.Errorf("failed to indent JSON: %w", err)
+			return fmt.Errorf("failed to pretty JSON: %w", err)
 		}
 		rows = append(rows, []string{
 			idx.Name,
 			fmt.Sprintf("%d", idx.Offset),
 			fmt.Sprintf("%d", idx.Length),
-			indented.String(),
+			prettyJSON,
 		})
 	}
 	utils.FormatTable(w, rows)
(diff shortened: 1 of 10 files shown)


wkalt:
todo:
* spellcheck
* ensure required flags are marked required
* docstring on RewriteMCAP explaining future plans

go/cli/mcap/cmd/metadata.go
```diff
@@ -88,6 +98,123 @@ var listMetadataCmd = &cobra.Command{
...
+
 func init() {
 	listCmd.AddCommand(listMetadataCmd)
+	addCmd.AddCommand(addMetadataCmd)
+	getCmd.AddCommand(getMetadataCmd)
+
+	addMetadataCmd.PersistentFlags().StringVarP(&addMetadataName, "name", "n", "", "name of metadata record to add")
+	addMetadataCmd.PersistentFlags().StringSliceVarP(&addMetadataKeyValues, "key", "k", []string{}, "key=value pair")
+
+	getMetadataCmd.PersistentFlags().StringVarP(&getMetadataName, "name", "n", "", "name of metadata to aggregate")
```
wkalt, line 219: the ordering of AddCommand vs PersistentFlags() is different here than in the attachments file above. We should standardize on the pattern in the other file.

wkalt: also, the word aggregate should probably be avoided or expanded on

go/cli/mcap/cmd/metadata.go
```diff
@@ -88,6 +98,123 @@ var listMetadataCmd = &cobra.Command{
...
+			}
+			buf := &bytes.Buffer{}
+			err = json.Indent(buf, jsonBytes, "", "    ")
+			if err != nil {
+				return fmt.Errorf("failed to format metadata: %w", err)
+			}
+			_, err = buf.Write([]byte{'\n'})
+			if err != nil {
+				return fmt.Errorf("failed to write terminal newline: %w", err)
+			}
```
wkalt, line 198: we can pull this indenting + newline termination into a json pretty print function

go/cli/mcap/cmd/metadata.go
```diff
@@ -88,6 +98,123 @@ var listMetadataCmd = &cobra.Command{
...
+var getMetadataCmd = &cobra.Command{
+	Use:   "metadata",
+	Short: "get metadata by name",
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := context.Background()
+		if len(args) != 1 {
+			die("Unexpected number of args")
+		}
+		filename := args[0]
+		err := utils.WithReader(ctx, filename, func(_ bool, rs io.ReadSeeker) error {
```
wkalt, line 155: in general, the run commands in this patch are too heavy. It would be easier to test if these dispatched to standalone functions.

go/cli/mcap/utils/utils.go
```diff
@@ -111,10 +115,152 @@ func FormatTable(w io.Writer, rows [][]string) {
 	tw.Render()
 }
 
+func inferWriterOptions(info *mcap.Info) *mcap.WriterOptions {
+	// assume if there are no chunk indexes, the file is not chunked
```
jhurliman, line 119: Can this assumption fail on an unindexed file?

jhurliman (approved):

go/cli/mcap/utils/utils.go
wkalt: if the file is chunked but not indexed, yeah it'll fail. I think though the only way to guard against that would be to scan the entire file to see if a chunk is present. I'm not sure that's worth it, but it's something we could do.

This is where it would be handy to have the writer options in the header.

PR #616, opened 2022-09-27
CLI: add API evolution policy

james-rms:
If we like, we can block this PR until #599 merges.

### DIFF
--- a/go/cli/mcap/README.md
+++ b/go/cli/mcap/README.md
@@ -177,4 +177,4 @@ Here "option" refers to either a subcommand (such as `mcap convert`) or a comman
   deprecate the old name. The deprecation warning for the old name should indicate the new name
   for this functionality.
 - If some option that you use is deprecated without a clear migration path, please
-  file a [Github issue](https://github.com/foxglove/mcap/issues) explaining your use-case.
+  file a [GitHub issue](https://github.com/foxglove/mcap/issues) explaining your use case.
--- a/go/cli/mcap/README.md
+++ b/go/cli/mcap/README.md
@@ -163,18 +163,15 @@ To ensure the resulting binary is statically linked, build with `make`:
 
 ### API Evolution Policy
 
-This tool uses [semantic versioning](https://semver.org) to indicate the scope of an update to
-users. In addition, we try to follow these guidelines when updating the functionality of the tool.
+This tool uses [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) to indicate the scope of an update to users. In addition, we try to follow these guidelines when updating the functionality of the tool.
 
 Here "option" refers to either a subcommand (such as `mcap convert`) or a command-line argument
 (such as `mcap --help`).
 
 - When a option needs to be removed, it is first marked as deprecated. Once deprecated,
-  it will remain functional for at least two releases or one month,
-  whichever is longer. Deprecated options will print a warning to stderr when used.
+  it will remain functional for at least one month. Deprecated options will print a warning to stderr when used.
   After the deprecation period, it will be removed in the next major release.
 - When an option needs to be renamed, we duplicate the functionality under the new name and
-  deprecate the old name. The deprecation warning for the old name should indicate the new name
-  for this functionality.
-- If some option that you use is deprecated without a clear migration path, please
+  deprecate the old name. The deprecation warning for the old name should point users to the new name.
+- If an option that you use is deprecated without a clear migration path, please
   file a [GitHub issue](https://github.com/foxglove/mcap/issues) explaining your use case.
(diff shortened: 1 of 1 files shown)


go/cli/mcap/README.md
```diff
@@ -162,3 +160,21 @@ The `mcap list` command can be used with chunks or attachments:
...
+### API Evolution Policy
+
+This tool uses [semantic versioning](https://semver.org) to indicate the scope of an update to
+users. In addition, we try to follow these guidelines when updating the functionality of the tool.
+
+Here "option" refers to either a subcommand (such as `mcap convert`) or a command-line argument
+(such as `mcap --help`).
+
+- When a option needs to be removed, it is first marked as deprecated. Once deprecated,
+  it will remain functional for at least two releases or one month,
```
jhurliman, line 173: Should we simplify this to just "one month"?

go/cli/mcap/README.md
```diff
@@ -162,3 +160,21 @@ The `mcap list` command can be used with chunks or attachments:
 To ensure the resulting binary is statically linked, build with `make`:
 
     make build
+
+### API Evolution Policy
+
+This tool uses [semantic versioning](https://semver.org) to indicate the scope of an update to
```
jtbandes, line 166:
```suggestion
This tool uses [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) to indicate the scope of an update to
```

go/cli/mcap/README.md
```diff
@@ -162,3 +160,21 @@ The `mcap list` command can be used with chunks or attachments:
...
+
+- When a option needs to be removed, it is first marked as deprecated. Once deprecated,
+  it will remain functional for at least two releases or one month,
+  whichever is longer. Deprecated options will print a warning to stderr when used.
+  After the deprecation period, it will be removed in the next major release.
+- When an option needs to be renamed, we duplicate the functionality under the new name and
+  deprecate the old name. The deprecation warning for the old name should indicate the new name
+  for this functionality.
+- If some option that you use is deprecated without a clear migration path, please
+  file a [Github issue](https://github.com/foxglove/mcap/issues) explaining your use-case.
```
jtbandes, line 180:
```suggestion
  file a [GitHub issue](https://github.com/foxglove/mcap/issues) explaining your use case.
```

jtbandes (commented):
Should we make this a repo-level policy? I think ideally it would appear on the mcap.dev docs site and apply to libraries in all languages.

go/cli/mcap/README.md
james-rms: Yeah, seems overcomplicated in retrospect.

james-rms:
> Should we make this a repo-level policy?

I think it's worth getting specific for this case, as well as a more broad statement in the toplevel README about how all of our libraries follow semver.

wkalt (approved):

go/cli/mcap/README.md
```diff
@@ -162,3 +160,18 @@ The `mcap list` command can be used with chunks or attachments:
 To ensure the resulting binary is statically linked, build with `make`:
 
     make build
+
+### API Evolution Policy
+
+This tool uses [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) to indicate the scope of an update to users. In addition, we try to follow these guidelines when updating the functionality of the tool.
```
wkalt, line 166: if we are using semver, I'm not sure why we need the "one month" policy. They can always stay on the old version as long as they want, and depending on where in the month they encounter the feature/bug that causes them to want to upgrade the fact that it's been deprecated a month may not matter or even be known to them. People probably won't upgrade the tool until they are affected by a bug or want a new feature.

I wonder if we could replace the "one month" concept with a statement like, "bugs will be backpatched up to one release back" (if we are comfortable committing to that).

The hypothetical issue I see with this 1 month policy is,
1. User is using v1 with feature X
2. We release v2 that deprecates feature X. User doesn't care/isn't aware; v1 works fine.
3. +31 days We release v3 that introduces valuable feature Y and retires X. User didn't know that was coming, but now needs to upgrade.

In this scenario it's not clear the 30 day wait time bought us anything. User is free to stay on the old version and upgrade on their own schedule either way.

The alternate issue where it's a bug not a feature is,

1. User is using v1 with feature X
2. We release v2 that deprecates feature X. User doesn't care/isn't aware; v1 works fine.
3. +29 days User encounters a bug in v1/v2
4. +31 days We release v3 that introduces valuable feature Y fixes the bug, but drops X. User now needs to upgrade to get the bugfix, but also loses X.

In this case I don't think the month waiting period wasn't helpful, but what would be helpful would be a policy of backpatching bugs so the user doesn't need to take a breaking change to get the issue fixed.

wkalt (changes requested):
retracting approval on consideration. Not sure the one month policy is worth committing to.

Issue #573, opened 2022-09-13
CLI: mcap bag conversion should support bz2-compressed bags


wkalt:
This should be supported.

jhurliman:
Elevating to a bug since we don't inform users that bz2 bags are not supported

PR #621, opened 2022-09-29
Add rosbag bz2 support to `mcap convert`

jhurliman:
**Public-Facing Changes**
- `mcap convert` supports rosbag `.bag` files with BZ2 compression

**Description**
Fixes https://github.com/foxglove/mcap/issues/573

### DIFF
--- a/go/ros/bag2mcap.go
+++ b/go/ros/bag2mcap.go
@@ -2,6 +2,7 @@ package ros
 
 import (
 	"bytes"
+	"compress/bzip2"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


go/ros/bag2mcap.go
```diff
@@ -202,6 +203,12 @@ func processBag(
 				} else {
 					chunkReader.Reset(r)
 				}
+			case "bz2":
+				if chunkReader == nil {
+					chunkReader = bzip2.NewReader(r)
```
wkalt, line 208: I think you will need to use,

```
&resettableByteReader{bzip2.NewReader(r)}
```

here to make it conform to the expected interface. The bzip2 reader in the compress package doesn't actually support Resetting - this will simply fake it.

wkalt (approved):

PR #623, opened 2022-09-29
Print help instructions if get or list are used without args

wkalt:


wkalt:
we could extend this behavior to all commands if called without args, but some of those print descriptive error messages at the moment so I haven't touched them for now.

james-rms (approved):

Issue #437, opened 2022-06-23
CLI: don't include timestamps error output


wkalt:
This is coming from usage of log.Fatalf instead of the die() function defined in cmd/root.go.

All usage of log.Fatal should be replaced with die(). The die function should maybe be renamed to "exit1" or something if the current name seems unclear.

PR #630, opened 2022-10-03
[cli] Replace log.Fatal*() calls with die()

jhurliman:
**Public-Facing Changes**
- [cli] Consistent formatting for stderr messages

**Description**
Fixes https://github.com/foxglove/mcap/issues/437


wkalt (approved):

wkalt:
to fix the test failure, the `%w` format signifier just needs to be changed to `%s` where used in these calls.

PR #632, opened 2022-10-03
Fix hashFiles() input for C++ CI caching

jhurliman:
**Public-Facing Changes**
- None

**Description**
This PR volume mounts `$HOME/.conan/data` to `/root/.conan/data` inside Docker containers specifically for the `ci` Makefile commands, and fixes `hashFiles()` usage in GitHub Actions to ingest conanfile.py inputs rather than the output artifact directory (which does not exist when the hash check is done). The end result is caching of conan build artifacts between CI runs, so the majority of CI time is no longer spent recompiling the protobuf library.

**New CI output:**

```
Packages
    lz4/1.9.3:63c2a85d57849e261f98f935b93ecac31ba71b84 - Cache
    nlohmann_json/3.10.5:5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9 - Cache
    protobuf/3.21.1:2b471b60d03a9d59f138a492b962f734618de80a - Cache
    zlib/1.2.12:63c2a85d57849e261f98f935b93ecac31ba71b84 - Cache
    zstd/1.5.2:54fbb82fadf7dac819625928c5a96d0047e07f77 - Cache
    mcap/0.5.0:5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9 - Editable
Installing (downloading, building) binaries...
lz4/1.9.3: Already installed!
nlohmann_json/3.10.5: Already installed!
zlib/1.2.12: Already installed!
zstd/1.5.2: Already installed!
protobuf/3.21.1: Already installed!
```

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -63,12 +63,12 @@ ci-image: hdoc-build
 
 .PHONY: ci
 ci: ci-image
-	docker run -t --rm -v $(CURDIR):/mcap/cpp mcap_cpp_ci ./build.sh
+	docker run -t --rm -v $(CURDIR):/mcap/cpp -v $(HOME)/.conan/data:/root/.conan/data mcap_cpp_ci ./build.sh
 
 .PHONY: ci-docs
 ci-docs: ci-image
 	docker build -t mcap_cpp_ci_jammy -f ci.Dockerfile .
-	docker run -t --rm -v $(CURDIR):/mcap/cpp -v $(CURDIR)/../__docs__/cpp:/hdoc-output mcap_cpp_ci_jammy ./build-docs.sh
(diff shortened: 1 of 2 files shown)


cpp/Makefile
```diff
@@ -63,12 +63,12 @@ ci-image: hdoc-build
 
 .PHONY: ci
 ci: ci-image
-	docker run -t --rm -v $(CURDIR):/mcap/cpp mcap_cpp_ci ./build.sh
+	docker run -t --rm -v $(CURDIR):/mcap/cpp -v $(HOME)/.conan/data:/root/.conan/data mcap_cpp_ci ./build.sh
```
jtbandes, line 66: Why do we want the user's local install of conan to interact with the version contained inside Docker?

jhurliman: To take advantage of caching. The "user" here is `ci`

jtbandes: I guess this makes sense as a way to allow it to be cached – but would caching the Docker layers also help?

jhurliman: We already cache the Docker layers, and the protobuf library (conan dependency) is rebuilt on virtually every PR which is why CI is so slow.

PR #640, opened 2022-10-07
[Python] mcap-ros2-support

jhurliman:
**Public-Facing Changes**
- Adds the `mcap-ros2-support` Python library for reading MCAP files containing ROS2 ("ros2msg" / "cdr") data without 
a ROS2 environment or any additional dependencies

**Description**
This adds Python-native ros2msg `.msg` parsing (via a vendored parser from rosidl_adapter), Python-native CDR 
deserialization, and a deserializer interface similar to ROS1's `dynamic.generate_dynamic()`. ROS2 MCAP files can be 
iterated with the high-level `read_ros2_messages()` method, and an example is included in the `ros2-noenv` directory. 
Writer support is not implemented in this PR.


python/.flake8
```diff
@@ -1,7 +1,10 @@
 [flake8]
 max-line-length = 100
 exclude =
+  **/build/**,
   examples/protobuf/*pb2.py,
   mcap-protobuf-support/tests/*pb2.py,
   mcap-ros1-support/mcap_ros1/vendor/genpy/*.py,
   mcap-ros1-support/mcap_ros1/vendor/genmsg/*.py,
+  mcap-ros2-support/mcap_ros2/vendor/rosidl_adapter/*.py,
+ignore = E203 W503
```
jtbandes, line 10: add a comment on why this is here? I assume this was the disagreement between black and flake8?

python/examples/ros2-noenv/dummy_writer.py
```diff
@@ -0,0 +1,57 @@
+import sys
```
jtbandes, line 1: Comment or docstring on what this file is for / what it does?

python/examples/ros2-noenv/reader.py
```diff
@@ -0,0 +1,14 @@
+import sys
+
+from mcap_ros2.reader import read_ros2_messages
+
+
+def main():
+    for msg in read_ros2_messages(sys.argv[1]):
```
jtbandes, line 7: hmm, seems like we should've gone with the MessageEvent naming scheme here, and I find the `.ros_msg` property name weird, but I see that this is the same as the ros1 support library

python/mcap-ros2-support/README.md
```diff
@@ -0,0 +1,26 @@
+# Python MCAP ROS2 support
+
+This package provides ROS2 support for the Python MCAP file format reader.
```
jtbandes, line 3: More detail – maybe mention this provides support for ROS2 when _not_ working in a ROS 2 environment? Or does it support both (and if so, do you get any additional features by being in a ROS environment)?

python/mcap-ros2-support/mcap_ros2/dynamic.py
```diff
@@ -0,0 +1,354 @@
...
+                    raise NotImplementedError(
+                        f"Parsing for type {field.type.type} is not implemented"
+                    )
+                value = parser_fn(reader)
+                setattr(msg, field.name, value)
+
+    return msg
+
+
+def generate_dynamic(schema_name: str, schema_text: str) -> Dict[str, DecoderFunction]:
```
jtbandes, line 292: this function name could use more info. are we generating a dynamic? maybe generate_msg_class? generate_decoder?

python/mcap-ros2-support/mcap_ros2/dynamic.py
```diff
@@ -0,0 +1,354 @@
...
+
+def __getstate__(self: Any):
+    """Support for Python pickling."""
+    return [getattr(self, x) for x in self.__slots__]
+
+
+def __setstate__(self: Any, state: Any):
+    """Support for Python pickling."""
+    for x, val in zip(self.__slots__, state):
+        setattr(self, x, val)
```
jtbandes, line 354: when un-pickling, does the dynamically generated class have to be present in the program already? or is the whole class encoded somehow? could be nice to include an example of pickling

jtbandes (approved):
sweet

python/examples/ros2-noenv/dummy_writer.py
```diff
@@ -0,0 +1,57 @@
...
+float64 x
+float64 y
+float64 z
+================================================================================
+MSG: geometry_msgs/Quaternion
+float64 x
+float64 y
+float64 z
+float64 w"""
+PAYLOAD_HEX = (
```
james-rms, line 31: Should we have a TODO here about being able to serialize messages?

python/mcap-ros2-support/mcap_ros2/cdr.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def uint16(self) -> int:
+        self.__align(2)
+        fmt = "<H" if self.little_endian else ">H"
+        value = struct.unpack_from(fmt, self.data, self.offset)[0]
+        self.offset += 2
+        return value
+
+    def int32(self) -> int:
+        self.__align(4)
```
james-rms, line 65: private members/methods only need one underscore before their name, two implies that you want to invoke [name mangling](https://en.wikipedia.org/wiki/Name_mangling), which I don't see a use for here.

python/mcap-ros2-support/mcap_ros2/cdr.py
```diff
@@ -0,0 +1,162 @@
...
+        fmt = "<i" if self.little_endian else ">i"
+        value = struct.unpack_from(fmt, self.data, self.offset)[0]
+        self.offset += 4
+        return value
+
+    def uint32(self) -> int:
```
james-rms, line 71: why not commonize the implementation here?

```python
def uint32(self) -> int:
    return self._unpack("<I" if self.little_endian else ">I", size=4)

def _unpack(self, fmt: str, size: int):
    if size > 1:
        self.__align(size)
    value = struct.unpack_from(fmt, self.data, self.offset)[0]
    self.offset += size
    return value
```

python/mcap-ros2-support/mcap_ros2/cdr.py
jtbandes: why not default to double-underscore for all private members? (by the way, I think we use a lot of `__` prefixing in the core mcap python lib)

python/mcap-ros2-support/mcap_ros2/cdr.py
```diff
@@ -0,0 +1,162 @@
...
+        new_offset = self.offset + relative_offset
+        if new_offset < 4 or new_offset >= len(self.data):
+            raise Exception(
+                f"seek({relative_offset}) failed, {new_offset} is outside the data range"
+            )
+        self.offset = new_offset
+
+    def seek_to(self, offset: int):
+        if offset < 4 or offset >= len(self.data):
+            raise Exception(
```
james-rms, line 154: avoid raising bare Exceptions, RuntimeError is appropriate here
```suggestion
            raise RuntimeError(
```

python/mcap-ros2-support/mcap_ros2/dynamic.py
```diff
@@ -0,0 +1,354 @@
...
+
+def parseInt32(reader: CdrReader) -> int:
+    return reader.int32()
+
+
+def parseUint32(parser: CdrReader) -> int:
+    return parser.uint32()
+
+
+def parseInt64(reader: CdrReader) -> int:
```
james-rms, line 63: why not just use the methods on `reader`?

python/.flake8
jhurliman: More specifically, E203 violates PEP8. Can I put comments in this file? I can't tell what the syntax is

python/mcap-ros2-support/mcap_ros2/dynamic.py
```diff
@@ -0,0 +1,354 @@
...
+def parseWstringArray(reader: CdrReader, array_length: int) -> List[str]:
+    raise NotImplementedError("wstring[] parsing is not implemented")
+
+
+def parseTimeArray(reader: CdrReader, array_length: int) -> List[Time]:
+    return [parseTime(reader) for _ in range(array_length)]
+
+
+FIELD_PARSERS = {
+    "bool": parseBool,
```
james-rms, line 142:
```suggestion
    "bool": CdrReader.bool,
```

and so on. These can be invoked by explicitly passing the `self` argument, ie:

```
r = CdrReader()
print(FIELD_PARSERS["bool"](r)) # "True" or "False"

File: python/mcap-ros2-support/mcap_ros2/cdr.py (hunk 8 above)
Comment (james-rms) [reply python/mcap-ros2-support/mcap_ros2/cdr.py:65]: It wouldn't break anything, it's just not idiomatic - unless we're using name mangling for something, let's not tell the reader that we need it.

Comment (james-rms) [reply python/mcap-ros2-support/mcap_ros2/cdr.py:65]: Also re: other libraries, I think we should clean that up at some point.

File: python/.flake8 (hunk 1 above)
Comment (james-rms) [reply python/.flake8:10]: Why disable them for the whole project?

File: python/mcap-ros2-support/mcap_ros2/dynamic.py [hunk 13]
```diff
@@ -0,0 +1,354 @@
...
+    "int64": parseInt64,
+    "uint64": parseUint64,
+    "string": parseString,
+    "wstring": parseWstring,
+    "duration": parseTime,
+    "time": parseTime,
+}
+
+ARRAY_PARSERS = {
+    "bool": parseBoolArray,
```
james-rms, line 162: could use a functor here?

```python
def array_parser(primitive_parse_func):
    def parser(reader: CdrReader, length: int):
        return [primitive_parse_func(reader) for _ in range(length)]
    return parser

ARRAY_PARSERS = {
    "bool": array_parser(CdrReader.bool),
    ....
}
```

Python type annotations are a PITA for generics like this unfortunately.

python/mcap-ros2-support/mcap_ros2/cdr.py
jtbandes: I guess I never learned to read this as “name mangling is _needed_”. Rather I just read it as “this variable does not (and cannot) need to be accessed anywhere outside the class definition” (much like new JS #private identifiers). Are there style guides that discuss this somewhere? I don’t feel strongly about changing it or not, just want to understand if I should change my thinking.

python/.flake8
jhurliman: > Why disable them for the whole project?

Because E203 violates PEP8. It should not be enabled on any project. It shouldn't exist at all. One of the first things flake8's config documentation shows you is how to disable E203 (https://flake8.pycqa.org/en/latest/user/configuration.html).

The others are disabled across the whole project because they directly conflict with the black formatting we're applying everywhere.

jtbandes: It looks like TOML. # comments probably work

jtbandes: Does flake8 have a mode that tells it to ignore everything that black would take care of? Kind of like eslint-config-prettier, which disables the builtin eslint rules that would conflict or duplicate things prettier will change during formatting.

python/mcap-ros2-support/mcap_ros2/cdr.py
james-rms: ["Generally, double leading underscores should be used only to avoid name conflicts with attributes in classes designed to be subclassed."](https://peps.python.org/pep-0008/#method-names-and-instance-variables)

PR #641, opened 2022-10-08
[Python] mcap-ros2-support writing, fixes for reading

jhurliman:
**Public-Facing Changes**
- [Python] `mcap-ros2-support` library can now write MCAP files in addition to reading

**Description**
Writing support. Also fixed a few reader bugs that were uncovered with more tests.


python/mcap-ros2-support/mcap_ros2/dynamic.py
```diff
@@ -196,28 +290,246 @@ def _read_complex_type(
...
+                else:
+                    # Limit the array to the upper bound length, if present
+                    if (
+                        ftype.is_upper_bound
+                        and ftype.array_size is not None
+                        and len(array) > ftype.array_size
+                    ):
+                        array = array[: ftype.array_size]
+
+                    # Dynamic length array, write a uint32 prefix
```
jtbandes, line 371: Is the encoding of variable length arrays specified as part of CDR or just a convention that ROS 2 uses on top of it? If the former, maybe this could move into the writer class

jtbandes (approved):

PR #642, opened 2022-10-08
Get rid from static linking of the zstd in mcap cpp

MichaelOrlov:
**Public-Facing Changes**
Get rid from static linking of the zstd in mcap cpp by removing #define ZSTD_STATIC_LINKING_ONLY

**Description**
We need to link zstd dynamically to be able to use zstd_vendor in ROS2

### DIFF
--- a/cpp/mcap/include/mcap/reader.inl
+++ b/cpp/mcap/include/mcap/reader.inl
@@ -3,7 +3,6 @@
 #include <cassert>
 #include <lz4frame.h>
 
-#define ZSTD_STATIC_LINKING_ONLY
 #include <zstd.h>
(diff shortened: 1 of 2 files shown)


jhurliman (approved):

MichaelOrlov:
@jhurliman Since this PR has been approved and merged can we bumped mcap version to be able to refer on it in ROS2 `mcap_vendor` package?

jhurliman:
cc @jtbandes who I believe knows the process for publishing new C++ library versions

jtbandes:
Adding release instructions here: https://github.com/foxglove/mcap/pull/655

PR #647, opened 2022-10-10
rust: Provide RawMessageStream of (header, data) tuples

mrkline:
**Public-Facing Changes**
Adds `read::RawMessageStream`, no changes to existing API

**Description**
We've run into use cases where we don't care about the specifics of each message's Channel, but just want a way to uniquely identify it. Hashing the Channel each time would be sad and slow, and MCAP already gives us a unique ID for the channel. Use that!

Happy to nitpick names, or whether we'd rather return some
```rust
struct RawMessage<'a> {
    header: records::MessageHeader,
    data: Cow<'a, [u8]>
}
```
instead of a tuple, etc.
```

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -7,7 +7,7 @@ categories = [ "science::robotics", "compression" ]
 repository = "https://github.com/foxglove/mcap"
 documentation = "https://mcap.dev/docs/rust/doc/mcap/index.html"
 readme = "README.md"
-version = "0.3.0"
+version = "0.3.1"
 edition = "2021"
 license = "MIT"
 
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -7,7 +7,7 @@ categories = [ "science::robotics", "compression" ]
 repository = "https://github.com/foxglove/mcap"
 documentation = "https://mcap.dev/docs/rust/doc/mcap/index.html"
 readme = "README.md"
-version = "0.3.1"
+version = "0.4.0"
 edition = "2021"
 license = "MIT"
 
--- a/rust/src/read.rs
+++ b/rust/src/read.rs
@@ -574,22 +574,21 @@ impl<'a> ChannelAccumulator<'a> {
 /// Reads all messages from the MCAP file---in the order they were written---and
 /// perform needed validation (CRCs, etc.) as we go.
 ///
-/// This stops at the end of the data section and does not read the summary.
+/// Unlike [`MessageStream`], this iterator returns the raw [`MessageHeader`](records::MessageHeader)
+/// and message data instead of constructing a [`Message`].
+/// This can be useful for situations where you don't need the specifics of each
+/// message's [`Channel`], but just want to be able to discriminate them _by_ their channel
+/// (e.g., build some map of `Channel -> Vec<Message>`).
 ///
-/// Because tying the lifetime of each message to the underlying MCAP memory map
-/// makes it very difficult to send between threads or use in async land,
-/// and because we assume _most_ MCAP files have _most_ messages in compressed chunks,
-/// yielded [`Message`](crate::Message)s have unbounded lifetimes.
-/// For messages we've decompressed into their own buffers, this is free!
-/// For uncompressed messages, we take a copy of the message's data.
-pub struct MessageStream<'a> {
+/// This stops at the end of the data section and does not read the summary.
+pub struct RawMessageStream<'a> {
     full_file: &'a [u8],
     records: ChunkFlattener<'a>,
     done: bool,
     channeler: ChannelAccumulator<'static>,
 }
 
-impl<'a> MessageStream<'a> {
+impl<'a> RawMessageStream<'a> {
     pub fn new(buf: &'a [u8]) -> McapResult<Self> {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: rust/Cargo.toml [hunk 1]
```diff
@@ -7,7 +7,7 @@ categories = [ "science::robotics", "compression" ]
 repository = "https://github.com/foxglove/mcap"
 documentation = "https://mcap.dev/docs/rust/doc/mcap/index.html"
 readme = "README.md"
-version = "0.3.0"
+version = "0.3.1"
```
james-rms, line 10:
```suggestion
version = "0.4.0"
```
This isn't a bug fix, may as well bump the minor version.

rust/src/read.rs
```diff
@@ -698,6 +684,63 @@ impl<'a> Iterator for MessageStream<'a> {
...
+
+                Some(Ok(Message {
+                    channel,
+                    sequence: header.sequence,
+                    log_time: header.log_time,
+                    publish_time: header.publish_time,
+                    data: Cow::Owned(data.into_owned()),
+                }))
+            }
+            // Coerce Option<McapResult<(header, data)>> into Option<McapResult<Message>>
```
james-rms, line 737: i feel like this comment should cover the match arm above?

rust/src/read.rs
```diff
@@ -640,25 +644,7 @@ impl<'a> Iterator for MessageStream<'a> {
...
-
-                    let m = Message {
-                        channel,
-                        sequence: header.sequence,
-                        log_time: header.log_time,
-                        publish_time: header.publish_time,
-                        data: Cow::Owned(data.into_owned()),
-                    };
-                    break Some(Ok(m));
+                    break Some(Ok((header, data)));
```
james-rms, line 647: why do we use this `break <result>` pattern instead of just returning the value from this function?

james-rms:
I prefer
```rust
struct RawMessage<'a> {
    header: records::MessageHeader,
    data: Cow<'a, [u8]>
}
```
over the tuple approach. Do we have a use for `MessageHeader` without `data` anywhere? otherwise my highest preference would be to fold `data` into `records::MessageHeader` and rename it `records::Message`. We could leave `read::Message` with the same name or change it to `read::FullMessage` or similar.

rust/src/read.rs
mrkline: I was trying to explain (maybe unsuccessfully :sweat_smile:) why we need these other two arms (which don't seem to do anything) instead of some catch-all
```rust
otherwise => otherwise
```
which complains that the result types don't match:
```
   --> src/read.rs:737:26
    |
716 |           match self.inner.next() {
    |           ----------------------- `match` arms have incompatible types
...
729 | /                 Some(Ok(Message {
730 | |                     channel,
731 | |                     sequence: header.sequence,
732 | |                     log_time: header.log_time,
733 | |                     publish_time: header.publish_time,
734 | |                     data: Cow::Owned(data.into_owned()),
735 | |                 }))
    | |___________________- this is found to be of type `std::option::Option<Result<Message<'static>, McapError>>`
736 |               },
737 |               otherwise => otherwise
    |                            ^^^^^^^^^ expected struct `Message`, found tuple
    |
    = note: expected enum `std::option::Option<Result<Message<'static>, _>>`
               found enum `std::option::Option<Result<(MessageHeader, Cow<'_, [u8]>), _>>````
```
Something like
```
            otherwise => otherwise.map(|r| r.map(|_never_ok| unreachable!())),
```
would work instead, but that seems a lot less clear to me.

Suggestions on a better comment to get that across here?

rust/src/read.rs
mrkline: It's for the bit at the end:
```rust
        if !matches!(n, Some(Ok(_))) {
            self.done = true;
        }
        n
```

If something goes wrong, we want the iterator to return the error, but then return `None` for all subsequent calls to `next()` instead of repeating over and over. Hence the `done` flag and the
```rust
        if self.done {
            return None;
        }
```
check up top. Alternatives seem strictly worse (Store the error and return _it_ over and over instead of `None`? Try to read some more after encountering an unrecoverable problem with the file?)

I'm all ears if you have an idea for a more elegant way to do it.

mrkline:
> Do we have a use for `MessageHeader` without `data` anywhere? otherwise my highest preference would be to fold `data` into `records::MessageHeader` and rename it `records::Message`.

Headers are separate structs so that we can serialize and deserialize them using `binrw` and its `#[derive(BinRead, BinWrite)]` traits. We could maybe find the right set of `#[...]` spells to get `binrw` to:
1. Read the header
2. Read the data length
3. Validate the data length and adjust for padding
4. Read the correct number of data bytes

But I think it would be more complicated and less readable than our current approach of
1. Use binrw to parse headers
2. Use our own code to validate the length and get the data slice

https://github.com/foxglove/mcap/blob/b187669b1cbf97819406b9485e08af2424b18809/rust/src/read.rs#L159-L240

rust/src/read.rs
james-rms: maybe something like `// Cover remaining match arms with trivial implementations`?

james-rms (approved):
LGTM - i'll handle the version numbers when merging.

PR #649, opened 2022-10-11
go: use internal version string

### LINKED ISSUE #591: Homebrew version of mcap records a strange library string
**Description**

After installing mcap with Homebrew, running `mcap filter` on a file followed by `mcap info` on the resulting file shows the following library string:

```
library: mcap go #(devel); mcap go #(devel)
```

The same string is repeated twice with a `;` separator, and the version is reported as "devel" even though this is the official 0.0.19 release distributed via Homebrew.

- Version: 0.0.19
- Platform: MacOS (M1)

james-rms:
**Public-Facing Changes**

None.

**Description**
With context from https://github.com/golang/go/issues/29228, the result of runtime/debug.BuildInfo.Main.Version is not well defined. Here we use an internally-defined Version as our library version in all contexts. We also add a test when using a go library release tag `go/mcap/v1.2.3` that the Version string is correct.

This PR also changes the behaviour of `Writer` to only append the existing library version if it's different from the current version. This removes the awkward behaviour of `mcap filter` where the resulting mcap Library would be `mcap go #(devel); mcap go #(devel); mcap go #(devel)...`.

Fixes #591


go/cli/check-tag/go.mod
```diff
@@ -0,0 +1,14 @@
+module github.com/foxglove/mcap/go/cli/check-tag
+
+go 1.18
+
+require github.com/foxglove/mcap/go/mcap v0.0.0-20221010221948-8b72814b2c50
```
jtbandes, line 5: what's the meaning of this version specifier? will this have to be bumped when we make commits to the mcap library?

.github/workflows/ci.yml
```diff
@@ -323,6 +323,12 @@ jobs:
         run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.49.0
       - run: make lint
       - run: make test
+      - name: Check library version
+        if: |
+          !github.event.pull_request.head.repo.fork &&
+          startsWith(github.ref, 'refs/tags/go/mcap/v')
+        working-directory: go/cli/check-tag
+        run: make build && bin/check-tag
```
jtbandes, line 331: Should we put this in the `go-release-cli` step instead? I see that go-release-cli has `needs: [go]`, however it still does its own build and as far as I can see doesn't actually depend on the output of the `go` job. So it seems a little nicer for this validation to be in the job that needs it (go-release-cli).

go/cli/check-tag/main.go
```diff
@@ -0,0 +1,38 @@
+/// Utility for checking that the current git tag matches the internal MCAP library version string.
```
jtbandes, line 1: I generally hate bash scripts, but it kinda feels like this whole program could be a few lines of bash. Is there some technical reason to make a separate go binary for this rather than just calling `mcap version --library` to get the library version and checking it against the git tag?

go/cli/mcap/cmd/version.go
```diff
@@ -3,22 +3,27 @@ package cmd
...
+	Use:   "version",
+	Short: "Output version information",
+	Run: func(cmd *cobra.Command, args []string) {
+		if printLibraryVersion {
+			fmt.Println(mcap.Version)
+		} else {
 			fmt.Println(Version)
-		},
-	}
+		}
```
jtbandes, line 22: Why do we want the library and CLI versions to be separate concepts at all?

.github/workflows/ci.yml
jtbandes: I now see that this is not relevant to the cli because it's for `go/mcap/vX` rather than `go/mcap-cli/vX`.

go/cli/mcap/cmd/version.go
james-rms: They seem separate to me 😅 A breaking change in the mcap library vs. mcap CLI API are pretty different things. To the extent that we have any users of the mcap library directly, they shouldn't care about what happens with the CLI when updating versions.

jtbandes: So to summarize the changes, the library version is in the code, and the cli version is still coming from the version variable at build time, right?

james-rms: correct.

go/check_tag.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+expected_tag="go/mcap/$($1 version --library)"
+read -ra all_tags <<< "$(git tag --points-at HEAD)"
+found="false"
+for tag in "${all_tags[@]}"; do
+    if [ "$tag" = "$expected_tag" ]; then
+        found="true"
+    fi
+done
```
jtbandes, line 18: Might be slightly simpler if you used `git rev-parse go/mcap/$(...)` and compared with `git rev-parse HEAD`.

jtbandes (approved):
So when this CI step fails, how will we notice? Just a red X on the latest commit on main? (It's not blocking any PRs right?)

james-rms:
Yes, that's about the best we can do I think.

PR #650, opened 2022-10-11
go: do not calculate CRC if not required

### LINKED ISSUE #638: Go mcap writer calculates crc even when IncludeCRC is false
CRC should not be calculated if IncludeCRC is false.

https://github.com/foxglove/mcap/blob/14fa18afb83e504ae03397fd4e9292a6c83a7e80/go/mcap/writer.go#L738
https://github.com/foxglove/mcap/blob/14fa18afb83e504ae03397fd4e9292a6c83a7e80/go/mcap/write_sizer.go#L20

james-rms:
**Public-Facing Changes**

**Description**

Fixes #638


go/mcap/write_sizer.go
```diff
@@ -15,9 +14,9 @@ func (w *writeSizer) Write(p []byte) (int, error) {
 	return w.w.Write(p)
 }
 
-func newWriteSizer(w io.Writer) *writeSizer {
+func newWriteSizer(w io.Writer, calculateCRC bool) *writeSizer {
 	return &writeSizer{
-		w: newCRCWriter(w),
+		w: newCRCWriter(w, calculateCRC),
```
alokpr, line 19: It seems a bit weird to pass a boolean to calculateCRC to crc-writer. How about this?

```
if (calculateCRC) {
  w = newCRCWriter(w)
}
return &writeSizer{w: w}
```

Issue #633, opened 2022-10-04
`test demo_round_trip has been running for over 60 seconds`


jhurliman:
The rust task in CI takes 5x longer than all other tasks, and it seems to be slow execution time of one or more tests. Example run:

https://github.com/foxglove/mcap/actions/runs/3178572685/jobs/5180207785

Relevant output:

```
test demo_round_trip has been running for over 60 seconds
```

PR #653, opened 2022-10-12
Make zstd optional in Rust library

### LINKED ISSUE #648: rust: make zstd dependency optional
Nice to be able to compile without zstd for users that don't want to use it.

foxymiles:
**Public-Facing Changes**
Makes zstd an optional dependency for the Rust library.

**Description**
This makes it possible to build the Rust library without built-in zstd compression support. The default build will still include it.

Note this also required adding an uncompressed version of the demo.mcap test data because when the Rust code is compiled without zstd it is unable to read the existing demo.mcap for test purposes.

Fixes #648

### DIFF
--- a/rust/tests/common.rs
+++ b/rust/tests/common.rs
@@ -13,8 +13,8 @@ pub fn map_mcap<P: AsRef<Utf8Path>>(p: P) -> Result<Mmap> {
 #[allow(dead_code)]
 pub fn mcap_test_file() -> Result<Mmap> {
     if cfg!(feature = "zstd") {
-        return map_mcap("tests/data/compressed.mcap");
+        map_mcap("tests/data/compressed.mcap")
     } else {
-        return map_mcap("tests/data/uncompressed.mcap");
+        map_mcap("tests/data/uncompressed.mcap")
     }
 }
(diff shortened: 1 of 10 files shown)


jtbandes (approved):

foxymiles:
Note - this doesn't seem to reduce the binary size by very much.

mrkline:
Would we be interested in doing a similar feature flag for lz4 in a subsequent PR?

james-rms (changes requested):
I think this is a good idea and the implementation is good, but I don't think this should merge until #633 is addressed.

foxymiles:
@mrkline It seems like it would be similarly straightforward to also make lz4 optional, right?

mrkline:
@foxymiles yeah, absolutely!

foxymiles:
@james-rms Ok to merge now?

rust/tests/common.rs
```diff
@@ -9,3 +9,12 @@ pub fn map_mcap<P: AsRef<Utf8Path>>(p: P) -> Result<Mmap> {
     let fd = fs::File::open(p).with_context(|| format!("Couldn't open {p}"))?;
     unsafe { Mmap::map(&fd) }.with_context(|| format!("Couldn't map {p}"))
 }
+
+#[allow(dead_code)]
+pub fn mcap_test_file() -> Result<Mmap> {
+    if cfg!(feature = "zstd") {
+        return map_mcap("tests/data/compressed.mcap");
```
mrkline, line 16: Nit: Rust automatically returns the last expression in a scope (a la Ruby, Haskell, etc.), so this could just be `map_mcap("tests/data/compressed.mcap")`

foxymiles: Thanks good catch. I'm still getting comfortable with Rust. (but enjoying it quite a bit)

james-rms (approved):

PR #664, opened 2022-10-14
Add rust writer conformance test runner

### LINKED ISSUE #657: rust: implement writer conformance tests
Right now we don't cover rust in writer conformance tests, but we could.

foxymiles:
**Public-Facing Changes**
Adds writer conformance test runner for rust.

**Description**
Adds writer conformance test runner for rust. 

Note that at present none of the tests actually pass since the rust writer doesn't currently expose some lower level details required to pass the tests, like specifying the `library` or explicit channel numbers. I expect these issues can be addressed in subsequent PRs.
Fixes #657

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,8 +206,7 @@ jobs:
       - run: yarn workspace @mcap/core test
 
       - name: Publish to NPM
-        if:
-          ${{ startsWith(github.ref, 'refs/tags/releases/typescript/core/v') }}
+        if: ${{ startsWith(github.ref, 'refs/tags/releases/typescript/core/v') }}
         run: yarn workspace @mcap/core publish --access public
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
@@ -250,9 +249,7 @@ jobs:
       - run: make build
       - name: Publish mcap to TestPyPI
         uses: pypa/gh-action-pypi-publish@release/v1
-        if:
-          ${{ !github.event.pull_request.head.repo.fork && github.actor !=
-          'dependabot[bot]' }}
+        if: ${{ !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]' }}
         with:
           user: __token__
           password: ${{ secrets.TESTPYPI_API_TOKEN }}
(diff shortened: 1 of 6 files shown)


rust/examples/conformance_writer.rs
```diff
@@ -0,0 +1,129 @@
...
+                    data: Cow::from(record.get_field_data("data")),
+                    log_time: record.get_field_u64("log_time"),
+                    publish_time: record.get_field_u64("publish_time"),
+                    sequence: record.get_field_u32("sequence"),
+                };
+                writer.write(&message).expect("Write message failed");
+            }
+            "Metadata" => {
+                // write metadata
+            }
```
jtbandes, line 94: missing implementation?

tests/conformance/scripts/run-tests/runners/RustWriterTestRunner.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  supportsVariant(variant: TestVariant): boolean {
+    return (
+      intersection(
+        [...variant.features],
+        ["ch", "mx", "st", "rsh", "rch", "ax", "mdx", "chx", "sum", "pad"],
+      ).length === 0
+    );
+  }
```
jtbandes, line 34: I don't see the point of doing this now, because you aren't running this test in CI anyway. Maybe just return true for now?

In general we should use the actual enum to make this more readable, e.g. some of the other runners use things like `variant.features.has(TestFeatures.AddExtraDataToRecords)`.

Also, if you do want to keep this behavior, it could've been written as `return variant.features.size === 0`.

rust/examples/conformance_writer.rs
foxymiles: This is in now.

jtbandes: no it's not...forgot to push? :)

.github/workflows/ci.yml
```diff
@@ -157,7 +157,7 @@ jobs:
       - run: cd rust && cargo build --example=conformance_reader
       - run: yarn install --frozen-lockfile
       - run: yarn test:conformance:generate-inputs --verify
-      - run: yarn test:conformance --runner rust-
+      - run: yarn test:conformance --runner rust-streamed-reader
```
jtbandes, line 160: can this change be removed now so it runs in CI?

tests/conformance/scripts/run-tests/runners/RustWriterTestRunner.ts
```diff
@@ -0,0 +1,29 @@
...
+      if (errText.length > 0) {
+        console.error(errText);
+      }
+    }
+    return stdout as unknown as Uint8Array;
+  }
+
+  supportsVariant(_variant: TestVariant): boolean {
+    return false;
+  }
```
jtbandes, line 28: 😢

foxymiles: Temporary sadness. Will continue on this in a followup.

jtbandes (approved):

james-rms (approved):

PR #666, opened 2022-10-14
CLI: Omit metadata and attachment counts if stats are not present

wkalt:


jtbandes (approved):

wkalt:
After this change,
```
$ mcap info tests/conformance/data/OneAttachment/OneAttachment.mcap
library: 
profile: 
channels:
```

We should probably apply similar logic to "channels", but that will require changing the library to return a pointer instead of a map, which will require a library release first.

jtbandes:
We could also consider keeping the output lines but saying `attachments: unknown` explicitly instead of just omitting the line.

wkalt:
yeah, I like that idea

wkalt:
Updated output:
```
$ mcap info tests/conformance/data/OneAttachment/OneAttachment.mcap
library: 
profile: 
channels:
attachments: unknown
metadata: unknown
```

jtbandes (approved):

james-rms (approved):

PR #667, opened 2022-10-16
rust: Don't force serde on users for our tests

### LINKED ISSUE #656: rust: serde dependency is heavy
**Description**

the `mcap` lib depends on `serde`, which is known to crank up compile time / binary size, but it only needs it for conformance tests.

TBD how important this is.

- Version:
- Platform:

**Steps To Reproduce**

**Expected Behavior**

**Actual Behavior**

mrkline:
**Public-Facing Changes**

Removes `serde` dependency for Rust crate

**Description**

We don't want to force Serde on users just for the sake of the conformance tests. (In what context would you want to serialize individual records of a MCAP to something else?)

Stamp out and stringify them ourselves.

fixes #656


james-rms (approved):

james-rms:
thanks!

PR #671, opened 2022-10-17
Fix rust tests by adding serde as a dev dependency

foxymiles:
**Public-Facing Changes**
None

**Description**
Fix rust tests by adding back serde as a dev dependency.


james-rms (approved):
why not `version = "1"`?

foxymiles:
TBH I'm a bit of a newbie with Cargo versioning but it looks like "1.0.145" means "greater than 1.0.145 and less than 2.0.0" whereas "1" would mean "greater than 1.0.0 and less than 2.0.0" right?

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

PR #687, opened 2022-10-28
(POC) optimize subcommand for preloading

wkalt:
This PR is for POC/discussion only.

Implements an "optimize" subcommand that accepts an mcap file and a list
of topics, aggregates the chosen topics from the input mcap file into a
second, in-memory mcap file, and attaches the second file to the end of
the first.

The idea of this is to enable efficient preloading of small topics for
display, regardless of the original size of the mcap. To exhibit the
functionality, select a large mcap file. For example, I have
cal_loop.mcap which is converted from cal_loop.bag. The file is 1.5 GB.

To simulate the current best performance for mcap preloading, run this
command with a selection of small topics:

    $ time mcap cat cal_loop.mcap --topics /diagnostics > /dev/null

    real    0m2.382s
    user    0m2.769s
    sys     0m0.408s

To "optimize" the file, run this command:

    $ time mcap optimize cal_loop.mcap -t /diagnostics > cal_loop.optimized.mcap

Now to simulate the optimized preloading performance, run this:

    $ time mcap get attachment cal_loop.optimized.mcap --name preload_topics | mcap cat > /dev/null

    real    0m0.010s
    user    0m0.017s
    sys     0m0.003s

Performance is 238x improved. Change the cat redirect to `--json` to
play the messages to stdout. Multiple topics may be supplied with
multiple -t flags.

### DIFF
--- a/.lfsconfig
+++ b/.lfsconfig
@@ -1,6 +1,0 @@
-[lfs]
-    # Specify the server to use when pulling Git LFS assets.
-    # This is here to work around a bug in the Swift Package Manager which prevents
-    # it from installing packages when the source repository uses Git LFS.
-    # https://forums.swift.org/t/swiftpm-with-git-lfs/42396/4
-    url = https://github.com/foxglove/mcap.git/info/lfs
--- a/go/cli/mcap/cmd/optimize.go
+++ b/go/cli/mcap/cmd/optimize.go
@@ -2,11 +2,11 @@ package cmd
 
 import (
 	"bytes"
-	"errors"
-	"fmt"
-	"io"
+	"math"
 	"os"
+	"regexp"
 
+	"github.com/foxglove/mcap/go/cli/mcap/utils"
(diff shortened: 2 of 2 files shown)


wkalt:
Here is the timing for accessing the same file remotely in GCS:

```
time mcap get attachment gs://foxglove-wyatt-scratch/cal_loop.optimized.mcap --name preload_topics | mcap cat > /dev/null

real    0m1.121s
user    0m0.019s
sys     0m0.028s
```

By comparison, getting using the cat command directly on that remote file takes minutes to run and hogs all my network bandwidth.

Also, @james-rms correctly noted this is just the filter command. I could have avoided writing code and just demoed this by running mcap filter on the input and attaching the output to the original file. Same thing.

wkalt:
we discussed this in comparison with https://github.com/foxglove/mcap/pull/747, and decided not to merge either one into the official mcap tooling.

PR #691, opened 2022-10-31
[go] Change zstd from fastest to default

### LINKED ISSUE #645: [CLI] zstd-compressed files are larger than what the C++ SDK produces
If you record a MCAP file in ROS 2 with default settings, then make any modifications to that recording with the mcap cli, the file size increases due to lower chunk compression ratios.

jhurliman:
**Public-Facing Changes**
- [cli] Improved compression ratio when creating zstd-compressed mcap files
- [go] Added `WriterOptions.CompressionLevel` to control the zstd or lz4 compression level

**Description**
Fixes #645


wkalt:
@jhurliman the CLI tool is an offline tool but the writer library is not necessarily. I think for that reason we are going to want to bundle this with a change that makes this a config option that can be passed into the writer so that nobody is forced to take a performance hit. At that point we may as well also pass it in as a flag in the CLI (potentially even retaining the old default in the writer).

jhurliman:
As it currently stands, our real-time rosbag2 recorder that we publish benchmarks on produces smaller files than the `mcap` cli tool.

jhurliman:
The `bag2mcap.ts` example now uses zstd compression and produces smaller files than our CLI tool as well.

defunctzombie:
This change seems good to me. @wkalt can you explain the reasoning why we would not want to use `default` in the CLI tool which is an offline tool?

wkalt (approved):

PR #692, opened 2022-11-01
CLI: merge preserves profile from inputs

### LINKED ISSUE #684: [CLI] mcap merge does not preserve profile
mcap merge command always sets profile to “”, even when both input files have a matching non-empty profile.

james-rms:
**Public-Facing Changes**
* Removes the `--profile` argument from `mcap merge`.
* converts `mcap.Reader.readHeader` into a public method `Header()`.  This works on seeking readers always, and non-seeking readers if no other records have been read from the file.

**Description**
Reads the profile from each input before merging messages, if they're all the same that's the profile for the new MCAP.

Fixes #684

### DIFF
--- a/go/mcap/reader.go
+++ b/go/mcap/reader.go
@@ -58,6 +58,7 @@ type Reader struct {
 	l        *Lexer
 	r        io.Reader
 	rs       io.ReadSeeker
+	header   *Header
 	channels map[uint16]*Channel
 }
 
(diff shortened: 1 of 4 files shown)


jhurliman (approved):

go/mcap/reader.go
```diff
@@ -141,29 +141,33 @@ func (r *Reader) Messages(
...
+// NOTE: on non-seeking readers, this method will return an error if any other records
+// have been read from the MCAP file before.
+func (r *Reader) Header() (*Header, error) {
+	var err error
+	if r.rs != nil {
+		_, err = r.rs.Seek(0, io.SeekStart)
+		if err != nil {
+			return nil, fmt.Errorf("failed to seek to start: %w", err)
+		}
+		r.l, err = NewLexer(r.r, &LexerOptions{EmitChunks: true})
```
wkalt, line 154: is it necessary to make a new lexer here, vs seeking the reader that underlies the one that is attached to the reader already?

go/mcap/reader.go
```diff
@@ -141,29 +141,33 @@ func (r *Reader) Messages(
 	return r.unindexedIterator(ro.Topics, uint64(ro.Start), uint64(ro.End)), nil
 }
 
-func (r *Reader) readHeader() (*Header, error) {
-	_, err := r.rs.Seek(8, io.SeekStart)
-	if err != nil {
-		return nil, fmt.Errorf("failed to seek to header: %w", err)
+// Header reads the Header record from the MCAP.
+// NOTE: on non-seeking readers, this method will return an error if any other records
+// have been read from the MCAP file before.
```
wkalt, line 146: for a seeking reader, won't it also cause issues if other records have been read (i.e if you are in the middle of reading some stuff and call this?). Seems like you would lose your place in the file. Maybe worth a note in the docstring.

wkalt: we could also potentially scan forward to the Header in the constructor of the reader (it already reads the magic via the lexer - header should be right after). Then we could attach it to a Header field on the reader struct so you would be able to refer to it at any point during your reading. That might simplify some of the code here and reduce the amount of seeking around.

wkalt: it should also reduce the latency of Info() when calling against a remotely-hosted file. Currently the way that will work is

1) Open a reader
2) Read forward through the magic (in lexer constructor)
3) Issue a range request to exactly where currently located (offset=8)
4) Read the header
5) Open range request back to the summary section of the file and read that.

If the constructor for the reader read both the magic and the header (magic in the lexer, header in the reader constructor) then we would cut out step 3 and save a roundtrip.

wkalt (approved):

PR #694, opened 2022-11-01
Ensure that fd set only includes one of every message type

pickledgator:
**Description**
* Updates the cpp protobuf example to correctly build the file descriptor set of the parent protobuf type
* Ensures that the file descriptor set doesn't include multiple copies of the same types in the same namespace

### DIFF
--- a/cpp/examples/protobuf/writer.cpp
+++ b/cpp/examples/protobuf/writer.cpp
@@ -53,13 +53,13 @@ std::string SerializeFdSet(const google::protobuf::Descriptor* toplevelDescripto
   while (!toAdd.empty()) {
     const google::protobuf::FileDescriptor* next = toAdd.front();
     toAdd.pop();
-    next->CopyTo(fdSet.add_file());
+    if (added.find(next->name()) == added.end()) {
+      next->CopyTo(fdSet.add_file());
+    }
     added.insert(next->name());
     for (int i = 0; i < next->dependency_count(); ++i) {
       const auto& dep = next->dependency(i);
-      if (added.find(dep->name()) == added.end()) {
-        toAdd.push(dep);
-      }
+      toAdd.push(dep);
     }
   }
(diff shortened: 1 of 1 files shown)


james-rms:
I'm not seeing the bug in the original code - are you able to provide an example?

pickledgator:
The example that failed for me before this change was generating the foxglove.SceneEntity file descriptor set.

cpp/examples/protobuf/writer.cpp
```diff
@@ -53,13 +53,13 @@ std::string SerializeFdSet(const google::protobuf::Descriptor* toplevelDescripto
   while (!toAdd.empty()) {
     const google::protobuf::FileDescriptor* next = toAdd.front();
     toAdd.pop();
-    next->CopyTo(fdSet.add_file());
+    if (added.find(next->name()) == added.end()) {
+      next->CopyTo(fdSet.add_file());
```
jtbandes, line 57: Should we have something like `else { continue }` after this, or restructure it as `if (find != end) { continue }`?

Or what about instead of moving the CopyTo inside a conditional, would it work to move `added.insert(next->name())` inside the for loop?

jtbandes (commented):
Thanks for the contribution!

@james-rms the issue was that duplicate items could end up in the `toAdd` queue, and `CopyTo` would be run before checking whether the items had already been added to the output fdSet. Compare with the Python implementation at https://github.com/foxglove/mcap/blob/21b737f99800bfe1d733b769022a726ab3ca4e0e/python/mcap-protobuf-support/mcap_protobuf/schema.py#L24 which does the `seen_dependencies.add` (equivalent of C++ `added.insert`) inside the `dep.name not in seen_dependencies` conditional.

jtbandes:
I've made another PR to replace this one and add a test: https://github.com/foxglove/mcap/pull/717

PR #696, opened 2022-11-02
cpp: update to v0.6.0

james-rms:


jtbandes:
FYI you will also have to publish to conan-center-index, as described here: https://github.com/foxglove/mcap/tree/main/cpp#releasing-new-versions

PR #700, opened 2022-11-03
CLI: doctor: warn on schema with empty data and encoding fields

### LINKED ISSUE #698: mcap doctor: raise warning on empty schema content
in MCAPs recorded by the `rosbag2_storage_mcap` plugin, an empty schema encoding string indicates that the storage plugin couldn't find the schema. It seems like it would make sense to warn on this in `doctor`. however, according to the spec:
https://mcap.dev/specification/index.html#schema-op0x03
> Format for the schema. The [well-known schema encodings](https://mcap.dev/specification/appendix.html#well-known-schema-encodings) are preferred. An empty string indicates no schema is available.

james-rms:
**Public-Facing Changes**
Adds a warning to `mcap doctor` when we encounter a schema record with no `data` and no `encoding`.
**Description**
An empty schema is a pretty odd corner case - the data it refers to is not schemaless, but there is no information that can be used to decode it. The only example I can think of for this is where a user has some JSON messages and there is no `jsonschema` to describe them, but still wants to provide some label to help the decoder figure out what to do with it.

In every other case, this is something the user ought to know about.

fixes #698

### DIFF
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -261,8 +261,12 @@ func (doctor *mcapDoctor) Examine() error {
 				doctor.error("Failed to parse schema:", err)
 			}
 
-			if schema.Encoding == "" && len(schema.Data) > 0 {
-				doctor.error("Schema.data field should not be set when Schema.encoding is empty")
+			if schema.Encoding == "" {
+				if len(schema.Data) == 0 {
+					doctor.warn("Schema with ID: %d, Name: %d has empty Encoding and Data fields", schema.ID, schema.Name)
+				} else {
+					doctor.error("Schema with ID: %d has empty Encoding but Data contains: %s", schema.ID, string(schema.Data))
+				}
 			}
 
 			if schema.ID == 0 {


go/cli/mcap/cmd/doctor.go
```diff
@@ -261,8 +261,12 @@ func (doctor *mcapDoctor) Examine() error {
 				doctor.error("Failed to parse schema:", err)
 			}
 
-			if schema.Encoding == "" && len(schema.Data) > 0 {
-				doctor.error("Schema.data field should not be set when Schema.encoding is empty")
+			if schema.Encoding == "" {
+				if len(schema.Data) == 0 {
+					doctor.warn("Schema record with ID: %d name: %d is empty", schema.ID, schema.Name)
+				} else {
+					doctor.error("Schema.Encoding is empty but Schema.Data contains: %s", string(schema.Data))
```
defunctzombie, line 268: Probably helpful to include the ID or name here? I'll let you decide.

go/cli/mcap/cmd/doctor.go
```diff
@@ -261,8 +261,12 @@ func (doctor *mcapDoctor) Examine() error {
 				doctor.error("Failed to parse schema:", err)
 			}
 
-			if schema.Encoding == "" && len(schema.Data) > 0 {
-				doctor.error("Schema.data field should not be set when Schema.encoding is empty")
+			if schema.Encoding == "" {
+				if len(schema.Data) == 0 {
+					doctor.warn("Schema record with ID: %d name: %d is empty", schema.ID, schema.Name)
```
defunctzombie, line 266: "is empty" might be ambiguous. Probably better to be explicit that the `data field is empty`

defunctzombie (approved):

PR #702, opened 2022-11-03
lexer: expose informative error on truncation

### LINKED ISSUE #697: `mcap doctor` should say something helpful on a truncated MCAP
Currently when doctoring a truncated MCAP, all we see is:

```
j@192-168-1-108 recovery % mcap doctor ~/Downloads/bag_2022_11_01-18_32_55/bag_2022_11_01-18_32_55_0.mcap
Examining /Users/j/Downloads/bag_2022_11_01-18_32_55/bag_2022_11_01-18_32_55_0.mcap
Failed to read token: unexpected EOF
```

james-rms:
**Public-Facing Changes**
* Improves error message from `mcap recover`  and `mcap doctor` on truncated MCAP files.

Before:
```
Failed to read token: Unexpected EOF
```
After:
```
Failed to read next token: MCAP truncated in message index (0x7) record content with expected length 51542, data ended after 50343 bytes
```

* Library users inspecting errors by equality will need to update their error handling of the `mcap.Lexer` next method. Previously they would expect an `io.UnexpectedEOF`, but now they get `mcap.ErrTruncatedRecord` when an MCAP is truncated. Note that `mcap.ErrTruncatedRecord` wraps `io.UnexpectedEOF`, so if they use `errors.Is` they don't have to modify their code.

fixes #697

### DIFF
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -124,7 +124,7 @@ func (doctor *mcapDoctor) examineChunk(chunk *mcap.Chunk) {
 			if errors.Is(err, io.EOF) {
 				break
 			}
-			die("Failed to read token: %s", err)
+			doctor.fatalf("Failed to read next token: %s", err)
 		}
 		if len(data) > len(msg) {
 			msg = data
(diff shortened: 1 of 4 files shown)


go/cli/mcap/cmd/doctor.go
```diff
@@ -140,8 +140,12 @@ func (doctor *mcapDoctor) examineChunk(chunk *mcap.Chunk) {
 				doctor.error("Failed to parse schema:", err)
 			}
 
-			if schema.Encoding == "" && len(schema.Data) > 0 {
-				doctor.error("Schema.data field should not be set when Schema.encoding is empty")
+			if schema.Encoding == "" {
```
james-rms, line 143: this is a missed fix from #700 - hope reviewers don't mind me merging in this PR.

jtbandes (approved):

PR #711, opened 2022-11-05
Windows dllexport

emersonknapp:
**Public-Facing Changes**
`dllexport`s all interfaces from `reader.hpp`, `writer.hpp`, and `types.hpp` for usage as a shared library on Windows.

**Description**
Add `visibility.hpp` to provide platform-dependent macros to allow for `dllexport`/`dllimport` and skipping on non-Windows.

Part of #716 
Part of https://github.com/ros-tooling/rosbag2_storage_mcap/issues/67

### DIFF
--- a/cpp/mcap/include/mcap/reader.hpp
+++ b/cpp/mcap/include/mcap/reader.hpp
@@ -3,6 +3,7 @@
 #include "intervaltree.hpp"
 #include "read_job_queue.hpp"
 #include "types.hpp"
+#include "visibility.hpp"
 #include <cstdio>
 #include <fstream>
 #include <map>
@@ -71,7 +72,7 @@ struct IReadable {
  * @brief IReadable implementation wrapping a FILE* pointer created by fopen()
  * and a read buffer.
  */
-class FileReader final : public IReadable {
+class MCAP_EXPORT FileReader final : public IReadable {
 public:
   FileReader(std::FILE* file);
 
@@ -88,7 +89,7 @@ private:
 /**
  * @brief IReadable implementation wrapping a std::ifstream input file stream.
  */
-class FileStreamReader final : public IReadable {
+class MCAP_EXPORT FileStreamReader final : public IReadable {
 public:
   FileStreamReader(std::ifstream& stream);
 
@@ -153,7 +154,7 @@ private:
  * @brief ICompressedReader implementation that decompresses Zstandard
  * (https://facebook.github.io/zstd/) data.
  */
-class ZStdReader final : public ICompressedReader {
+class MCAP_PUBLIC ZStdReader final : public ICompressedReader {
 public:
   void reset(const std::byte* data, uint64_t size, uint64_t uncompressedSize) override;
   uint64_t read(std::byte** output, uint64_t offset, uint64_t size) override;
@@ -187,7 +188,7 @@ private:
  * @brief ICompressedReader implementation that decompresses LZ4
  * (https://lz4.github.io/lz4/) data.
  */
-class LZ4Reader final : public ICompressedReader {
+class MCAP_PUBLIC LZ4Reader final : public ICompressedReader {
 public:
   void reset(const std::byte* data, uint64_t size, uint64_t uncompressedSize) override;
   uint64_t read(std::byte** output, uint64_t offset, uint64_t size) override;
@@ -267,7 +268,7 @@ public:
 /**
  * @brief Provides a read interface to an MCAP file.
  */
-class McapReader final {
+class MCAP_PUBLIC McapReader final {
 public:
   ~McapReader();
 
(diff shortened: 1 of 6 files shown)


cpp/mcap/include/mcap/visibility.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#if defined _WIN32 || defined __CYGWIN__
+#ifdef __GNUC__
+#define MCAP_EXPORT __attribute__((dllexport))
+#define MCAP_IMPORT __attribute__((dllimport))
+#else
+#define MCAP_EXPORT __declspec(dllexport)
+#define MCAP_IMPORT __declspec(dllimport)
+#endif
+#ifdef MCAP_BUILDING_DLL
```
jhurliman, line 11: Where does `MCAP_BUILDING_DLL` get set?

cpp/mcap/include/mcap/visibility.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#if defined _WIN32 || defined __CYGWIN__
```
jtbandes, line 3: can we get some indentation in this file so it is legible?

emersonknapp: yes :+1: i think some windows vscode setting auto-changed this, it started indented - will fix

cpp/mcap/include/mcap/visibility.hpp
emersonknapp: I'll add a comment here in-file at least so it's not totally undocumented

cpp/mcap/include/mcap/visibility.hpp
emersonknapp: oh it was also clang-format didn't like it, I added the setting to allow for preprocessor indents

jtbandes: I've also seen some files keep the `#` at the beginning of the line and put the spaces right after it. Not sure which is better 🤷

jtbandes: oh I see that's exactly what you did 👍

cpp/mcap/include/mcap/visibility.hpp
```diff
@@ -0,0 +1,31 @@
...
+#  ifdef __GNUC__
+#    define MCAP_EXPORT __attribute__((dllexport))
+#    define MCAP_IMPORT __attribute__((dllimport))
+#  else
+#    define MCAP_EXPORT __declspec(dllexport)
+#    define MCAP_IMPORT __declspec(dllimport)
+#  endif
+// MCAP_BUILDING_DLL to be defined by build process producing a shared library
```
james-rms, line 11: Given this comment, it's not clear to me why I seemingly had to define `MCAP_BUILDING_DLL` for the test binaries to build in https://github.com/foxglove/mcap/pull/713 . Part of the problem is how opaque conan is about what is getting built as what, but my understanding is that `MCAP_BUILDING_DLL` is set for both the MCAP library and the consuming executable.

james-rms: in that context i think i'm not even building a DLL, i'm just including the source as a header-only library.

cpp/mcap/include/mcap/reader.hpp
```diff
@@ -131,7 +132,7 @@ class ICompressedReader : public IReadable {
  * @brief A "null" compressed reader that directly passes through uncompressed
  * data. No internal buffers are allocated.
  */
-class BufferReader final : public ICompressedReader {
+class MCAP_PUBLIC BufferReader final : public ICompressedReader {
```
james-rms, line 135: does it make sense that BufferReader is public but ICompressedReader is not? I get a warning about this here: https://github.com/foxglove/mcap/actions/runs/3415978239/jobs/5685665885#step:9:153

emersonknapp: It does not make sense, no. I've fixed it

cpp/mcap/include/mcap/visibility.hpp
emersonknapp: Discussed offline and will use `MCAP_IMPLEMENTATION` for `dllexport` since in an exe it doesn't do anything, and otherwise it's needed

cpp/mcap/include/mcap/visibility.hpp
```diff
@@ -0,0 +1,29 @@
...
+#  define MCAP_LOCAL
+#else
+#  define MCAP_EXPORT __attribute__((visibility("default")))
+#  define MCAP_IMPORT
+#  if __GNUC__ >= 4
+#    define MCAP_PUBLIC __attribute__((visibility("default")))
+#    define MCAP_LOCAL __attribute__((visibility("hidden")))
+#  else
+#    define MCAP_PUBLIC
+#    define MCAP_LOCAL
```
james-rms, line 26: looks like we're not using `MCAP_PUBLIC_TYPE` or `MCAP_LOCAL`, might be cleanest to 🔪

PR #712, opened 2022-11-07
Add ROS 2 storage plugin evaluation

### LINKED ISSUE #704: Publish performance analysis on mcap.dev

james-rms:
**Public-Facing Changes**
Adds a plugin throughput comparison to https://mcap.dev.

This PR leaves out the `rosbag2_performance_benchmarking` results - these are too noisy to draw conclusions from at this point, and will need some iteration on test setup to include.

Fixes #704


docs/performance-comparison/rosbag2-plugin-comparison.md
```diff
@@ -0,0 +1,152 @@
...
+
+- `sqlite_default`: The default configuration of the SQLite storage plugin, effectively measuring the performance that ROS 2 users get out-of-the-box. This is equivalent to running the following statements when opening a SQLite file:`PRAGMA journal_mode=MEMORY; PRAGMA synchronous=OFF;`. These settings produce the highest write performance available from SQLite at the risk of corruption in the case of an interruption or power loss.
+- `sqlite_resilient`: The "resilient" preset profile of the SQLite storage plugin. This setting is equivalent to running the following statements when opening a SQLite file: `PRAGMA journal_mode=WAL; PRAGMA synchronous=NORMAL`. This removes the risk of an entire `.db3` file being corrupted, at the cost of performance. By using a write-ahead log, only the messages in the last few transactions may be lost if the data from those transactions were not fully committed to disk. Therefore in this mode, SQLite3 offers similar guarantees against corruption as MCAP.
+- `mcap_default`: The default recording configuration of the MCAP storage plugin. Produces un-compressed, chunked MCAPs with a message index for efficient read performance.
```
MichaelOrlov, line 33: It is unclear what chunk size would be for configurations which are using chunking. In particular for the `mcap_default` configuration. 
Maybe it would be better to add some note about default chunk size and that it will be automatically adjusted if message size is bigger.

docs/performance-comparison/rosbag2-plugin-comparison.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Message sizes tested
+
+- 1MiB
+- 10KiB
+- 100B
+- Mixed: Messages from all of the above sizes at the following ratios (by number of bytes, not messages)
+  - 1MiB: 70%
+  - 10KiB: 20%
+  - 100B: 10%
```
MichaelOrlov, line 75: Here would be great to add some notes about using abbreviations at least for `MiB`.
For me it wasn't clear that MiB is a `Mebibyte` not a `Megabit` and it is equal to 2^20  or 1,048,576 bytes. A megabyte is equal to 10^6 1,000,000 bytes.

docs/performance-comparison/rosbag2-plugin-comparison.md
```diff
@@ -0,0 +1,152 @@
...
+When testing write throughput, 250MiB of messages are stored as quickly as possible.
+
+## Cache sizes tested
+
+The `rosbag2::SequentialWriter` collects messages in an internal cache before calling the storage plugin `write()` call with all messages at once. The cache size determines the size of each "batch" that is written to a storage plugin at once. This matters in particular for the SQLite storage plugin, because each batch constitutes a SQL transaction, and it's generally better for write performance to write many messages in one transaction.
+
+Cache sizes tested were:
+
+- 1KiB
+- 10MiB
```
MichaelOrlov, line 86: @james-rms You are mentioning that it was used cache size `1KiB` and `10Mib` for the `SequentialWriter`, but we are using number of messages not bytes for SequentialWriter's cache. 
It seems this is wrong statement in article.

MichaelOrlov (commented):
@james-rms Entire article and content presentation looks great!  
Although I've found some wrong statements, and have some proposals that need to clarify some details for readers.

Also it was unclear for me if it was using in memory `tempfs` or real disk write operations. 
It would make sense to explicitly mention about it in article. And if it was used real disk writes it would be useful to add disk write benchmark as well.

It's still unclear for me how to explain charts with message sizes: 1MiB and batch size: 10Mib and 1KiB on Apple M1 Pro
https://github.com/foxglove/mcap/blob/b568bb58d6acb24a378296bfedc232db2a97ca11/docs/performance-comparison/throughput/m1/1MiB_1KiB.png ? 
In particular that `mcap_default` with index and chunking has better throughput than `mcap_nochunking` ?

docs/performance-comparison/rosbag2-plugin-comparison.md
james-rms: Will clarify - this benchmark tries to pack enough messages into each write() batch to meet a target size, not a target number of messages.

docs/performance-comparison/rosbag2-plugin-comparison.md
```diff
@@ -0,0 +1,152 @@
...
+| rosbag2_storage_plugin_comparison | 0.1.0   |
+| rosbag2_storage_sqlite3           | 0.17.0  |
+| rosbag2_test_common               | 0.17.0  |
+| shared_queues_vendor              | 0.17.0  |
+| sqlite3_vendor                    | 0.17.0  |
+| zstd_vendor                       | 0.17.0  |
+
+## Message sizes tested
+
+- 1MiB
```
james-rms, line 69: add method of seelcting message content here as well.

docs/performance-comparison/rosbag2-plugin-comparison.md
james-rms: Actually, I think i'm right here. The cache buffer decides when it's full based on a number of bytes, not a number of messages. Also, the cache size parameter [here](https://github.com/ros2/rosbag2/blob/rolling/rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp#L31) gets passed down to [here](https://github.com/ros2/rosbag2/blob/c611a8010a0aaaba8bff47c42846aae2374fe81f/rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_buffer.hpp#L39), where it is interpreted as a limit on the size of the buffer in bytes.

docs/performance-comparison/rosbag2-plugin-comparison.md
```diff
@@ -0,0 +1,184 @@
+# Comparison of `rosbag2` Storage Plugins
+
+_Oct, 2022_  
+_James Smith ([@james-rms](https://github.com/james-rms))_  
+_[Foxglove](https://foxglove.dev)_
```
amacneil, line 5:
```suggestion
_James Smith ([@james-rms](https://github.com/james-rms))_  
_[Foxglove](https://foxglove.dev)_  
_November 2022_
```

docs/performance-comparison/rosbag2-plugin-comparison.md
```diff
@@ -0,0 +1,184 @@
+# Comparison of `rosbag2` Storage Plugins
+
+_Oct, 2022_
```
amacneil, line 3:
```suggestion
_Nov 2022_  
```

PR #713, opened 2022-11-08
Windows Action CI

### LINKED ISSUE #716: C++ client Windows compatibility and CI
Provide Windows CI and a warning-free build with proper DLL visibility attributes for the C++ mcap client

james-rms:
**Public-Facing Changes**
Add windows CI for the C++ library. This ensures that our C++ library builds and tests pass on windows.

Closes #716


emersonknapp:
I agree that it probably makes sense to fix Windows warnings rather than disabling, just because it adds another layer of sanity checking on various potential issues in the language

emersonknapp:
Discussion: windows 11 mins - I'm inclined to think it's not a problem.
Discussion: `fopen_s`, why not? Seems easy enough to use this alternative interface
Discussion: `C4251`, probably best to move forward via suppressing it for the near term at least. First users via ROS 2 will definitely have an environment using the same compiler/STL versions. Getting around it would consider significant rework and I'm sure it would even be a good idea

james-rms:
@emersonknapp I decided:

1. Since 11 minutes is twice as long as any other CI step, i decided to only build and run unit tests in CI. This should be sufficient to ensure that the core MCAP library builds on windows.
2. Since `fopen_s` is not available on many POSIX platforms, i've decided to stick with the more available standard function `fopen`.

jhurliman (approved):
I think elsewhere we do primitive typecasts with the form `uint32_t(index.records.size())` instead of `(uint32_t)(index.records.size())`, but otherwise LGTM

PR #726, opened 2022-11-15
Add Mcap typescript flatbuffer encoding example

snosenzo:
- Added an example showing how to create an mcap file using flatbuffers encoding in typescript

- schemas (not dependent, but uses output from compiling the added schemas) PR: https://github.com/foxglove/schemas/pull/70


typescript/examples/flatbufferswriter/scripts/main.ts
```diff
@@ -0,0 +1,258 @@
...
+    useChunks: true,
+    useChunkIndex: true,
+  });
+
+  await mcapFile.start({
+    profile: "",
+    library: "mcap example",
+  });
+  const FrameTransformSchemaBuffer = fs.readFileSync(
+    `${__dirname}/../../flatbuffer/bin/FrameTransform.bfbs`,
```
jtbandes, line 179: Let's get the .bfbs files stored with Git LFS so we don't bloat the repo size too much

typescript/examples/flatbuffer/output/Grid_generated.ts
```diff
@@ -0,0 +1,7 @@
+// automatically generated by the FlatBuffers compiler, do not modify
```
jtbandes, line 1: You can use a .gitattributes file to mark these as `linguist-generated` so they don't show up in PRs. See examples in the schemas repo.

typescript/examples/flatbufferswriter/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@foxglove/mcap-example-flatbuffer-writer",
```
jtbandes, line 2: let's have this match the dir name, so `typescript/examples/X` -> `@foxglove/mcap-example-X`. seems like other packages are doing this consistently already.

typescript/examples/flatbufferswriter/scripts/main.ts
```diff
@@ -0,0 +1,258 @@
...
+  const gridChannelId = await mcapFile.registerChannel({
+    schemaId: gridSchemaId,
+    topic: "grid",
+    messageEncoding: "flatbuffer",
+    metadata: new Map(),
+  });
+
+  const { mcapTimeLength, gridMessageFrequency } = scriptParameters;
+
+  const timeBetweenMessages = 1000 / gridMessageFrequency; // 1s (ms) / frequency
```
jtbandes, line 230: confused about units here, seems like ms / Hz would give ms*s which is a surprising unit...

typescript/examples/flatbufferswriter/scripts/flatbufferUtils.ts
```diff
@@ -0,0 +1,118 @@
...
+  FbPackedElementField.addName(builder, name);
+  FbPackedElementField.addOffset(builder, json.offset);
+  FbPackedElementField.addType(builder, json.type);
+  return FbPackedElementField.endPackedElementField(builder);
+}
+
+export function buildGridMessage(builder: Builder, json: Grid): number {
+  const frameId = builder.createString(json.frame_id);
+
+  const pose = buildPose(builder, json.pose);
```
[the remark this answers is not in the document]
snosenzo: The builder as far as I cant tell allows for abstracted use of the underlying buffer, expansion of buffer and tracks how much of it is allocated. So when you start a new object on the buffer and add attributes to it it allocates a given space for it in the continuous chunk and tracks where the next object can begin. It also allows the buffer to have a `root` schema, hence the importance of ending the buffer with the main message type. All of the fields and data in a message have to be in the same buffer so when you create/end a `Quaternion` the index of that quaternion needs to be used when adding it to the object of the schema that includes it so that it can reference it in the same buffer.

typescript/examples/flatbufferswriter/scripts/main.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  (grid as Grid).data = data;
+}
+
+async function addMessageAtTime(
+  mcapFile: McapWriter,
+  gridChannelId: number,
+  getMessageData: (time: number) => Message["data"],
+  time: bigint,
+): Promise<void> {
```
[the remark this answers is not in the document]
snosenzo: definitely right, will remove

typescript/examples/flatbufferswriter/scripts/main.ts
snosenzo: was trying to do 2 things at once: interval length is 1 / frequency and there are 1000 ms in a second, but yeah the comment wasn't clear. will udpate

typescript/examples/flatbufferswriter/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "jest": "27.3.1",
+    "lodash": "4.17.21",
+    "prettier": "2.4.1",
+    "protobufjs": "6.11.2",
+    "ts-jest": "27.0.7",
+    "ts-node": "10.4.0",
+    "tsconfig-paths": "3.12.0",
+    "typescript": "4.4.4",
+    "wasm-lz4": "2.0.0",
+    "zstd-codec": "^0.1.4"
```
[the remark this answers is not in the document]
snosenzo: lol these were just copied from the basicwriter, which also doesn't use some of these. will clean up both

typescript/examples/flatbufferswriter/scripts/main.ts
```diff
@@ -0,0 +1,258 @@
...
+    return BigInt(this.totalBytesWritten);
+  }
+}
+function roundUp(numToRound: number, multiple: number) {
+  if (multiple !== 0 && (multiple & (multiple - 1)) === 0) {
+    return (numToRound + multiple - 1) & -multiple;
+  } else {
+    throw Error(`invalid multiple ${multiple}, num to round ${numToRound}`);
+  }
+}
```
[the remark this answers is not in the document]
snosenzo: I just simplified this code. Not sure if there's a lodash equivalent. It's not strictly necessary but nice to  have

typescript/examples/flatbufferswriter/scripts/main.ts
```diff
@@ -0,0 +1,247 @@
...
+
+const gridParameters = {
+  cell_size: { x: 0.1, y: 0.1 },
+  column_count: 512,
+  row_count: 512,
+  cell_stride: 4,
+  row_stride: 0,
+};
+
+// set row stride to be next largest multiple of 2
```
jtbandes, line 49:
```suggestion
// set row stride to be next largest power of 2
```

jtbandes (approved):

PR #729, opened 2022-11-16
CLI: filtering with multiple exclude matchers works

### LINKED ISSUE #724: CLI : Unable to exclude multiple topics
**Description**

I am trying to use the mcap CLI to remove multiple topics from a file that was recorded from a ROS2 environment. Unfortunately, when I specify multiple `-n` flags, no topics are filtered. On the other hand, using a single `-n` flag with a more complex regex works, so this seems to be down to argument handling.

- Version:  `mcap version` says `0.2.0` but I downloaded `0.0.23`, not sure which is which
- Platform: Linux amd64

**Steps To Reproduce**

- Get a mcap capture
- Run `mcap filter input.mcap -n /topic_1 -n /topic_2 -o filtered.mcap`

**Expected Behavior**

`filtered.mcap` should contain neither `topic_1` nor `topic_2`

**Actual Behavior**

`filtered.mcap` contains both

james-rms:
**Public-Facing Changes**
users can now use multiple exclude patterns with `mcap filter` and it will work.

**Description**

Fixes #724

### DIFF
--- a/go/cli/mcap/cmd/filter.go
+++ b/go/cli/mcap/cmd/filter.go
@@ -240,16 +240,25 @@ func filter(
 			if err != nil {
 				return err
 			}
+			// if any topics match an includeTopic, add it.
 			for _, matcher := range opts.includeTopics {
 				if matcher.MatchString(channel.Topic) {
 					channels[channel.ID] = markableChannel{channel, false}
 				}
 			}
-			for _, matcher := range opts.excludeTopics {
-				if !matcher.MatchString(channel.Topic) {
+			// if a topic does not match any excludeTopic, add it.
+			if len(opts.excludeTopics) != 0 {
+				shouldInclude := true
+				for _, matcher := range opts.excludeTopics {
+					if matcher.MatchString(channel.Topic) {
+						shouldInclude = false
+					}
+				}
+				if shouldInclude {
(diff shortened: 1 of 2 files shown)


go/cli/mcap/cmd/filter.go
```diff
@@ -240,16 +240,25 @@ func filter(
...
+				shouldInclude := true
+				for _, matcher := range opts.excludeTopics {
+					if matcher.MatchString(channel.Topic) {
+						shouldInclude = false
+					}
+				}
+				if shouldInclude {
 					channels[channel.ID] = markableChannel{channel, false}
 				}
 			}
```
jtbandes, line 260: break?

Wonder if the outer if is necessary – seems like this could be structured as "if no exclude topics match, add the channel"

jtbandes (approved):

go/cli/mcap/cmd/filter.go
james-rms: we're in nested for-loop territory, breaks can get confusing and don't seem neccessary here.
The outer if can't be trivially removed, otherwise if the user specifies no excludeTopics but they do specify includeTopics, all topics will get included no matter what. I could change the condition to `if len(opts.includeTopics) == 0` but not sure if that'd reduce complexity or increase readability.

PR #730, opened 2022-11-17
mcap_protobuf: add MessageFactory internal to Decoder

james-rms:
**Public-Facing Changes**

**Description**
For some versions of the `protobuf` library, `GetMessages` fails when called more than once with the same set of definitions. This can happen because GetMessages uses an internal global MessageFactory which is shared between all callers, and this raises a TypeError if you attempt to add the same definitions to the DescriptorPool more than once:

```
.../pip_protobuf/google/protobuf/message_factory.py in _AddFile(file_proto)
    180         # Remove from elements to be visited, in order to cut cycles.
    181         _AddFile(file_by_name.pop(dependency))
--> 182     _FACTORY.pool.Add(file_proto)
    183   while file_by_name:
    184     _AddFile(file_by_name.popitem()[1])

TypeError: Couldn't build proto file into descriptor pool!
Invalid proto descriptor for file "google/protobuf/timestamp.proto":
  google/protobuf/timestamp.proto: A file with this name is already in the pool.
```

This PR uses an internal MessageFactory for each decoder instance, which should ensure that a given DescriptorPool only gets each FileDescriptorProto once.


python/mcap-protobuf-support/mcap_protobuf/decoder.py
```diff
@@ -18,6 +18,29 @@ class Decoder:
...
+
+        def _add_file(file_proto: FileDescriptorProto):
+            for dependency in file_proto.dependency:
+                if dependency in file_by_name:
+                    # Remove from elements to be visited, in order to cut cycles.
+                    _add_file(file_by_name.pop(dependency))
+            self._factory.pool.Add(file_proto)
+
+        while file_by_name:
+            _add_file(file_by_name.popitem()[1])
```
jtbandes, line 39: Add comment? "Adapted from https://github.com/protocolbuffers/protobuf/blob/72977f951c1f20807534a9d9af4e9583c6a9049b/python/google/protobuf/message_factory.py#L154"

jtbandes (approved):
What an adventure

Issue #736, opened 2022-11-20
specification: split attachment length into a separate field from attachment data


wkalt:
Currently the `Data` field of attachment records is described as length-prefixed bytes. I think we should split it into two fields: Data size and Data.

This does not affect the layout of binary data, but would require a breaking change to SDKs. The reason for the change is supporting streaming writes of attachments. When writing an attachment from a stream, the length needs to be written prior to the stream body. It is not in general possible to know the length of a stream until the whole thing has been read (into the output file), which means backfilling the length would require a backwards seek.

Rather than relying on a seekable writer, we should have the data length supplied up-front in the attachment type supplied to WriteAttachment, and instead rely on the user to fetch it beforehand through some mechanism. For the typical case of attaching files read off disk this will be easy to do.

PR #737, opened 2022-11-20
Go: add support for stream handling of attachments

wkalt:
Adds support for streaming reads and writes of attachment data. With this change, ParseAttachment is dropped as a public function and replaced with a private function, parseAttachmentReader.

A new type AttachmentReader is exposed, which exposes the Data portion of an attachment as a reader that can be streamed. The Attachment type is modified to include a new ContentLength field, to allow the writer to write the attachment length before stream processing the body.

A new option, AttachmentCallback, is exposed in the Lexer. Rather than emitting attachments, the Lexer now parses an attachment reader and calls the supplied callback (if supplied), and consumes any unconsumed bytes from the attachment record (whether or not a callback is supplied).

This is a breaking change in the Go SDK, and introduces some awkwardness due to the presence of the new ContentLength field in the Attachment record. In the specification, that value is folded into the Data field, however this is not a tenable approach for stream handling. The actual binary format is not affected by this - only the way in which the binary format is described.

This change is required for readers and writers to be able to handle attachments bigger than available RAM. Without it, attaching a sufficiently large attachment will crash the reader/writer with an OOM.

**Public-Facing Changes**

**Description**

### DIFF
--- a/go/mcap/lexer.go
+++ b/go/mcap/lexer.go
@@ -80,8 +80,6 @@ const (
 	TokenMessageIndex
 	// TokenChunkIndex represents a chunk index token.
 	TokenChunkIndex
-	// TokenAttachment represents an attachment token.
-	TokenAttachment
 	// TokenAttachmentIndex represents an attachment index token.
 	TokenAttachmentIndex
 	// TokenStatistics represents a statistics token.
@@ -122,8 +120,6 @@ func (t TokenType) String() string {
 		return "message index"
 	case TokenChunkIndex:
 		return "chunk index"
-	case TokenAttachment:
-		return "attachment"
 	case TokenAttachmentIndex:
 		return "attachment index"
 	case TokenStatistics:
@@ -155,10 +151,12 @@ type Lexer struct {
 	inChunk                  bool
 	buf                      []byte
 	uncompressedChunk        []byte
-	validateCRC              bool
+	validateChunkCRCs        bool
+	computeAttachmentCRCs    bool
 	emitInvalidChunks        bool
 	maxRecordSize            int
 	maxDecompressedChunkSize int
+	attachmentCallback       func(*AttachmentReader) error
 }
 
 // Next returns the next token from the lexer as a byte array. The result will
@@ -191,16 +189,52 @@ func (l *Lexer) Next(p []byte) (TokenType, []byte, error) {
 		if l.maxRecordSize > 0 && recordLen > uint64(l.maxRecordSize) {
 			return TokenError, nil, ErrRecordTooLarge
 		}
-		if opcode == OpChunk && !l.emitChunks {
-			err := loadChunk(l, recordLen)
-			if err != nil {
-				if l.emitInvalidChunks {
-					var invalidCrc *errInvalidChunkCrc
-					if errors.As(err, &invalidCrc) {
-						return TokenInvalidChunk, nil, err
+
+		// Chunks and attachments require special handling to avoid
+		// materialization into RAM. If it's a chunk, open up a decompressor and
+		// swap it in as the active reader, then continue on the next message
+		// (which will be from the chunk data). If it's an attachment, parse the
+		// record into an AttachmentReader and call any user-supplied callback.
+		// Then discard any remaining data and continue to the next record.
+		switch opcode {
+		case OpChunk:
+			if !l.emitChunks {
+				err := loadChunk(l, recordLen)
+				if err != nil {
+					if l.emitInvalidChunks {
+						var invalidCrc *errInvalidChunkCrc
+						if errors.As(err, &invalidCrc) {
+							return TokenInvalidChunk, nil, err
+						}
 					}
+					return TokenError, nil, err
 				}
-				return TokenError, nil, err
+				continue
+			}
+		case OpAttachment:
+			limitReader := &io.LimitedReader{
+				R: l.reader,
+				N: int64(recordLen),
+			}
+
+			if l.attachmentCallback != nil {
+				attachmentReader, err := parseAttachmentReader(
+					limitReader,
+					l.computeAttachmentCRCs,
+				)
+				if err != nil {
+					return TokenError, nil, fmt.Errorf("failed to parse attachment: %w", err)
+				}
+				err = l.attachmentCallback(attachmentReader)
+				if err != nil {
+					return TokenError, nil, fmt.Errorf("failed to handle attachment: %w", err)
+				}
+			}
+
+			// skip the base reader ahead to cover any unconsumed bytes of the attachment
+			err := skipReader(limitReader.R, limitReader.N)
+			if err != nil {
+				return TokenError, nil, fmt.Errorf("failed to consume unhandled attachment data: %w", err)
 			}
 			continue
 		}
@@ -238,8 +272,6 @@ func (l *Lexer) Next(p []byte) (TokenType, []byte, error) {
 			return TokenChannel, record, nil
 		case OpFooter:
 			return TokenFooter, record, nil
-		case OpAttachment:
-			return TokenAttachment, record, nil
 		case OpAttachmentIndex:
 			return TokenAttachmentIndex, record, nil
 		case OpChunkIndex:
@@ -394,7 +426,7 @@ func loadChunk(l *Lexer, recordLen uint64) error {
 	// validation. If we are not validating CRCs, we can use incremental
 	// decompression for the chunk's data, which may be beneficial to streaming
 	// readers.
-	if l.validateCRC {
+	if l.validateChunkCRCs {
 		if l.maxDecompressedChunkSize > 0 && uncompressedSize > uint64(l.maxDecompressedChunkSize) {
 			return ErrChunkTooLarge
 		}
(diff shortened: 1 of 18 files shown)


wkalt:
The point raised in the commit message about ContentLength relates to this issue: https://github.com/foxglove/mcap/issues/736

Before merging this patch we should pick what name we want to use for that field, if people are on board with the idea of splitting Data into ContentLength and Data. I don't feel too strongly about whether that spec change needs to go in before or after this merges, but before is probably slightly preferable.

Also whatever approach we go with for stream handling (a callback mechanism in this patch) should probably land in the other SDKs at some point, so we should make sure we are all comfortable with the approach.

go/mcap/lexer.go
```diff
@@ -191,16 +188,43 @@ func (l *Lexer) Next(p []byte) (TokenType, []byte, error) {
...
+				return TokenError, nil, fmt.Errorf("failed to parse attachment: %w", err)
+			}
+			if l.attachmentCallback != nil {
+				err := l.attachmentCallback(attachmentReader)
+				if err != nil {
+					return TokenError, nil, fmt.Errorf("failed to handle attachment: %w", err)
 				}
-				return TokenError, nil, err
+			}
+			_, err = io.Copy(io.Discard, limitReader)
```
james-rms, line 225: can we check if this is a Seeker and seek ahead instead if possible?

go/mcap/parse.go
```diff
@@ -264,41 +264,40 @@ func ParseChunkIndex(buf []byte) (*ChunkIndex, error) {
 	}, nil
 }
 
-// ParseAttachment parses an attachment record.
-func ParseAttachment(buf []byte) (*Attachment, error) {
-	logTime, offset, err := getUint64(buf, 0)
+func parseAttachmentReader(r io.Reader) (*AttachmentReader, error) {
+	buf := make([]byte, 8)
+	crcReader := newCRCReader(r)
```
james-rms, line 269: Can we make CRC validation optional somehow?

go/mcap/parse.go
```diff
@@ -264,41 +264,40 @@ func ParseChunkIndex(buf []byte) (*ChunkIndex, error) {
...
-		return nil, fmt.Errorf("failed to read CRC: %w", err)
-	}
-	return &Attachment{
-		LogTime:    logTime,
-		CreateTime: createTime,
-		Name:       name,
-		MediaType:  mediaType,
-		Data:       data,
-		CRC:        crc,
+	limitReader := io.LimitReader(crcReader, int64(dataSize))
```
james-rms, line 290: how do you feel about implementing a custom `LimitedReadSeeker` and returning that instead? The LimitReader source is not particularly complicated, it doesn't feel too hard to adapt.

go/mcap/mcap.go
```diff
@@ -214,12 +216,65 @@ type ChunkIndex struct {
...
+	crcReader  *crcReader
+	crc        *uint32
+}
+
+// ComputedCRC discards any remaining data in the Data portion of the
+// AttachmentReader, then returns the checksum computed from the fields of the
+// attachment up to the CRC. If it is called before the data portion of the
+// reader has been fully consumed, an error will be returned.
+func (ar *AttachmentReader) ComputedCRC() (uint32, error) {
+	n, err := io.Copy(io.Discard, ar.Data)
```
james-rms, line 246: could alternatively cast this back to [LimitedReader](https://cs.opensource.google/go/go/+/go1.19.3:src/io/io.go;l=465) and check N

wkalt, line 246: this should just read one byte - that's all that's necessary to check if we need to fail

go/mcap/parse.go
wkalt: what's the need for seeking?

wkalt: also, I think that's going to require retention of the already-read data in memory if it doesn't exist on disk. Exposing io.Reader seems correct to me here; the limit aspect is just to ensure it EOFs at the right place.

go/mcap/parse.go
wkalt: yeah, I can add the option to skip the computation of CRC. Actual validation of the CRC is up to the callback function the user supplies.

go/mcap/lexer.go
wkalt: yep, will do

go/mcap/mcap.go
wkalt: yeah - I think maybe the thing to do here is convert `Data` to a private `data` typed as a LimitReader, and then expose it through a `Data() io.Reader` method. This can then avoid the cast and otherwise do as you suggest. No need to read.

wkalt: same considerations should apply to the next function as well

go/mcap/lexer.go
```diff
@@ -451,19 +473,24 @@ type LexerOptions struct {
 	// MaxRecordSize defines the maximum size record the lexer will read.
 	// Records larger than this will result in an error.
 	MaxRecordSize int
+
+	// AttachmentCallback
+	AttachmentCallback func(*AttachmentReader) error
```
wkalt, line 478: expand comment

go/mcap/lexer_test.go
```diff
@@ -309,6 +301,76 @@ func TestChunkCRCValidation(t *testing.T) {
...
+					parsedCRC, err := ar.ParsedCRC()
+					assert.Nil(t, err)
+					assert.Equal(t, computedCRC, parsedCRC)
+					called = true
+					return nil
+				},
+			})
+
+			for !errors.Is(err, io.EOF) {
+				_, _, err = lexer.Next(nil)
```
wkalt, line 367: make some assertion here for clarity, should be nil or io.EOF

go/mcap/mcap.go
```diff
@@ -214,12 +216,62 @@ type ChunkIndex struct {
...
+}
+
+// AttachmentReader represents an attachment for handling in a streaming manner.
+type AttachmentReader struct {
+	LogTime       uint64
+	CreateTime    uint64
+	Name          string
+	MediaType     string
+	ContentLength uint64
+	data          *io.LimitedReader
```
wkalt, line 234: group this with the other private fields

go/mcap/testutils.go
```diff
@@ -142,7 +142,15 @@ func record(op OpCode) []byte {
 }
 
 func attachment() []byte {
-	buf := make([]byte, 9)
+	recordLen := 9 + // opcode + record length
+		8 + // record time
+		8 + // create time
+		4 + // attachment name length
+		4 + // media type length
+		4 + // data size
```
wkalt, line 150: seems like this should be 8. why doesn't this cause a parse failure? I guess because the test isn't doing anything with the CRC and is interpreting the CRC bytes as part of the data size. Since it's all zeros due to the empty content that would erroneously pass.

go/mcap/lexer.go
wkalt: added that

go/mcap/mcap.go
wkalt: added in both places

go/mcap/parse.go
james-rms: If the underlying reader is a Seeker (which we already specialize for in some cases), then it may make sense to hand the user the ability to seek through the attachment for parts that they want.

wkalt: I'm struggling to see a way to make this work that won't seem clunky or work against the CRC validation features.

* CRC validation seems like it would definitely break as soon as the user seeks around the attachment. If they don't read the whole thing, or read it in the wrong order, their CRC isn't going to validate.
* I don't think AttachmentReader should type the Data section with an io.ReadSeeker if there is any possibility that the thing cannot Seek (which is the case when doing a streaming read). So I think it would return an io.Reader, and if we got this working in some manner it would be up to the caller to attempt to cast to a ReadSeeker, and if successful seek away and break their own CRC validation.
* The brute force mechanism I think would be to make CRCReader a public type, and expose its underlying reader. Since the underlying reader is a LimitedReader that also exposes its underlying reader, by just making the CRCReader public the user would be able to call, effectively, `rs, ok := ar.Data().(CRCReader).R.(LimitedReader).R.(io.ReadSeeker)` and seek on that if successful. That would actually need to be split across multiple type assertions in case one of the intermediate ones fail. Definitely seems too clunky to me to ship.

Open to ideas on clean ways to handle this. I don't think it's a solution to just implement a LimitedReadSeeker and return that (even behind the io.Reader) because in common cases it would not actually be seekable. My inclination at the moment is to keep this as is and force the reader to consume the whole attachment (assuming they want any of it), and wait for feedback before optimizing. I think the usual case for dealing with attachments in mcap files is going to be to dump them out to disk for post-processing rather than do seeking reads on them in place - but we won't know till it gets some use.

If users have a Seekable reader they also have the option of addressing the attachment via the index and seeking around however they want, which the SDK fully supports. IMO that is probably enough to cover this usecase for seeking reads on an attachment in-place.

james-rms (approved):
This looks like it's going in a good direction - once it's possible to skip CRC calculation when reading attachments, I'm happy for this to merge.

Issue #686, opened 2022-10-28
CLI: Ability to write in columnar order


amacneil:
Ability to create mcap files (probably using `mcap merge` cli) in columnar (i.e. channel oriented) order.

e.g.

- no more than one channel per chunk
- fully write each channel before writing the next channel

This would allow more efficient remote access of multi-channel files (readers can easily fetch an entire topic without scanning the whole file).

james-rms:
Preliminary thoughts:

- MCAPs laid out in column order are difficult for the standard studio use-case: playing back several topics in time order, because a reader may need to decompress the entire chunk for each topic it needs before playing any messages, which could result in loading the entire uncompressed message content into memory.
- This can be mitigated if the reader knows before decompressing that all messages in a given chunk are in time order internally. then it can decompress each chunk progressively while emitting messages rather than decompressing the entire chunk.

wkalt:
Additional notes -

We've discussed the strategy above, as well as an alternative of using attachments to attach topics intended for preloading to the end of the file. The way I imagine that would work would be an `mcap optimize [file] [-t topic]` command, which would scan the input file in linear order, write to memory or disk a new mcap file containing only the specified topics (in original file order), and attach that mcap to the end of the original mcap with a specific naming convention (which would probably need to go in the spec or some implementation notes eventually).

A reader can leverage this by reading the EOF index of the attachment to determine what topics are preloadable, and if the topic being sought is in that list, download that entire attachment prior to downloading the main file.

The advantages of this approach as I see them are,
* It does not take any extra memory to read or write, vs the original file
* Row-oriented read performance characteristics of the original file are entirely preserved, no matter how many channels in the original file.
* It can be implemented without overwriting the original file (pending implementation of https://github.com/foxglove/mcap/issues/677)
* It means all preloaded topics can be fetched with a single linear read of the attachment (the expectation being that typical topics that require preloading, for plots and things are not the heavy topics).

Disadvantages I see are,
* If you try and attach large topics to the end of file attachment, you will suffer the same as in the original case (though this is also true of trying to preload large messages from the "column-based" layout)
* It requires a new convention to be documented in the spec. It's unclear if this should be an mcap feature or a studio/foxglove-cli feature.
[...]

PR #747, opened 2022-11-25
CLI: add columnize subcommand

james-rms:
**Public-Facing Changes**

**Description**

Fixes https://github.com/foxglove/mcap/issues/686


wkalt:
If we end up baking this functionality into the tool, I'd propose calling it something more like "sort" or "group by" than columnize (or some combination of sort and group by - sort can also be used to get a physically time-ordered file, but some modifier seems needed to convey the partitioning into group-wise chunks, which doesn't apply in the case of a time sort). The reason is that storing something in a columnar format specifically means that columns (e.g fields in ROS schemas) would be stored contiguously, whereas this is still storing rows contiguously, just in a different order. The concept is similar but the answer to "is mcap a columnar format" would definitely be "no" in standard usage of the term. A row-oriented database stores rows in different contiguous tables just like the chunks here - but that doesn't make it columnar.

In terms of implementation and thinking about this, I think memory usage is going to be the main thing to evaluate. The go reader (and presumably the others) will hold fully decompressed whatever chunks are backing the messages in its heap, so if you have hundreds (or thousand+) topics and you are recording large files (such that you have enough messages on your topics to make full chunks), you can end up needing gigabytes of RAM to read one of these.

We have seen mcap files in the wild that are both in the 10s of GB in size, and that have over a thousand topics inside, so the scenario isn't hypothetical. The main issue that results from this IMO is that tooling/processes built for MCAP files needs to size memory allocations sufficiently to handle two very different kinds of requirements - so a process that handles large files on 1GB now might need 8GB or more if both types of input are anticipated.

james-rms:
Closing for now.

Issue #748, opened 2022-11-26
go: lexer leaks a goroutine when lexing zstd-compressed mcap


wkalt:
The go lexer (and reader) currently do not expose a Close() method, but the zstd reader in use requires a Close() to be called, or it will leak a goroutine. The lz4 reader does not require a Close().

We need to update both the lexer and reader interfaces to expose Close(). The lexer's method must call the zstd reader's Close method when zstd is in use, and the reader's must call the lexer's Close().

PR #749, opened 2022-11-26
Go: address goroutine leak when lexing zstd-compressed mcap

wkalt:
Adds Close() methods to the Lexer and Reader types. These are required to avoid a goroutine leak when working with zstd-compressed files.

Fixes: https://github.com/foxglove/mcap/issues/748

### DIFF
--- a/go/cli/mcap/cmd/attachment.go
+++ b/go/cli/mcap/cmd/attachment.go
@@ -77,6 +77,7 @@ var getAttachmentCmd = &cobra.Command{
 			if err != nil {
 				return fmt.Errorf("failed to construct reader: %w", err)
 			}
+			defer reader.Close()
 			info, err := reader.Info()
 			if err != nil {
 				return fmt.Errorf("failed to get mcap info: %w", err)
--- a/go/cli/mcap/cmd/cat.go
+++ b/go/cli/mcap/cmd/cat.go
@@ -192,6 +192,7 @@ var catCmd = &cobra.Command{
 			if err != nil {
 				die("Failed to create reader: %s", err)
 			}
+			defer reader.Close()
 			it, err := reader.Messages(getReadOpts(false)...)
 			if err != nil {
 				die("Failed to read messages: %s", err)
@@ -213,6 +214,7 @@ var catCmd = &cobra.Command{
 			if err != nil {
 				return fmt.Errorf("failed to create reader: %w", err)
 			}
+			defer reader.Close()
 			it, err := reader.Messages(getReadOpts(true)...)
 			if err != nil {
 				return fmt.Errorf("failed to read messages: %w", err)
--- a/go/cli/mcap/cmd/cat_test.go
+++ b/go/cli/mcap/cmd/cat_test.go
@@ -32,6 +32,7 @@ func TestCat(t *testing.T) {
 		t.Run(c.assertion, func(t *testing.T) {
 			reader, err := mcap.NewReader(r)
 			assert.Nil(t, err)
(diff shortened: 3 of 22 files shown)


wkalt:
I discovered the leak by running this program (note this includes the new Close() method - to test code prior to the change just remove that):

```
package main

import (
        "errors"
        "fmt"
        "io"
        "log"
        "os"

        "github.com/foxglove/mcap/go/mcap"
        "github.com/foxglove/mcap/go/mcap/readopts"
        "github.com/go-echarts/statsview"
)

func main() {
        mgr := statsview.New()
        go mgr.Start()
        defer mgr.Stop()

        // a zstd-compressed mcap file
        inputfile := "demo.mcap"

        c := 0
        for {
                fmt.Printf("iteration %d\n", c)
                f, err := os.Open(inputfile)
                if err != nil {
                        log.Fatal(err)
                }
                reader, err := mcap.NewReader(f)
                if err != nil {
                        log.Fatal(err)
                }
                defer reader.Close()
                iterator, err := reader.Messages(readopts.UsingIndex(true), readopts.InOrder(readopts.LogTimeOrder))
                if err != nil {
                        log.Fatal(err)
                }
                for {
                        _, _, _, err := iterator.Next(nil)
                        if errors.Is(err, io.EOF) {
                                break
                        }
                        if err != nil {
                                log.Fatal(err)
                        }
                }
                c++
                f.Close()
        }
}
```

Using that statsview package, you can visit the browser at http://localhost:18066/debug/statsview to see some useful graphs. The relevant ones are,

Before:


After:
fixed

The before plot reflects state after approximately 100 iterations. The after plot reflects state around 500 iterations.

wkalt:
The CI failures seem unrelated to this change but they are (partially) reproducible.

At one point we were failing,
* swift conformance
* python pages
* swift pages

I changed to test image from ubuntu-latest to ubuntu:20.04 and this seemed to resolve the swift conformance test, however the other two still fail.

Github has recently updated their ubuntu runner images. This is probably the root of the failure.

wkalt:
I have reverted the CI image pinning from this patch. We will need to follow that up in a separate PR.

wkalt:
I was looking over that reproducer code and realized it shouldn't work. The defer reader.Close() should not happen until the main() function exits, so if there is a leak the code there should still leak it. Yet it doesn't leak. This violates my expectations but it's possible the compiler is handling this intelligently.

I have two versions of the reproducer. One is running against the released v0.3.0 library (which does not include the Close()) method. This one leaks as pictured.

For the other one I have moved the exact same code into the mcap repo to test with my patch, using a replace directive in go.mod. This one does not leak, whether or not the Close() is present. This is the code pasted above.

The one in the standalone module doesn't leak on lz4 data, only zstd. There seems to be an interesting discrepancy between behavior with the released module and the code in main when run with a replace directive (the code has not changed in this area since last release).

wkalt:
This is the commit in the zstd library that fixed this issue:
https://github.com/klauspost/compress/commit/308a7519695d912cb0356f39209690d514396632

we are absorbing it in the included upgrade. I am still not sure whether the root cause is a bug in the zstd library or an issue with our usage, but from my reading it seems like calling Close() on the reader was not properly closing down the goroutines that patch removes.

wkalt:
@james-rms this is ready for a look now. I still don't have a full grip on the behavior I observe in the old code, but the zstd upgrade does seem to resolve the problem. I can demonstrate what I see on a call.

james-rms (approved):

PR #754, opened 2022-11-28
rust: writer automatically breaks chunks at size

### LINKED ISSUE #727: rust: no way to limit chunk size automatically
**Description**

Right now if you write an MCAP with the rust client, you will end up with one large chunk full of all of your messages unless you explicitly call `finish_chunk`. This can result in giant chunks that are larger than `2**31` bytes. This can cause errors in readers which assume that a chunk can be held in memory all at once.

- Version: `releases/rust/v0.5.0`
- Platform: All

**Steps To Reproduce**

**Expected Behavior**

**Actual Behavior**

james-rms:
**Public-Facing Changes**

Adds a new writer option, `chunk_size`, which will automatically break up chunks at a target (uncompressed) size.

**Description**

Fixes #727

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -107,6 +107,7 @@ fn write_record<W: Write>(w: &mut W, r: &Record) -> io::Result<()> {
 pub struct WriteOptions {
     compression: Option<Compression>,
     profile: String,
+    chunk_size: Option<u64>,
 }
 
 impl Default for WriteOptions {
@@ -117,6 +118,7 @@ impl Default for WriteOptions {
             #[cfg(not(feature = "zstd"))]
             compression: None,
             profile: String::new(),
+            chunk_size: None,
         }
     }
 }
@@ -126,6 +128,7 @@ impl WriteOptions {
         Self::default()
     }
 
+    /// Specifies the compression that should be used on chunks.
     pub fn compression(self, compression: Option<Compression>) -> Self {
         Self {
             compression,
@@ -133,9 +136,23 @@ impl WriteOptions {
         }
     }
 
+    /// specifies the profile that should be written to the MCAP Header record.
     pub fn profile<S: Into<String>>(self, profile: S) -> Self {
         Self {
             profile: profile.into(),
+            ..self
+        }
+    }
+
+    /// specifies the target uncompressed size of each chunk.
+    ///
+    /// Messages will be written to chunks until the uncompressed chunk is larger than the
+    /// target chunk size, at which point th  chunk will be closed and a new one started.
+    /// If `None`, chunks will not be automatically closed and the user must call `flush()` to
+    /// begin a new chunk.
+    pub fn chunk_size(self, chunk_size: Option<u64>) -> Self {
+        Self {
+            chunk_size: chunk_size,
             ..self
         }
     }
@@ -152,7 +169,7 @@ impl WriteOptions {
 /// and check for errors when done; otherwise the result will be unwrapped on drop.
 pub struct Writer<'a, W: Write + Seek> {
     writer: Option<WriteMode<W>>,
-    compression: Option<Compression>,
+    options: WriteOptions,
     schemas: HashMap<Schema<'a>, u16>,
     channels: HashMap<Channel<'a>, u16>,
     stats: records::Statistics,
@@ -172,14 +189,14 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
         write_record(
             &mut writer,
             &Record::Header(records::Header {
-                profile: opts.profile,
+                profile: opts.profile.clone(),
                 library: String::from("mcap-rs-") + env!("CARGO_PKG_VERSION"),
             }),
         )?;
 
         Ok(Self {
             writer: Some(WriteMode::Raw(writer)),
-            compression: opts.compression,
+            options: opts,
             schemas: HashMap::new(),
             channels: HashMap::new(),
             stats: records::Statistics::default(),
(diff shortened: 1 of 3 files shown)


james-rms:
cc. @mrkline

foxymiles (approved):

rust/src/write.rs
```diff
@@ -765,6 +795,8 @@ impl<W: Write + Seek> ChunkWriter<W> {
                 data: Cow::Borrowed(data),
             },
         )?;
+        // if the chunk is now larger than the chunk size limit, start a new chunk.
```
james-rms, line 798: remove this

rust/src/write.rs
```diff
@@ -277,6 +293,20 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
+        };
+        if let (Some(current_chunk_size), Some(target)) =
+            (current_chunk_size, self.options.chunk_size)
+        {
+            if current_chunk_size > target {
+                self.finish_chunk()?;
+            }
+        }
+
         self.chunkin_time()?.write_message(header, data)?;
```
jtbandes, line 310: > self.chunkin_time

wut

rust/tests/chunks.rs
```diff
@@ -0,0 +1,50 @@
...
+use tempfile::tempfile;
+
+/// Checks that the writer will automatically close chunks when they hit a target size.
+#[test]
+fn auto_cut_chunks() -> Result<()> {
+    let mapped = mcap_test_file()?;
+
+    let messages = mcap::MessageStream::new(&mapped)?;
+
+    let mut tmp = tempfile()?;
```
jtbandes, line 19: seems like this could be a bit cleaner (and less `unsafe`) if we just write to memory rather than a temp file

jtbandes (approved):

rust/src/write.rs
mrkline: It's ~~turtles~~ [memes](https://knowyourmeme.com/memes/its-morbin-time) all the way down.

rust/src/write.rs
```diff
@@ -126,20 +128,34 @@ impl WriteOptions {
...
+    /// specifies the profile that should be written to the MCAP Header record.
     pub fn profile<S: Into<String>>(self, profile: S) -> Self {
         Self {
             profile: profile.into(),
             ..self
         }
     }
 
+    /// specifies the target size of each chunk. Messages will be written to chunks
+    /// until the chunk size is larger than the target chunk size, at which point the
```
mrkline, line 148: Nit:

```suggestion
    /// specifies the target size of each chunk.
    ///
    /// Messages will be written to chunks until the chunk size
    /// is larger than the target chunk size, at which point the
```

Otherwise rustdoc will put the whole paragraph in the "brief" description.

rust/tests/flush.rs
```diff
@@ -16,7 +16,9 @@ fn flush_and_cut_chunks() -> Result<()> {
     let messages = mcap::MessageStream::new(&mapped)?;
 
     let mut tmp = tempfile()?;
-    let mut writer = mcap::Writer::new(BufWriter::new(&mut tmp))?;
+    let mut writer = mcap::WriteOptions::new()
+        .chunk_size(None)
+        .create(BufWriter::new(&mut tmp))?;
```
mrkline, line 21: Is the goal here just to be explicit? Asking since the default is to not automatically start new chunks.

rust/tests/chunks.rs
mrkline: Yeah, that's true of all the tests. I think we were originally testing with larger inputs, and I didn't want to assume they could fit in-memory. But now that `mcap_test_file()` is 2 meg, we should just write to a `Vec<u8>`.

I can open a PR to do that unless one of you wants to beat me to it.

rust/src/write.rs
```diff
@@ -277,6 +293,20 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
             .entry(header.channel_id)
             .or_insert(0) += 1;
 
+        // if the current chunk is larger than our target chunk size, finish it
+        // and start a new one.
```
mrkline, line 297: Should we be explicit (both here and in the docs for the option) that we're talking about a target _uncompressed_ size for the chunk? That seems to be the proximal problem in #727 AIUI - readers might want to make sure an uncompressed chunk will fit in a certain amount of memory.

rust/tests/chunks.rs
```diff
@@ -0,0 +1,50 @@
...
+        assert_eq!(ours?, theirs?)
+    }
+
+    // ensure that more than one chunk is present in the new MCAP
+    let num_chunks = mcap::read::LinearReader::new(&mapped)?
+        .filter(|maybe_record| match maybe_record {
+            Ok(mcap::records::Record::Chunk { .. }) => true,
+            _ => false,
+        })
+        .count();
```
mrkline, line 46:
```suggestion
    let num_chunks = mcap::read::LinearReader::new(&mapped)?
        .filter(|maybe_record| matches!(maybe_record, Ok(mcap::records::Record::Chunk { .. })))
        .count();
```

mrkline (commented):
Thanks for making it optional and keeping the existing behavior as the default!

I left it up to users to manually start new chunks with `flush()` because I didn't want to make assumptions about what criteria they want. (Anduril uses several depending on the context - either uncompressed chunk size _and/or_ time since the chunk was started.)

It seems like "max uncompressed size" is a common enough one, and it's built into some of the other writers.

rust/tests/chunks.rs
mrkline: Yeah, that's true of all the tests. I think we were originally testing with larger inputs, and I didn't want to assume they could fit in-memory. But now that `mcap_test_file()` is 2 meg, we should just write to a `Vec<u8>`.

I can open a PR to do that unless one of you wants to beat me to it.

rust/tests/flush.rs
james-rms: Just being explicit, since we're testing flush here.

rust/tests/chunks.rs
james-rms: I'll do that here.

PR #760, opened 2022-12-05
cpp: read messages in deterministic order, yield message offset in MessageView

james-rms:
This PR contains a few changes required to get the `rosbag2_storage_mcap` plugin to actually behave correctly with respect to message ordering and selecting messages by time range.

* Messages with the same timestamp are returned in the order of their position in the file. This is neccessary to have a well-defined return order from ReadMessages, since log timestamps are not required to be unique.
* `MessageView` includes a new "messageOffset" field, informing the reader of the returned messages' offset within the file. This allows readers to tell if a message returned from one iterator is the same message as returned from a different iterator.

## Testing
- [x] unit test added to test new read message order.

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -4,6 +4,9 @@ default: build
 dev-image:
 	docker build -t mcap_cpp_dev -f dev.Dockerfile .
 
+dev-shell: dev-image
+	docker run --rm -v $(CURDIR):/src -it mcap_cpp_dev /bin/bash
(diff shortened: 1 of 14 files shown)


james-rms:
Looking at a better approach now.

jhurliman (approved):
LGTM. It would be helpful to add a test where there are two chunks, A and B, but the time range of B is partially or fully before A.

PR #761, opened 2022-12-06
spec: clarify context of message sequence count

james-rms:
**Public-Facing Changes**
Right now the spec is vague around what the message sequence count is a sequence count of. This PR clarifies that the sequence count is for messages within a given channel.

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -163,10 +163,16 @@ The message encoding and schema must match that of the Channel record correspond
 | Bytes | Name | Type | Description |
 | --- | --- | --- | --- |
 | 2 | channel_id | uint16 | Channel ID |
-| 4 | sequence | uint32 | Optional message sequence count, which should increase by one for each successive message written with the same channel ID. This sequence is expected to roll over from 2<sup>32</sup> to 0. This can be used by readers to detect dropped or out-of-order messages. If not used, all `sequence` values in Message records with the same `channel_id` should be 0. |
+| 4 | sequence | uint32 | Optional message sequence count, which can be used by readers to detect dropped or out-of-order messages. If not used, all `sequence` values in Message records with the same `channel_id` should be 0. |
 | 8 | log_time | Timestamp | Time at which the message was recorded. |
 | 8 | publish_time | Timestamp | Time at which the message was published. If not available, must be set to the log time. |
 | N | data | Bytes | Message data, to be decoded according to the schema of the channel. |
+
+!!! note "Using Message Sequence Counts"
+
+    Message sequence counts are specific to a channel, meaning that each successive Message record in an MCAP with the same channel ID should increase in `sequence` by 1. After this sequence count reaches (2<sup>32</sup> - 1), it is expected to roll over to 0.
+
+    Readers should not expect the sequence count for a given channel to start at 0. This is because the sequence count might originate from a publisher within the pub/sub system that was initialized before the MCAP was created.
 
 ### Chunk (op=0x06)
 
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -163,7 +163,7 @@ The message encoding and schema must match that of the Channel record correspond
 | Bytes | Name | Type | Description |
 | --- | --- | --- | --- |
 | 2 | channel_id | uint16 | Channel ID |
-| 4 | sequence | uint32 | Optional message counter assigned by publisher. If not assigned by publisher, must be recorded by the recorder. |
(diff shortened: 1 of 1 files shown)


docs/specification/README.md
```diff
@@ -163,11 +163,17 @@ The message encoding and schema must match that of the Channel record correspond
...
 | 2 | channel_id | uint16 | Channel ID |
-| 4 | sequence | uint32 | Optional message counter assigned by publisher. If not assigned by publisher, must be recorded by the recorder. |
+| 4 | sequence | uint32 | Optional message sequence count, which can be used by readers to detect dropped or out-of-order messages. If not used, all `sequence` values in Message records with the same `channel_id` should be 0. |
 | 8 | log_time | Timestamp | Time at which the message was recorded. |
 | 8 | publish_time | Timestamp | Time at which the message was published. If not available, must be set to the log time. |
 | N | data | Bytes | Message data, to be decoded according to the schema of the channel. |
 
+!!! note "Using Message Sequence Counts"
+
+    Message sequence counts are specific to a channel, meaning that each successive Message record in an MCAP with the same channel ID should increase in `sequence` by 1. After this sequence count reaches (2<sup>32</sup> - 1), it is expected to roll over to 0.
```
jhurliman, line 173: This seems overly restrictive. If I'm using topic filtering for recording so only a subset of messages on a topic are recorded into an MCAP, the `sequence` would not increment by one if I'm using sequence numbers assigned by my pub/sub middleware. I might have a policy of only recording every tenth message, so I would expect the sequence numbers to go 10, 20, 30, ...

docs/specification/README.md
```diff
@@ -163,11 +163,17 @@ The message encoding and schema must match that of the Channel record correspond
 | Bytes | Name | Type | Description |
 | --- | --- | --- | --- |
 | 2 | channel_id | uint16 | Channel ID |
-| 4 | sequence | uint32 | Optional message counter assigned by publisher. If not assigned by publisher, must be recorded by the recorder. |
+| 4 | sequence | uint32 | Optional message sequence count, which can be used by readers to detect dropped or out-of-order messages. If not used, all `sequence` values in Message records with the same `channel_id` should be 0. |
```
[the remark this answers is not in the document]
james-rms: That's what I'm trying to say, yes. I wanted to leave the opportunity open for some channels to be recorded with sequence counts and some not.

docs/specification/README.md
jhurliman: Those message sequence numbers may be required to make a system function. For example, sequence numbers could be associated with H.264 frame numbers in a side channel topic so `10, 20, 30, ...` values must be present for replay to work.

james-rms:
> or example, sequence numbers could be associated with H.264 frame numbers in a side channel topic so 10, 20, 30, ... values must be present for replay to work.

If we have a field in each message record where its interpretation is specific to the channel schema, then we shouldn't have that field. that data should be serialised with the rest of the message data and the corresponding `Schema` should tell the reader how to interpret it.

> it's not feasible for generic tooling to understand the context of sequence numbers

If this were the case, it'd be a good argument to remove the sequence count field from the spec. either the sequence count has meaning that can be described within the spec, or it has no meaning and shouldn't be there. It's just application-specific data that should be folded into the message and described using a schema.

jhurliman:
> > or example, sequence numbers could be associated with H.264 frame numbers in a side channel topic so 10, 20, 30, ... values must be present for replay to work.
> 
> If we have a field in each message record where its interpretation is specific to the channel schema, then we shouldn't have that field. that data should be serialised with the rest of the message data and the corresponding `Schema` should tell the reader how to interpret it.
> 
> > it's not feasible for generic tooling to understand the context of sequence numbers
> 
> If this were the case, it'd be a good argument to remove the sequence count field from the spec. either the sequence count has meaning that can be described within the spec, or it has no meaning and shouldn't be there. It's just application-specific data that should be folded into the message and described using a schema.

We have a precedent for this type of field, `publish_time`. The spec suggests what `publish_time` could be defined as, but ultimately it is profile-specific, and readers cannot make assertions about data correctness by looking at the `publish_time` field without considering the profile. I think `sequence` currently follows the same pattern. If you know the profile is "ros2" and we update rosbag2 to use that field in a specific way and the well-known profiles addendum is updated to describe this, it would be useful for a ros2-aware reader to surface this.

james-rms:
I still don't understand what the driving need comes from for us to leave this ambiguous and profile-specific. is is H.264 frames?

jhurliman:
> I still don't understand what the driving need comes from for us to leave this ambiguous and profile-specific. is is H.264 frames?

Yes

PR #770, opened 2022-12-07
Add conformance test for reading messages in deterministic order

### LINKED ISSUE #564: Conformance tests for indexed reading in log_time order
We should prove that all implementations implement correct log_time playback

james-rms:
**Public-Facing Changes**
* Removes message index sorting from the Go writer. Previously there was no test fixture with more than one message, which allowed this difference to go unnoticed. 
* Fixed a bug in the Go writer when writing statistics. When writing a message with timestamp 0, the statistics record written by the Go writer would not count that message timestamp when calculating the message start time. This is because 0 is also the sentinel value for MessageStartTime which was previously assumed to mean "no messages".

**Description**
* Adds a special case in the test framework for indexed message readers. This tests the ability of those readers to read messages in log time order.
* Adds an indexed reader test for the C++, Python and Go readers.

Fixes #564


tests/conformance/scripts/run-tests/types.ts
```diff
@@ -0,0 +1,22 @@
...
+export type IndexedReadTestResult = {
+  schemas: SerializableMcapRecord[];
+  channels: SerializableMcapRecord[];
+  messages: SerializableMcapRecord[];
+  statistics: SerializableMcapRecord[];
+};
+
+export type StreamedReadTestResult = {
+  records: SerializableMcapRecord[];
+};
```
jtbandes, line 12: 👍

tests/conformance/scripts/run-tests/runners/stringifyRecords.ts
```diff
@@ -1,7 +1,7 @@
 import { McapTypes } from "@mcap/core";
-import stringify from "json-stringify-pretty-compact";
```
jtbandes, line 2: Maybe this dependency can be removed from package.json?

tests/conformance/scripts/run-tests/runners/stringifyRecords.ts
```diff
@@ -29,19 +29,8 @@ function normalizeRecord(record: McapTypes.TypedMcapRecord) {
...
-        records: normalizedRecords,
-        meta: { variant: { features } },
-      },
-      { replacer },
-    ) + "\n"
-  );
+export function toSerializableMcapRecord(
+  record: McapTypes.TypedMcapRecord,
+): SerializableMcapRecord {
+  return JSON.parse(JSON.stringify(normalizeRecord(record), replacer)) as SerializableMcapRecord;
```
jtbandes, line 35: Maybe don't need to roundtrip stringify/parse here?

tests/conformance/scripts/run-tests/runners/stringifyRecords.ts
```diff
@@ -29,19 +29,8 @@ function normalizeRecord(record: McapTypes.TypedMcapRecord) {
...
-  return (
-    stringify(
-      {
-        records: normalizedRecords,
-        meta: { variant: { features } },
-      },
-      { replacer },
-    ) + "\n"
-  );
+export function toSerializableMcapRecord(
```
jtbandes, line 32: Rename the file to match the new export name?

tests/conformance/scripts/run-tests/types.ts
```diff
@@ -0,0 +1,22 @@
+import { TestFeatures } from "../../variants/types";
+
+export type IndexedReadTestResult = {
+  schemas: SerializableMcapRecord[];
+  channels: SerializableMcapRecord[];
+  messages: SerializableMcapRecord[];
+  statistics: SerializableMcapRecord[];
```
jtbandes, line 7: Do we ever expect to have more than 1 statistics record?

tests/conformance/scripts/run-tests/runners/TestRunner.ts
```diff
@@ -20,9 +41,81 @@ export abstract class ReadTestRunner {
    * Execute the reader test. This may involve calling out to separate process, e.g. with
    * `child_process.exec`.
    * @param filePath A path to a `.mcap` file that should be read.
-   * @returns A JSON-encoded object representing the input file.
+   * @returns an IndexedReadTestResult object.
    */
-  abstract runReadTest(filePath: string, variant: TestVariant): Promise<string>;
+  abstract runReadTest(filePath: string): Promise<IndexedReadTestResult>;
+
+  expectedResult(testCase: TestCase): IndexedReadTestResult {
```
jtbandes, line 48: Seems like this could be `static` or not be a class method at all?

jtbandes (approved):
Awesome

tests/conformance/scripts/run-tests/runners/stringifyRecords.ts
james-rms: the other alternative is to do the recursive replace myself, which i'm not perfectly confident in how to do in JS without hitting a corner case

tests/conformance/scripts/run-tests/types.ts
james-rms: if a writer accidentally puts one there, sure.

tests/conformance/scripts/run-tests/runners/TestRunner.ts
james-rms: static makes sense - i wanted to associate it with the class somehow.

wkalt (approved):

PR #776, opened 2022-12-20
go: add compression level option to writer

james-rms:
**Public-Facing Changes**
* Adds a CompressionLevel field to `mcap.WriterOptions`, which controls the compression level of the zstd or lz4 compressors.
* Plumbs that through the CLI for all subcommands that write MCAP data.

**Description**


wkalt (approved):
looks good - thanks!

PR #777, opened 2022-12-22
rust: set default chunk size to a sane value

james-rms:
**Public-Facing Changes**
Changes the default chunking behavior of the Rust writer to break chunks at a finite value. The current default behavior is a footgun and can lead to a situation where a user loses all of their MCAP data if they never break chunks.

**Description**


james-rms:
cc. @mrkline re: breaking API change

PR #781, opened 2023-01-05
python: add `iter_and_decode_messages` for decoding messages

james-rms:
**Public-Facing Changes**
- Adds `iter_and_decode_messages` to Reader
- makes the `Schema` returned by iter_messages optional, to handle schemaless channels properly

** Description **
Adds a method to the Reader class to iterate through messages and decode them. This is intended as a replacement for the high level reader iterators exposed by support libraries like `mcap-ros1-support`. The advantage of this approach is that it can decode heterogenous MCAPs. Also, it's easier to keep the argument list for `iter_and_decode_messages` matching that of `iter_messages` because it is defined in the same file.

Before:
```python
from mcap_ros1.reader import read_ros1_messages
for view in read_ros1_messages("ros1.mcap"):
    print(type(view.ros_msg))
```

After:
```python
from mcap.well_known import MessageEncoding
from mcap_ros1.decoder import Decoder as ROS1Decoder
from mcap_ros2.decoder import Decoder as ROS2Decoder

with open("heterogenous.mcap", "rb") as f:
    reader = make_reader(f)
    decoders = {
        MessageEncoding.ROS1: ROS1Decoder().decode,
        MessageEncoding.ROS2: ROS2Decoder().decode,
    }
    for _, _, _, decoded_message in reader.iter_and_decode_messages(decoders):
        print(type(decoded_message))
```

### DIFF
--- a/python/mcap/mcap/message_queue.py
+++ b/python/mcap/mcap/message_queue.py
@@ -3,7 +3,9 @@ from typing import Union, Tuple, List, Optional
 
 from .records import Schema, Channel, Message, ChunkIndex
 
-QueueItem = Union[ChunkIndex, Tuple[Tuple[Schema, Channel, Message], int, int]]
+QueueItem = Union[
+    ChunkIndex, Tuple[Tuple[Optional[Schema], Channel, Message], int, int]
+]
 
 
 class _Orderable:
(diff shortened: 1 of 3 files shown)


defunctzombie (approved):
Approving this since I think it is moving in the right direction. As discussed in Slack, my personal preference/intuition is that the decoders (deserializers?) are part of the reader setup rather than the message iteration. As a user of a reader, I would initialize the reader with various settings deserializers and then proceed to use the reader one or more times to read a set of topics or ranges. I wouldn't want to pass along the decoders at every read site.

james-rms:
Going with the alternative approach suggested by @defunctzombie in a separate PR.

PR #787, opened 2023-01-12
CLI: search optimistically for message definitions

james-rms:
**Public-Facing Changes**
Replaces the logic in `mcap convert` for ROS 2 files that looks for schema definitions. Previously we would look for message definitions using the ament resource index, but this is a fragile mechanism especially for workspaces with message definitions in nested subdirectories. Several users have had trouble using this, and debugging why their install space isn't exactly correct is a slog. Instead, `mcap convert` searches for a path with the form `<packageName>/msg/<MsgName>.msg` anywhere in one of the directories specified with `ament-prefix-path`.

### DIFF
--- a/go/ros/ros2db3_to_mcap.go
+++ b/go/ros/ros2db3_to_mcap.go
@@ -6,8 +6,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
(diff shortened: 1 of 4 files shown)


jtbandes:
Is this an alternative to #789 or how are they related? Seems like #789 is better?

james-rms:
This was a misclick, I mistakenly merged. I'll merge #789 over it before the next release.

PR #797, opened 2023-01-25
Rust: Add separate methods for flushing writer and terminating chunks to public API 

dmweis:
**Public-Facing Changes**
No API changes. But changes to behaviour

**Description**
I noticed that when calling flush from the public API it will terminate the chunk and then flush the writer. 
But when chunks are triggered automatically based on size the writer is not flushed.

I am not sure if this was intentional but this could result in data not being written to disk for a while. And the only way a user can trigger an explicit flush is to also terminate a chunk which is not always desirable.

Maybe having a separate `.terminate_chunk()` and `.flush()` on the public API might be better? But that would be a breaking change


b-camacho:
Interesting point! 

> I am not sure if this was intentional but this could result in data not being written to disk for a while
This was an intentional choice. The rationale was the OS knows the optimal time to flush, and there's no need to force extra flushes in almost all cases.

The only exceptions I can think of are:
- environment where the host might power down unexpectedly, but I trust devs under those conditions have to think about frequently flushing their data to involatile storage anyway
- the client crashing, but 🤷

Completely agree that 
> And the only way a user can trigger an explicit flush is to also terminate a chunk
is undesirable.

What do you think about:
- leaving `.flush()` as is, and
- adding a method that flushes but doesn't terminate the chunk, and
- adding that `.terminate_chunk()` method you mentioned?

dmweis:
> leaving .flush() as is, and
adding a method that flushes but doesn't terminate the chunk, and
adding that .terminate_chunk() method you mentioned?

I like this idea! I will update this PR

mrkline:
Hey @dmweis! Sorry for a delayed response - I've been traveling for a friend's wedding.

> I am not sure if this was intentional but this could result in data not being written to disk for a while. And the only way a user can trigger an explicit flush is to also terminate a chunk which is not always desirable.

My original goal was to have some function that would sync everything written so far, such that:
1. The MCAP file was in a good state - i.e., a streaming reader could read valid `Message`s up to this point
2. The MCAP file was written to disk  (inasmuch as disk drivers aren't lying to us :sweat_smile: )
Maybe calling that `flush()` was a bad idea since that usually just implies the latter, but I didn't imagine that being super useful without the former (i.e., also finishing the current chunk).

In hindsight it's probably a good idea to make `finish_chunk()` public (there's plenty of times when you might want to start a new chunk while letting the OS decide when to flush the file to disk, like @b-camacho suggested). But forgive my lack of imagination - what are some scenarios where flushing a chunk in the middle of a compression scheme would help?

dmweis:
Hey @mrkline! No worries! Thanks for the reply.

So I had to pause and rethink what I was trying to do. I originally just wanted to flush on automated chunk finishes so that I could have some confidence about data being written to disk in time.
But as suggested above flushing at automated chunking is not necessarily desirable.

My main goal is that I can offer some guarantee about data being written to disk in time. Say in case our robot was to crash (in either sense of the word). I want to guarantee that we do not miss more than last N seconds of data, or last N messages.

> what are some scenarios where flushing a chunk in the middle of a compression stream would help?

So to my understanding even a partially compressed chunk is still readable? I have limited understanding of `zstd` is that as long as full frame is written it can be read? Is that not correct?

This way I could call flush every N seconds from the client side and be confident that we are unlikely to lose too much data in case of an issue

defunctzombie:
@dmweis From your last comment it sounds like maybe this PR is not the direction you want to go? If it is no longer relevant lets close it. If you do still want to pursue this - is there additional feedback you need to make the requested changes?

dmweis:
> @dmweis From your last comment it sounds like maybe this PR is not the direction you want to go? If it is no longer relevant lets close it. If you do still want to pursue this - is there additional feedback you need to make the requested changes?

Hey, thanks for checking in. We ended up going with a different solution. Thanks for your help here!

PR #798, opened 2023-01-26
Update mcap dependencies in CLI tool

wkalt:
This should fix an installation error that occurs when installing via

    go install github.com/foxglove/mcap/go/cli/mcap@latest

When running the install locally the build succeeds; I think usage of go workspaces is overriding the dependency specified in the go.mod file. Need to understand the mechanics of that better to propose a long-term fix to keep these in sync. For now this just updates the deps which I expect will resolve the immediate problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,7 +208,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: 3.7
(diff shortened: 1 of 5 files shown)


jhurliman (approved):

jhurliman:
> Current Conan version (1.53.0) does not satisfy the defined one (>=1.54.0).

:-\

PR #801, opened 2023-01-27
Spec: prohibit zero opcodes

wkalt:
This makes an explicit prohibition on zero opcodes. Zero opcodes were previously not designated as reserved for either future mcap development or application extensions - nothing was said about them. However some special handling is warranted because readers should not ignore them, or else empty file content (interpreted as zero opcodes followed by zero-length messages) may parse without an error.

This changes the behavior from undefined to defined as an error.

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -92,7 +92,7 @@ The summary offset section aids random access reading.
 
 ## Records
 
-MCAP files may contain a variety of records. Records are identified by a single-byte **opcode**. Record opcodes in the range 0x01-0x7F are reserved for future MCAP format usage. 0x80-0xFF are reserved for application extensions and user proposals.
+MCAP files may contain a variety of records. Records are identified by a single-byte **opcode**. Record opcodes in the range 0x01-0x7F are reserved for future MCAP format usage. 0x80-0xFF are reserved for application extensions and user proposals. The opcode 0x00 is an error.
 
 All MCAP records are serialized as follows:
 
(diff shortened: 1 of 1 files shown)


defunctzombie:
`The opcode 0x00 is an error.` This language makes it sound like this is an opcode to write an error record. I'd expect this instead to read as `Opcode 0x00 is unused`?

docs/specification/README.md
```diff
@@ -92,7 +92,7 @@ The summary offset section aids random access reading.
 
 ## Records
 
-MCAP files may contain a variety of records. Records are identified by a single-byte **opcode**. Record opcodes in the range 0x01-0x7F are reserved for future MCAP format usage. 0x80-0xFF are reserved for application extensions and user proposals.
+MCAP files may contain a variety of records. Records are identified by a single-byte **opcode**. Record opcodes in the range 0x01-0x7F are reserved for future MCAP format usage. 0x80-0xFF are reserved for application extensions and user proposals. The opcode 0x00 is an error.
```
james-rms, line 95: Possibly clearer about what kind of error it is:

```suggestion
MCAP files may contain a variety of records. Records are identified by a single-byte **opcode**. Record opcodes in the range 0x01-0x7F are reserved for future MCAP format usage. 0x80-0xFF are reserved for application extensions and user proposals. 0x00 is not a valid opcode.
```

james-rms (approved):

docs/specification/README.md
defunctzombie: 👍

jhurliman (approved):

jtbandes (approved):

PR #804, opened 2023-01-31
Configureable Zstd multi-threading

AlexanderEkdahl:
By default the Zstd compression would use the same number of threads as the number of physical cores on the system.

This causes issues in systems with many cores and/or multiple MCAP writers.

This adds a new `zstd_worker_threads` configuration options which allows the users the ability to override this default.

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -108,6 +108,8 @@ pub struct WriteOptions {
     compression: Option<Compression>,
     profile: String,
     chunk_size: Option<u64>,
+    #[cfg(feature = "zstd")]
+    zstd_worker_threads: u32,
 }
 
 impl Default for WriteOptions {
@@ -115,6 +117,8 @@ impl Default for WriteOptions {
         Self {
             #[cfg(feature = "zstd")]
             compression: Some(Compression::Zstd),
+            #[cfg(feature = "zstd")]
+            zstd_worker_threads: num_cpus::get_physical() as u32,
             #[cfg(not(feature = "zstd"))]
             compression: None,
             profile: String::new(),
@@ -132,6 +136,24 @@ impl WriteOptions {
     pub fn compression(self, compression: Option<Compression>) -> Self {
         Self {
             compression,
+            ..self
+        }
+    }
+
+    #[cfg(feature = "zstd")]
+    /// Override the number of threads used for zstd compression.
+    /// By default it will use up the number of physical cores.
+    /// 
+    /// * If `zstd_worker_threads == 0`, then multithreaded compression
+    ///   will be disabled.
+    /// * If `zstd_worker_threads >= 1`, then compression will be done 
+    ///   in separate threads.
+    ///
+    /// So even `n_workers = 1` may increase performance by separating
+    /// IO and compression.
+    pub fn zstd_worker_threads(self, zstd_worker_threads: u32) -> Self {
+        Self {
+            zstd_worker_threads,
             ..self
         }
     }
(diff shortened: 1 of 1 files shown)


defunctzombie:
Is it a good default to have the worker threads be the same as number of CPU cores? In my past experience with robot software - we would be very careful about the allocation of CPU threads across processes. Maybe a better default is to not enable threaded zstd and allow the user to enable it by configuring the number of threads (as this PR does).

@mrkline @jhurliman

AlexanderEkdahl:
Yes, I would also be in favour of not checking the number of cores and not use multi-threading by default. This would change the API though.

jhurliman:
I disagree with disabling multithreading by default, but this should definitely be configurable either way. I'm not familiar enough with Rust to review this PR but the overall idea seems right to me.

> Is it a good default to have the worker threads be the same as number of CPU cores?

Yes, I think this is a reasonable out-of-the-box default (use the CPU resources available, don't automatically throttle). If you want to leave unused CPU cycles, lower system thread count, reduce RAM usage, etc. those are situation-specific concerns that can be managed with ulimits, cgroups, containers, or reducing the configured thread count as this PR allows.

rust/src/write.rs
```diff
@@ -108,13 +108,17 @@ pub struct WriteOptions {
     compression: Option<Compression>,
     profile: String,
     chunk_size: Option<u64>,
+    #[cfg(feature = "zstd")]
```
defunctzombie, line 111: Did the library already use this feature configuration option or is it being introduced here in this PR?

rust/src/write.rs
```diff
@@ -406,7 +428,12 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
             WriteMode::Raw(w) => {
                 // It's chunkin time.
                 self.stats.chunk_count += 1;
-                WriteMode::Chunk(ChunkWriter::new(w, self.options.compression)?)
+                WriteMode::Chunk(ChunkWriter::new(
+                    w,
+                    self.options.compression,
+                    #[cfg(feature = "zstd")]
+                    self.options.zstd_worker_threads,
```
defunctzombie, line 435: Is there a nice way to make this `compression_options` or some way to allow for configuring options that might come up for other compression schemes? Maybe this is the way this is done in rust and we would add more feature checks while adding more arguments to the function? To me it feels unnatural if I use this library with the `zstd` feature and the `lz4` feature (for example) and that means my list of arguments grows unbounded here?

AlexanderEkdahl: Yes, I agree that it is unnatural and I was mostly following the existing paradigm. If we're not too concerned about API breakage (it's not `1.0` after all) I'll put up a PR for how I think the configuration should ideally look like.

defunctzombie: I personally would like to see that to help us shape the decision here.

Maybe do it as an additional commit to this PR so we can easily peel it back if the folks more familiar with the rust code and direction of this library feel differently.

I think we are still in the early days of this library so it is reasonable to iterate like this.

james-rms: I would like to request that the number of worker threads be added as a field to `Compression::Zstd`, for example:

```rust

WriteOptions::new().compression(Compression::Zstd {num_worker_threads: 2})
```

defunctzombie:
Seems this PR went stale. The last comment I see is from @james-rms asking for an change. I believe once that is in then we are good to merge this? I've seen no other change requests to indicate otherwise.

jtbandes:
I'm going to close this PR since it's been several months since the last activity. If you are still interested in pursuing it, feel free to reply and we can reopen and merge once the prior review comments are addressed!

PR #814, opened 2023-02-16
Add image_example.py, use foxglove-schemas-protobuf package

jhurliman:
**Public-Facing Changes**
- [Python] New `image_example.py` for writing RawImage, CameraCalibration, and ImageAnnotation
- [Python] Use `foxglove-schemas-protobuf` package instead of in-repo proto/pb2 files


jhurliman:
```bash
pipenv run black --check --diff --color mcap
Traceback (most recent call last):
  File "/home/runner/.local/share/virtualenvs/python--WAQ1xVa/bin/black", line 5, in <module>
    from black import patched_main
  File "src/black/__init__.py", line 45, in <module>
  File "/home/runner/.local/share/virtualenvs/python--WAQ1xVa/lib/python3.10/site-packages/black/files.py", line 34, in <module>
    import tomli as tomllib
ModuleNotFoundError: No module named 'tomli'
make: *** [Makefile:42: lint] Error 1
```

How did this happen

jtbandes:
It looks like the `tomli` issue was fixed in https://github.com/psf/black/issues/3447  Maybe we can upgrade black to avoid this?

jhurliman:
@jtbandes this PR already upgrades black to the latest version. It turned out to be a setuptools issued, which was fixed by upgrading the pipenv version we use in CI.

PR #818, opened 2023-02-16
[Python] Add enable_crcs and enable_data_crcs Writer options. Expose chunk_size, compression, and enable_crcs to support writers

jhurliman:
**Public-Facing Changes**
- [Python] Add Writer class constructor args `enable_crcs=True`, `enable_data_crcs=False`
- [Python] Expose `chunk_size`, `compression`, and `enable_crcs` args to Writer constructors for mcap-ros1-support, 
mcap-ros2-support, and mcap-protobuf-support
- [Python] Fixed mcap-ros2-support to throw McapROS2WriteError when schema parsing fails

**Description**
Also includes misc typing improvements

### DIFF
--- a/python/mcap-ros2-support/mcap_ros2/writer.py
+++ b/python/mcap-ros2-support/mcap_ros2/writer.py
@@ -6,7 +6,7 @@ import mcap
 from mcap.exceptions import McapError
 from mcap.records import Schema
 from mcap.well_known import SchemaEncoding
-from mcap.writer import Writer as McapWriter
+from mcap.writer import CompressionType, Writer as McapWriter
 
 from . import __version__
 from ._dynamic import EncoderFunction, serialize_dynamic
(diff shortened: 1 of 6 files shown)


james-rms (approved):

python/mcap-ros2-support/mcap_ros2/writer.py
```diff
@@ -76,9 +87,10 @@ def write_message(
             type_dict = serialize_dynamic(  # type: ignore
                 schema.name, schema.data.decode()
             )
-            encoder = type_dict[schema.name]
-            if encoder is None:
+            # Check if schema.name is in type_dict
```
james-rms, line 90: fyi, you can change this line to:
```python
encoder = type_dict.get(schema.name, None)
```
and leave the rest the same, and have the correct behaviour while only searching the dictionary once. In this context it'd have negligible performance impact, so 🤷 but I figured i'd say it anyway.

PR #819, opened 2023-02-17
[cli] Add `mcap decompress` command

jhurliman:
**Public-Facing Changes**
- [cli] `mcap decompress` command, alias for `mcap compress --compression "none"`

### DIFF
--- a/go/cli/mcap/cmd/filter.go
+++ b/go/cli/mcap/cmd/filter.go
@@ -64,6 +64,7 @@ func buildFilterOptions(flags filterFlags) (*filterOpts, error) {
 	case "lz4":
 		opts.compressionFormat = mcap.CompressionLZ4
 	case "none":
+	case "":
 		opts.compressionFormat = mcap.CompressionNone
 	default:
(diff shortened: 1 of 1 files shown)


go/cli/mcap/cmd/filter.go
```diff
@@ -64,6 +64,7 @@ func buildFilterOptions(flags filterFlags) (*filterOpts, error) {
 	case "lz4":
 		opts.compressionFormat = mcap.CompressionLZ4
 	case "none":
+	case "":
```
james-rms, line 67: why this change? i figure it's clearer to require a non-empty string if `--compression` is supplied.

james-rms (commented):
I don't really see the use-case, but if you have a use for it 👍

go/cli/mcap/cmd/filter.go
jhurliman: I tried to run the command `mcap compress --compression "" input.mcap -o output.mcap` and it failed, telling me to use `"none"` instead. Not a terrible outcome (it told me what to do), but why not just do what I asked and use the `""` compression mode as defined in the MCAP spec?

PR #820, opened 2023-02-17
[cli] `mcap doctor --verbose`, default to silence

jhurliman:
**Public-Facing Changes**
- [cli] `mcap doctor` now defaults to silent for conforming MCAP files. The previous behavior can be restored with `--verbose`/`-v`

**Description**
Following http://www.linfo.org/rule_of_silence.html as the default behavior for the `mcap doctor` command. The previous behavior is still available through a new verbose flag.

### DIFF
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -18,6 +18,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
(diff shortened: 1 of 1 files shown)


james-rms (approved):

defunctzombie:
👍 Personally I've often found that rule to lead to a confusing experience for tools like this one. I've no idea if it actually did anything.

PR #821, opened 2023-02-17
Update TypeScript dependencies

jhurliman:
**Public-Facing Changes**
- None

**Description**
Updates TypeScript dependencies to latest versions


cspell.config.yaml
```diff
@@ -29,6 +29,7 @@ ignorePaths:
   - mkdocs.yml
   - fathom.html
   - _vendor
+  - go
```
jtbandes, line 32: Why is the go sdk exempt from spell checking?

jhurliman: Go programs are apparently not written in English 🤷🏼

PR #823, opened 2023-02-20
add note recommending against mixed chunking

### LINKED ISSUE #792: Ambiguous wording in the MCAP File Format Specification
The following wording in the specification of a Chunk is a bit ambiguous -

> All messages in the chunk must reference channels recorded earlier in the file (in a previous chunk or earlier in the current chunk).

It may be interpreted as "all channels must be in a chunk (previous or current)". However, as far as I'm aware, this is not necessary. Channels can be logged standalone outside a chunk, in the Data section. Please correct me if I'm wrong.

james-rms:
**Public-Facing Changes**
* The MCAP specification now recommends against MCAPs from mixing chunked-and un-chunked writing in the same file.
* `mcap doctor` now reports a warning for MCAPs containing Message records in the data section with Chunk Indexes in the summary section.

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -59,7 +59,6 @@ The following records are allowed to appear in the data section:
 - [Chunk](#chunk-op0x06)
 - [Message Index](#message-index-op0x07)
 - [Metadata](#metadata-op0x0C)
-
 - [Data End](#data-end-op0x0F)
 
 The last record in the data section MUST be the [Data End](#data-end-op0x0F) record.
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -65,7 +65,7 @@ The last record in the data section MUST be the [Data End](#data-end-op0x0F) rec
 
 #### Use of chunk records
 
-MCAP files can have Schema, Channel, and Message records written directly to the data section, or they can be written into Chunk records to facilitate indexing and compression. Schema, Channel, and Message records must either all be written inside chunks, or all outside of chunks. Reader implementations may assume that if any Chunk records exist, all Schema, Channel, and Message records can be found inside Chunk records.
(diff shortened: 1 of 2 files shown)


docs/specification/README.md
```diff
@@ -59,10 +59,15 @@ The following records are allowed to appear in the data section:
 - [Chunk](#chunk-op0x06)
 - [Message Index](#message-index-op0x07)
 - [Metadata](#metadata-op0x0C)
+
```
jhurliman, line 62: Typo?

docs/specification/README.md
```diff
@@ -59,10 +59,15 @@ The following records are allowed to appear in the data section:
...
 - [Data End](#data-end-op0x0F)
 
 The last record in the data section MUST be the [Data End](#data-end-op0x0F) record.
 
+#### Use of chunk records
+
+MCAP files can have Schema, Channel, and Message records written directly to the data section, or they can be written into Chunk records to facilitate indexing and compression. Schema, Channel, and Message records must either all be written inside chunks, or all outside of chunks. A data section containing Chunk records may not contain any Schema, Channel, or Message records and vice-versa.
```
jhurliman, line 69: > A data section containing Chunk records may not contain any Schema, Channel, or Message records and vice-versa.

They're still in the data section, just inside a chunk. I wonder if there's another way to word this.

james-rms: Reworded, might be slightly better?

defunctzombie:
Is this a backwards compatible change?. We can make doctor do this check and produce warnings since doctor can go beyond the spec and guide you towards better mcap files but I don't think we can change the spec in this way.

What is the problem with combining chunked and un-chunked? Is it that readers will typically skip any message records not in chunks when reading in chunked mode?

james-rms:
It's a backwards-incompatible change. The core problem is that if there are MCAP files with chunks and chunk indexes and messages outside of those chunks which are not covered by the index, then readers have no way to read using the index without possibly skipping some messages.

If we wanted to make this change in a compatible way, i'd change `must` to `should`, and add a warning about what will happen if this recommendation is not followed. I'm OK with doing that.

defunctzombie (approved):

PR #826, opened 2023-02-21
Update dependencies & remove go.work

wkalt:
Specify that installing with go install is not supported.

### DIFF
--- a/go/cli/mcap/README.md
+++ b/go/cli/mcap/README.md
@@ -8,7 +8,10 @@ details.
 ### Installing:
 
 Either install from [releases
-binaries](https://github.com/foxglove/mcap/releases), from [Homebrew](https://formulae.brew.sh/formula/mcap#default), or by using go.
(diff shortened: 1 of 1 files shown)


james-rms:
Just noting that this PR makes MCAP library changes a four-step serial process:

1. PR changing go/mcap
2. Tagged release of go/mcap/v1.x.y
3. PR changing go/cli/
4. Tagged release of releases/mcap-cli/vx.y.z

As opposed to now which is:
1. PR changing go/mcap
2. Tagged release of go/mcap/v1.x.y
3. Tagged release of releases/mcap-cli-vx.y.z

where 2 and 3 can happen in parallel.

Our alternative is to insert some code to make `go install` fail with an error message saying to go grab binaries from github, or to clone the repo and build from the workspace.

wkalt:
@james-rms see update - we are now just saying it's not supported in the README.

james-rms (approved):

PR #831, opened 2023-02-23
[TypeScript] Add `noMagicPrefix` option to McapStreamReader to allow parsing to start past MCAP_MAGIC

jhurliman:
**Public-Facing Changes**
- [TypeScript] `McapStreamReader` accepts a `noMagicPrefix` option that allows parsing to start past the beginning of an MCAP file

### DIFF
--- a/typescript/core/src/McapStreamReader.ts
+++ b/typescript/core/src/McapStreamReader.ts
@@ -128,14 +128,12 @@ export default class McapStreamReader {
   }
 
   private *read(): Generator<TypedMcapRecord | undefined, TypedMcapRecord | undefined, void> {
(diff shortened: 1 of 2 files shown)


typescript/core/src/McapStreamReader.ts
```diff
@@ -119,11 +129,13 @@ export default class McapStreamReader {
 
   private *read(): Generator<TypedMcapRecord | undefined, TypedMcapRecord | undefined, void> {
     {
```
jtbandes, line 131: extra braces for scope are no longer necessary

jtbandes (approved):

PR #833, opened 2023-02-24
Update ROS library to support bounded fields in mcap convert

wkalt:


wkalt:
This adds support for the cases mentioned in https://docs.ros.org/en/humble/Concepts/About-ROS-Interfaces.html. If there are any cases not mentioned in that, they may be mishandled - it's not clear to me if that document is the official spec or not.

james-rms (approved):

PR #840, opened 2023-02-28
Remove explicit dependency on mcap libs in CLI

wkalt:
Perhaps if we do this, go.work will be respected by go install.


jtbandes (approved):

wkalt:
I determined that this does not work. We have since addressed the issue by documenting go install not supported.

PR #842, opened 2023-03-01
Add README and FileHandleReadable to @mcap/support

jhurliman:
**Public-Facing Changes**
- [TypeScript] Adds `FileHandleReadable` to @mcap/support and README docs

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -44,6 +44,7 @@ words:
   - conans
   - crc
   - crcs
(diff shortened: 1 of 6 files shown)


defunctzombie:
Thanks for the docs notes. I've reverted the @mcap/support PR since that got no review and that's not our process for releasing things. We will make the mcap support stuff available but after it has received a review and we are happy with the structure and documentation.

PR #853, opened 2023-03-13
Add SkipMagic option to go writer

wkalt:
Adds a "SkipMagic" option to the go writer. When provided, the leading magic bytes will not be written. This may be used for appending to an existing MCAP file.

### DIFF
--- a/go/mcap/writer.go
+++ b/go/mcap/writer.go
@@ -777,6 +777,10 @@ type WriterOptions struct {
 	// SkipSummaryOffsets skips summary offset records.
 	SkipSummaryOffsets bool
 
+	// SkipMagic skips the magic bytes at the start of the file. This may be
+	// useful for resuming writes midway through a file.
+	SkipMagic bool
+
 	// OverrideLibrary causes the default header library to be overridden, not
 	// appended to.
 	OverrideLibrary bool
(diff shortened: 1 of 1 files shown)


go/mcap/writer.go
```diff
@@ -817,8 +821,10 @@ func encoderLevelFromZstd(level CompressionLevel) zstd.EncoderLevel {
 // NewWriter returns a new MCAP writer.
 func NewWriter(w io.Writer, opts *WriterOptions) (*Writer, error) {
 	writer := newWriteSizer(w, opts.IncludeCRC)
-	if _, err := writer.Write(Magic); err != nil {
-		return nil, err
+	if !opts.SkipMagic {
```
james-rms, line 824: do we want to enforce that includeCRC is false here?

james-rms (commented):
Overall seems easy to write malformed MCAPs using this - we can't really be sure that any of the summary section would be correct if we resumed writing after writing a few messages into the data section.

What's the driving use case for this feature?

wkalt:
@james-rms I want this for an MCAP writer that can resume an interrupted download. I am using it in conjunction with indexes/summaries disabled, and after I open it I need to run messages through channel ID translation - so it isn't trivial to use in general (but it's off by default).

The situation is basically,
1. Download starts as a regular copy from remote source to local destination
2. On download interrupt, the output so far is parsed for chunk offsets and channel/schema IDs.
3. Scan chunks in reverse order to identify a suitable message timestamp to use to form the next request
4. Open a writer at location of chunk containing that message (*here is the problem - the magic is written to the middle of your file*)
5. Write the contents of the split chunk, and resume the reads on the new writer with the new request.

The constraint is that the backend server is not a file server; it doesn't understand range requests.

If you'd prefer though, I think I can handle this in my code too, by just seeking backward a few bytes after the writer gets initialized.

wkalt:
Closing this after discussion - I think I have a better way to go without this feature.

PR #854, opened 2023-03-14
Speed up the info subcommand

wkalt:
Reads full summary section into memory before parsing, rather than pulling it into memory over multiple calls to Read(). Will use more memory than before (I think memory usage for the associated data will double during the function call) but outside the function call should be same as before.

I think the users who will benefit from this will be,
* Users with huge numbers of chunks
* Users with slow and/or remote disks

### DIFF
--- a/go/mcap/indexed_message_iterator.go
+++ b/go/mcap/indexed_message_iterator.go
@@ -1,6 +1,7 @@
 package mcap
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/binary"
 	"errors"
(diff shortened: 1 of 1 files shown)


james-rms (approved):
One idea: could pick a threshold where the summary is too large to read all at once, and stick with the old strategy in that case.

go/mcap/indexed_message_iterator.go
```diff
@@ -65,9 +66,28 @@ func (it *indexedMessageIterator) parseSummarySection() error {
 	if err != nil {
 		return fmt.Errorf("failed to seek to summary start")
 	}
+
+	summarySection := make([]byte, footer.SummaryOffsetStart-footer.SummaryStart)
```
james-rms, line 70: conformance tests caught this - SummaryOffsetStart might be 0 if there are no summary offsets, but there is a summary.

wkalt: thanks - updated

PR #855, opened 2023-03-14
Update example for mcap_ros2.reader

idrilirdi:
### Public-Facing Changes

None

### Description

The example was outdated as commit 8ee1d0f removed the 'topic' member for the McapROS2Message class. It is now under 'channel.topic'


jtbandes (approved):
Thanks!

Issue #847, opened 2023-03-13
bug(cli): `mcap cat` Sigsev on schemaless encoding


jon-chuang:
**Description**
The following error is encountered:
```
running /tmp/tmpbiz8i177
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xb77ca5]

goroutine 1 [running]:
github.com/foxglove/mcap/go/cli/mcap/cmd.printMessages({0xc00034fc40?, 0xc000355480?}, {0xf5fee0, 0xc000010018}, {0xf5cd80, 0xc00050c1c0}, 0x1)
	/home/jonch/Desktop/Programming/systems/mcap/go/cli/mcap/cmd/cat.go:113 +0x4e5
github.com/foxglove/mcap/go/cli/mcap/cmd.glob..func5.1(0x75?, {0x7fe3f42e82f8?, 0xc000010af8})
	/home/jonch/Desktop/Programming/systems/mcap/go/cli/mcap/cmd/cat.go:236 +0x1c5
github.com/foxglove/mcap/go/cli/mcap/utils.WithReader({0xf64f50, 0xc0000360d8}, {0x7ffe00805e75?, 0x0?}, 0xc0002dfd90)
	/home/jonch/Desktop/Programming/systems/mcap/go/cli/mcap/utils/utils.go:104 +0x431
github.com/foxglove/mcap/go/cli/mcap/cmd.glob..func5(0x145c420?, {0xc00009b980, 0x1, 0x2?})
	/home/jonch/Desktop/Programming/systems/mcap/go/cli/mcap/cmd/cat.go:226 +0x312
github.com/spf13/cobra.(*Command).execute(0x145c420, {0xc00009b960, 0x2, 0x2})
	/home/jonch/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:876 +0x67b
github.com/spf13/cobra.(*Command).ExecuteC(0x145b520)
	/home/jonch/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:990 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
	/home/jonch/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:918
github.com/foxglove/mcap/go/cli/mcap/cmd.Execute()
	/home/jonch/Desktop/Programming/systems/mcap/go/cli/mcap/cmd/root.go:21 +0x25
[...]

PR #857, opened 2023-03-15
fix(cli): sigsev on null schema

jon-chuang:
### Description
Previously, we did not handle schema being a nullptr

Fixes: https://github.com/foxglove/mcap/issues/847

### DIFF
--- a/go/cli/mcap/README.md
+++ b/go/cli/mcap/README.md
@@ -32,7 +32,7 @@ To install using [Homebrew](https://brew.sh) on macOS or Linux, run:
 
     brew install mcap
 
-#### Using go (not recommended)
+#### Using go
 
 To install from the latest commit, use
 
--- a/go/cli/mcap/README.md
+++ b/go/cli/mcap/README.md
@@ -38,6 +38,8 @@ To install from the latest commit, use
 
     go install github.com/foxglove/mcap/go/cli/mcap@latest
 
+Ensure that the go installation directory is in your path (e.g. `$HOME/go/bin` by default on Ubuntu).
+
 ### Examples:
 
 #### Bag to mcap conversion
--- a/go/cli/mcap/README.md
+++ b/go/cli/mcap/README.md
@@ -32,7 +32,7 @@ To install using [Homebrew](https://brew.sh) on macOS or Linux, run:
 
     brew install mcap
 
-#### Using go
+#### Using go (not recommended)
 
 To install from the latest commit, use
 
--- a/go/cli/mcap/cmd/cat.go
+++ b/go/cli/mcap/cmd/cat.go
@@ -249,5 +249,5 @@ func init() {
 	catCmd.PersistentFlags().Int64VarP(&catEnd, "end-secs", "", math.MaxInt64, "end time")
 	catCmd.PersistentFlags().StringVarP(&catTopics, "topics", "", "", "comma-separated list of topics")
 	catCmd.PersistentFlags().BoolVarP(&catFormatJSON, "json", "", false,
-		`print messages as JSON. Supported message encodings: ros1msg, protobuf, and JSON.`)
+		`print messages as JSON. Supported message encodings: ros1, protobuf, and json.`)
 }
--- a/go/cli/mcap/cmd/cat.go
+++ b/go/cli/mcap/cmd/cat.go
@@ -169,7 +169,7 @@ func printMessages(
 					return fmt.Errorf("failed to write message bytes: %w", err)
 				}
 			default:
-				return fmt.Errorf("JSON output only supported for ros1msg, protobuf, and json schemas. Found: %s", schema.Encoding)
+				return fmt.Errorf("JSON output only supported for ros1msg, protobuf, and jsonschema schemas. Found: %s", schema.Encoding)
 			}
 		}
 		target.Topic = channel.Topic
@@ -248,5 +248,7 @@ func init() {
 	catCmd.PersistentFlags().Int64VarP(&catStart, "start-secs", "", 0, "start time")
 	catCmd.PersistentFlags().Int64VarP(&catEnd, "end-secs", "", math.MaxInt64, "end time")
 	catCmd.PersistentFlags().StringVarP(&catTopics, "topics", "", "", "comma-separated list of topics")
-	catCmd.PersistentFlags().BoolVarP(&catFormatJSON, "json", "", false, "print messages as JSON")
+	catCmd.PersistentFlags().BoolVarP(&catFormatJSON, "json", "", false,
+		`print messages as JSON. Supported schema encodings: ros1msg, protobuf, and jsonschema. 
+		Supported schema-less encodings: json.`)
 }
--- a/go/cli/mcap/cmd/cat.go
+++ b/go/cli/mcap/cmd/cat.go
@@ -16,7 +16,6 @@ import (
 	"github.com/foxglove/mcap/go/cli/mcap/utils/ros"
 	"github.com/foxglove/mcap/go/mcap"
 	"github.com/foxglove/mcap/go/mcap/readopts"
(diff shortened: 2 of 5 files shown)


go/cli/mcap/README.md
```diff
@@ -38,6 +38,8 @@ To install from the latest commit, use
 
     go install github.com/foxglove/mcap/go/cli/mcap@latest
 
+Ensure that the go installation directory is in your path (e.g. `$HOME/go/bin` by default on Ubuntu).
```
jtbandes, line 41: Actually we should probably remove this section (see https://github.com/foxglove/mcap/pull/826/files) -- cc @wkalt

go/cli/mcap/cmd/cat.go
```diff
@@ -109,57 +109,68 @@ func printMessages(
...
+					return fmt.Errorf("failed to write message bytes: %w", err)
 				}
-				transcoders[channel.SchemaID] = transcoder
-			}
-			msgReader.Reset(message.Data)
-			err = transcoder.Transcode(msg, msgReader)
-			if err != nil {
-				return fmt.Errorf("failed to transcode %s record on %s: %w", schema.Name, channel.Topic, err)
+			default:
+				return fmt.Errorf("For schemaless encodings, JSON output only supports: JSON. Found: %s", channel.MessageEncoding)
```
jtbandes, line 119:
```suggestion
				return fmt.Errorf("For schemaless channels, JSON output is only supported with 'json' message encoding. Found: %s", channel.MessageEncoding)
```

go/cli/mcap/cmd/cat.go
```diff
@@ -109,57 +109,68 @@ func printMessages(
...
-				return fmt.Errorf("failed to write message bytes: %w", err)
+				if _, err = msg.Write(bytes); err != nil {
+					return fmt.Errorf("failed to write message bytes: %w", err)
+				}
+			case "jsonschema":
+				if _, err = msg.Write(message.Data); err != nil {
+					return fmt.Errorf("failed to write message bytes: %w", err)
+				}
+			default:
+				return fmt.Errorf("JSON output only supported for the following encodings - with schema: ros1msg, protobuf, and JSON; schemaless: JSON. Found: %s", schema.Encoding)
```
jtbandes, line 172: I don't think we need to mention schemaless channels when the channel already has a schema
```suggestion
				return fmt.Errorf("JSON output only supported for ros1msg, protobuf, and json schemas. Found: %s", schema.Encoding)
```

jtbandes (commented):
Can we add some explanation of the supported encodings to the command help as well?

go/cli/mcap/README.md
jon-chuang: Hmm we already state that it is not recommended, but I think we should leave it in as it's helpful for dev.

jon-chuang: I can add a (not recommended) to make it obvious

jon-chuang:
> Can we add some explanation of the supported encodings to the command help as well?

What do you mean by explanation? 

Are the changes made sufficient?

jtbandes:
> > Can we add some explanation of the supported encodings to the command help as well?
> What do you mean by explanation?

I was talking about this line at the bottom of the file:
```
 	catCmd.PersistentFlags().BoolVarP(&catFormatJSON, "json", "", false, "print messages as JSON")
```
we could add some similar list of supported encodings so that the command's `--help` explains what is supported.

The other changes look pretty good to me as-is, but will leave to @wkalt / @james-rms for final approval.

go/cli/mcap/cmd/cat.go
```diff
@@ -109,57 +109,68 @@ func printMessages(
...
 		}
-		switch schema.Encoding {
-		case "ros1msg":
-			transcoder, ok := transcoders[channel.SchemaID]
-			if !ok {
-				packageName := strings.Split(schema.Name, "/")[0]
-				transcoder, err = ros.NewJSONTranscoder(packageName, schema.Data)
-				if err != nil {
-					return fmt.Errorf("failed to build transcoder for %s: %w", channel.Topic, err)
+		if schema == nil || schema.Encoding == "" {
```
james-rms, line 112: Schemaless channels should not have a schema record associated.
```suggestion
		if schema == nil {
```

go/cli/mcap/README.md
```diff
@@ -32,12 +32,14 @@ To install using [Homebrew](https://brew.sh) on macOS or Linux, run:
 
     brew install mcap
 
-#### Using go
+#### Using go (not recommended)
```
james-rms, line 35: this feels unnecessary - users can read the warning and make their own judgement around what to do.

james-rms (changes requested):
leave the`(not recommended)` out of the title in the README, looks good otherwise!

go/cli/mcap/cmd/cat.go
jon-chuang: Well, actually, at least according to https://github.com/foxglove/mcap/blob/0842cbf26dbf114a0fa0eb2dc36355f6db75382f/python/mcap/mcap/well_known.py#L20

This schemaless == `SelfDescribing` and can have `""` as a valid tag.

go/cli/mcap/README.md
jon-chuang: It was to address @jtbandes concern. But it does feel a little in your face.

go/cli/mcap/cmd/cat.go
james-rms: My mistake, you're right.

go/cli/mcap/cmd/cat.go
```diff
@@ -109,57 +109,68 @@ func printMessages(
...
+					return fmt.Errorf("failed to write message bytes: %w", err)
 				}
-				transcoders[channel.SchemaID] = transcoder
-			}
-			msgReader.Reset(message.Data)
-			err = transcoder.Transcode(msg, msgReader)
-			if err != nil {
-				return fmt.Errorf("failed to transcode %s record on %s: %w", schema.Name, channel.Topic, err)
+			default:
+				return fmt.Errorf("For schema-less channels, JSON output is only supported with 'json' message encoding. Found: %s", channel.MessageEncoding)
```
jon-chuang, line 119: Should we use `self-describing` as the terminology for `schema-less` to be consistent? @james-rms

jon-chuang: Seems that both terminology is used:
https://github.com/foxglove/mcap/search?q=self-describing
https://github.com/foxglove/mcap/search?q=schemaless

go/cli/mcap/cmd/cat.go
```diff
@@ -237,5 +248,7 @@ func init() {
 	catCmd.PersistentFlags().Int64VarP(&catStart, "start-secs", "", 0, "start time")
 	catCmd.PersistentFlags().Int64VarP(&catEnd, "end-secs", "", math.MaxInt64, "end time")
 	catCmd.PersistentFlags().StringVarP(&catTopics, "topics", "", "", "comma-separated list of topics")
-	catCmd.PersistentFlags().BoolVarP(&catFormatJSON, "json", "", false, "print messages as JSON")
+	catCmd.PersistentFlags().BoolVarP(&catFormatJSON, "json", "", false,
+		`print messages as JSON. Supported schema encodings: ros1msg, protobuf, and jsonschema.
```
james-rms, line 252:
```suggestion
		`print messages as JSON. Supported message encodings: ros1msg, protobuf, and JSON.`)
```

james-rms (approved):

go/cli/mcap/cmd/cat.go
jtbandes: ros1msg isn't a message encoding, it's a schema encoding. Also would rather see `json` than `JSON` since that's what the actual value of the encoding string is.

PR #860, opened 2023-03-17
feat(docs): add spec for `capnproto` as well-known encoding

jon-chuang:
### Public-Facing Changes
add spec for `capnproto` as well-known encoding

### Description
The spec is overly verbose to be descriptive but we will move the details into the tooling.

Follow up: 
- use cgo to allow the CLI to perform the schema serialization. (there is currently no canonical serialization of capnproto schemas. See e.g. https://groups.google.com/g/capnproto/c/Ivl3u4JJoEQ)


docs/specification/appendix.md
```diff
@@ -67,6 +71,12 @@ Schema `encoding` may only be omitted for self-describing message encodings such
...
 
+### capnproto
+
+- `name`: Fully qualified name: `[schema display name].Name`. For example, the struct `Bar` in schema file with path `path/to/files/schema.capnp` has the default qualified name `path/to/files/schema.capnp.Bar`. `[schema display name]` can be overridden in the `displayName` field of each schema file.
+- `encoding`: `capnproto`
+- `data`: The target `Schemas` and their dependencies serialized into a byte stream in the following format `[ schema 0 | schema 1 | ... | schema n ]`, via the method [`Schema::asUncheckedMessage`](https://github.com/capnproto/capnproto/blob/b2fdf71399accdf4ab93d38421a3225440b1d108/c%2B%2B/src/capnp/schema.h#L91). The schemas do not need to be an any particular order. TODO: This can be generated by running e.g. `mcap dump-capnp-schema --files my_schemas/*.capnp [--unqualified-names] > my_schemas.capnp.schema`. `--unqualified-names` will cause e.g. the `name` for `Bar` in `my_schemas/foo.capnp` to be `foo.Bar`. Note that this will fail for duplicate file names (we will validate duplication of filenames).
```
jtbandes, line 78: Hm, I wonder if we should use the CodeGeneratorRequest format instead, described here: https://github.com/capnproto/capnproto/issues/673#issuecomment-388994957

jon-chuang: Ooh, great catch! Will look into that

jon-chuang: Updated to use it, seems to work. PTAL.

jon-chuang:
I've validated that it correctly load the openpilot `log.capnp:Event` schema in the browser
Screenshot from 2023-03-18 14-26-21

docs/specification/appendix.md
```diff
@@ -67,6 +71,12 @@ Schema `encoding` may only be omitted for self-describing message encodings such
 - `encoding`: `flatbuffer`
 - `data`: A [reflection.Schema](https://github.com/google/flatbuffers/blob/master/reflection/reflection.fbs) flatbuffer describing the parsed schema encoded as a binary flatbuffer. This can be generated by running `flatc -b --schema` on the fbs file.
 
+### capnproto
+
+- `name`: Fully qualified name: `[schema display name]:Name`. For example, the struct `Bar` in schema file with compilation path `path/to/files/schema.capnp` has the default qualified name `path/to/files/schema.capnp:Bar`. `[schema display name]` can be overridden to limited effect via `--src-prefix`.
```
jtbandes, line 76: From the docs it sounds like this qualified name may not be unique. Do you have any other ideas for how to identify nodes in the schema?

Also discussing at https://github.com/capnproto/capnproto/issues/673#issuecomment-1476783025

I found some examples of dynamic schema usage here:
- https://capnproto.org/cxx.html
- https://groups.google.com/g/capnproto/c/sNEVK0WKa7o/m/7gFCFmEpvcEJ
- https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/capnp/schema-loader.h#L33-L34

But I don't see any discussion here of how to use a human-readable name to identify a schema in a CodeGeneratorRequest.

jon-chuang: > From the docs it sounds like this qualified name may not be unique.

I think this is an unlikely edge case. In most cases, the names will be unique. We will throw an error in `capnp-reflection` library or mcap CLI when loading two schemas with the same `displayName`. 

If users report this to us, we can suggest fixes.

The specific scenario where displayName may not be unique that I can think of is:
```bash
> ls tree
schemas
|-- my_schema.capnp
|-- subdir
     |-- my_schema.capnp
       
> cat my_schema.capnp

import "../dep.capnp"

struct MyStruct {
  my_dep_struct: "myschema.capnp:MyStruct",
}

> capnp compile --src-prefix=my_schema ./subdir/my_schema.capnp | capnp convert binary:json CodeGeneratorRequest
..
nodes: [
  .. 
  id: 1234,
  displayName: "my_schema.capnp:MyStruct"
  ..
  id: 5432,
  displayName: "my_schema.capnp:MyStruct"
  ..
 
```

 In this case, the error is totally fixable if the user omits `--src-prefix`. Naming two files the same name is a bad practice and unlikely in my opinion.

jon-chuang: The alternative is to use the unique schema `Id`, however, this is not human readable and not present in the schema file (it requires the capnp compiler to show)

docs/specification/appendix.md
```diff
@@ -33,6 +33,10 @@ The Channel `message_encoding` field describes the encoding for all messages wit
 
 - `message_encoding`: [`flatbuffer`](https://google.github.io/flatbuffers/)
 
+### capnproto
+
+- `message_encoding`: [`capnproto`](https://capnproto.org/)
```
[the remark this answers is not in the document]
jon-chuang: > Do we need to specify packed or not packed

Good question. 

> Cap’n Proto messages are often “packed” with a simple compression algorithm that deflates zero-value bytes.

It does seem that one has to select either packed or unpacked. I'm not sure if we should first serialize the leading byte of `data` to indicate pack-edness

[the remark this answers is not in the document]
jon-chuang: In our use case, I don't think we would encourage users to pack, since we already compress the file, and unpacking will cause an additional unpacking step.

I am for not supporting packed messages.

banisadr:
@jon-chuang based on your messages in the Foxglove Community Slack, it seems like you've already got deserialization working for capnproto in Studio? Would you be willing to make a PR to [foxglove/studio](https://github.com/foxglove/studio) to add deserialization support as well as add campnproto to [foxglove/schemas](https://github.com/foxglove/schemas)?

jon-chuang:
> Would you be willing to make a PR to [foxglove/studio](https://github.com/foxglove/studio) to add deserialization support as well as add campnproto to [foxglove/schemas](https://github.com/foxglove/schemas)?

Yes. But let’s get agreement on the spec here first.

For foxglove/schemas, I believe this is lower priority (L2) since it doesn’t block useability (user-script/plugin workarounds are just fine)

james-rms (changes requested):
A few changes required here:

### Schema.name
As discussed above, the `schema display name` is explicitly not meant to be parsed, not intended to be unique, and is not resistant to renames. With the proposal as it stands, a reader has no way to know if a schema with different content than the one on disk is an evolution of that schema, or just a different type with the same name. Using the schema display name as a schema identifier is explicitly against the intentions of the capnproto documentation.

The MCAP schema `name` needs to include the capnproto schema ID somehow. Responding to your earlier comment:

> The alternative is to use the unique schema Id, however, this is not human readable

This could be solved by including both display name and ID in the `Schema.name` field, perhaps separated by a space character.

>  and not present in the schema file for e.g. struct schemas (it requires the capnp compiler to generate)

This PR is already requiring the user to use the capnp compiler to generate the `Schema.data` field, so I don't think it's too onerous to use one to generate the `Schema.name` field as well.

#### Channel.message_encoding

This PR should specify that messages should be encoded in their "unpacked" form.

defunctzombie:
> This could be solved by including both display name and ID in the `Schema.name` field, perhaps separated by a space character.

Does the display name include any path information? Or is that not considered part of the display name?

jon-chuang:
> Does the display name include any path information? Or is that not considered part of the display name?

This is optional/configurable. Generally, it does.

banisadr:
@jon-chuang is this pr still on your radar? I'm still aligned that its a good idea if you're able to address @james-rms's feedback.

amacneil:
In case anyone is still following this issue - revived proposal to add capnproto to mcap:

- https://github.com/foxglove/mcap/pull/1788

PR #863, opened 2023-03-21
[C++] Add setChunkOptions method

jhurliman:
### Public-Facing Changes
- [C++] McapWriter.setChunkOptions() for closing the current chunk and starting a new chunk with new options (chunk size and compression settings)

### Description

The C++ library gives fine-grained control over closing the current chunk and starting a new chunk, but the compression and chunk size options cannot be modified after opening a file for writing. This new method allows a new chunk to be started with new chunk size and compression options. This is practically useful when writing channels to separate parts of an MCAP file, such as writing small compressible topics first with 1MB size + ZSTD, followed by large incompressible video/image topics with 4MB size + NONE.

### DIFF
--- a/cpp/mcap/include/mcap/writer.hpp
+++ b/cpp/mcap/include/mcap/writer.hpp
@@ -389,6 +389,17 @@ public:
   Status write(const Metadata& metadata);
 
   /**
+   * @brief Set the chunk size, compression type, and compression level to use
+   * for the next Chunk record. This will close any open chunk and start a new
+   * one with the specified options.
+   *
+   * @param chunkSize Upper limit of the chunk size in bytes.
+   * @param type Compression type to use.
+   * @param level Compression level to use.
+   */
+  void setChunkOptions(size_t chunkSize, Compression type, CompressionLevel level);
+
+  /**
    * @brief Current MCAP file-level statistics. This is written as a Statistics
    * record in the Summary section of the MCAP file.
    */
--- a/cpp/mcap/include/mcap/writer.inl
+++ b/cpp/mcap/include/mcap/writer.inl
@@ -640,6 +640,30 @@ Status McapWriter::write(const Metadata& metadata) {
   return StatusCode::Success;
 }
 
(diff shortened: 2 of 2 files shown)


james-rms (commented):
It'd be good to add a test here.

cpp/mcap/include/mcap/writer.inl
```diff
@@ -640,6 +640,30 @@ Status McapWriter::write(const Metadata& metadata) {
...
+  uncompressedChunk_.reset();
+  lz4Chunk_.reset();
+  zstdChunk_.reset();
+
+  chunkSize_ = chunkSize;
+  compression_ = type;
+
+  switch (type) {
+    case Compression::None:
+    default:
```
james-rms, line 655: i think in the default case we should assert that the switch is exhaustive.

cpp/mcap/include/mcap/writer.hpp
```diff
@@ -388,6 +388,17 @@ class MCAP_PUBLIC McapWriter final {
    */
   Status write(const Metadata& metadata);
 
+  /**
+   * @brief Set the chunk size, compression type, and compression level to use
+   * for the next Chunk record. This will close any open chunk and start a new
```
defunctzombie, line 393: I find it strange to have a function called `setChunkOptions` that has this side effect.

defunctzombie: What about removing the side effect? Set these options on the next chunk that is started? The user can close the current chunk if they'd like?

Your documentation already has the language "for the next Chunk record" so all you'd change is removing "This will close any open chunk".

Is there a method to close the current chunk?

jhurliman: Yes there is a `closeLastChunk()` method. I will update this to use additional McapWriter member variables for "next chunk settings".

jtbandes:
Seems like some potential improvements were identified but this PR hasn't been revisited recently – any interest in continuing to pursue it?

defunctzombie:
Closing this for inactivity. If you do come back to this and want to address the comments we can revisit.

PR #864, opened 2023-03-23
go/ros/bag2mcap: re-map 32-bit connection IDs to channel IDs

james-rms:
### Public-Facing Changes

users with ROS bags that have extremely high connection ID values (perhaps from large or very-long-running stacks) will see their bags get correctly converted and not have connections clobber eachother.

### Description

Fixes a bug where we drop the 2 most significant bytes from the connection header when reading it. This could lead to accidentally merging some channels.

### DIFF
--- a/go/ros/bag2mcap.go
+++ b/go/ros/bag2mcap.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math"
 
 	"github.com/foxglove/mcap/go/mcap"
 	"github.com/pierrec/lz4/v4"
@@ -15,6 +16,10 @@ import (
 
 var (
 	BagMagic = []byte("#ROSBAG V2.0\n")
+)
+
+var (
+	ErrTooManyConnections = fmt.Errorf("bag contains more than %d connection IDs", math.MaxUint16)
 )
 
 type BagOp byte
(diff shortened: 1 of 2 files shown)


go/ros/bag2mcap.go
```diff
@@ -254,14 +276,15 @@ func Bag2MCAP(w io.Writer, r io.Reader, opts *mcap.WriterOptions) error {
...
+	connIds := make([]uint32, 0)
 	schemas := make(map[string]uint16)
 	return processBag(r,
 		func(header, data []byte) error {
 			conn, err := extractHeaderValue(header, headerConn)
 			if err != nil {
 				return err
 			}
-			connID := binary.LittleEndian.Uint16(conn)
+			connID := binary.LittleEndian.Uint32(conn)
```
wkalt, line 287: instead of adding the scan to each message iteration, what about just doing something like this?

```
func safeUint32Downcast(x uint32) (uint16, error) {
  if x > math.MaxUint16 {
    return 0, ErrTooManyConnections
  }
  return uint16(x), nil
}
```

then call it right here -

```
chanID, err := safeUint32Downcast(binary.LittleEndian.Uint32(conn)))
if err != nil {
  return err
}
...
```

this gets us constant time performance in connections - no need to traverse the list of connIds each time. The only thing it hurts us on afaict is if someone is assigning connection IDs out of sequence, _and_ surpassing this limit. I think that seems unlikely enough that we can wait for the first user report to attempt to accommodate it. The change in behavior to crash when this happens instead of silently corrupting data, is already a huge improvement IMO.

wkalt (approved):

PR #865, opened 2023-03-25
Better example for metadata name

amacneil:
### Public-Facing Changes
The metadata `name` of `map_metadata` is redundant, we know it is a metadata record so we don't need that in the name. Also, the `map` part doesn't match any of the example key/values in the next line.

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -279,7 +279,8 @@ A metadata record contains arbitrary user data in key-value pairs.
 
 | Bytes | Name | Type | Description |
 | --- | --- | --- | --- |
-| 4 + N | name | String | Example: `mycompany`. |
+| 4 + N | name | String | Example: `my_company_name_hardware_info`. |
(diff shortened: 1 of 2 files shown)


docs/specification/README.md
```diff
@@ -279,7 +279,7 @@ A metadata record contains arbitrary user data in key-value pairs.
 
 | Bytes | Name | Type | Description |
 | --- | --- | --- | --- |
-| 4 + N | name | String | Example: `map_metadata`. |
+| 4 + N | name | String | Example: `mycompany`. |
```
james-rms, line 282: `mycompany` on its own might be a bit oblique, better to make the example specific to a category of metadata.

```suggestion
| 4 + N | name | String | Example: `my_company_name_hardware_info`. |
| 4 + N | metadata | `Map<string, string>` | Example keys: `part_id`, `serial`, `board_revision` |
```

wkalt (approved):

PR #866, opened 2023-03-27
Upgrade to Yarn v3

amacneil:
### Public-Facing Changes
N/A

### Description
Upgrade to Yarn v3 for consistency with other repos.


typescript/core/.yarnrc
```diff
@@ -1,2 +0,0 @@
-version-tag-prefix "releases/typescript/core/v"
-version-git-message "TypeScript v%s"
```
amacneil, line 2: @jtbandes I did some brief searching and can't find the equivalent for these in yarn 3. Maybe we should switch to using npm to bump versions? Or have you used any equivalent in our yarn 3 repos?

jtbandes: Using npm or yarn to bump versions + create tags only works if you have permissions to push directly to main. Generally it seems like that is no longer the case in many of our repos, so I don't know if we still need to bother with these.

package.json
```diff
@@ -22,5 +22,6 @@
     "prettier": "^2.8.4",
     "tailwindcss": "^3.2.7",
     "typedoc": "^0.23.25"
-  }
+  },
+  "packageManager": "yarn@3.5.0"
```
jtbandes, line 26: Does this mean we are using corepack? Does using corepack mean we could get rid of the .yarn/releases/yarn-3.5.0.cjs that we check in?

jtbandes (approved):
🤷 still haven't seen what the real benefit of using yarn v3 is. I think adrianbot has several other TS repos it needs to update next 🙌

package.json
amacneil: Oh, you are right! I assumed it would not be possible since none of our other repos do that. But I just tested it and seems to work great.

PR #872, opened 2023-03-30
python-ros2-support: provide type and full text attributes

james-rms:
### Public-Facing Changes

Provides the `_type` and `_full_text` names as specified in https://mcap.dev/docs/python/ros2_noenv_example.html .


james-rms:
cc. @jhurliman if you're interested

python/mcap-ros2-support/mcap_ros2/_dynamic.py
```diff
@@ -246,28 +246,29 @@ def _make_encode_message(
...
         (SimpleNamespace,),
         {
-            "__name__": msg_name,
-            "__slots__": [field.name for field in fields],
+            "__name__": msgdef.msg_name,
+            "__slots__": [field.name for field in msgdef.fields],
             "__repr__": __repr__,
             "__str__": __repr__,
+            "_type": str(msgdef.base_type),
+            "_full_text": str(msgdef),
```
jtbandes, line 262: Let's bikeshed these names a bit. If we are intending people to use these attributes in their everyday code, I don't think they should be `_`-prefixed.

james-rms: the colour of this bikeshed is consistent with the ROS 1 bikeshed: https://github.com/foxglove/mcap/pull/873

PR #873, opened 2023-03-30
go/mcap: do not write empty message indexes

### LINKED ISSUE #862: MCAP cli produces empty message index
**Description**
The MCAP cli produces an empty message index instead of omitting it. It seems that this slows down Foxglove and the Python lib.

- Version: CLI v0.0.24
- Platform: mac OS

**Steps To Reproduce**
`mcap compress` on any file

**Expected Behavior**
no message index at all for topics not in this chunk

james-rms:
### Public-Facing Changes

Go MCAP writer no longer produces message index records for channels not present in a chunk.

### Description

Fixes #862


go/mcap/writer.go
```diff
@@ -463,7 +463,8 @@ func (w *Writer) flushActiveChunk() error {
 	messageIndexOffsets := make(map[uint16]uint64)
 	if !w.opts.SkipMessageIndexing {
 		for _, chanID := range w.channelIDs {
-			if messageIndex, ok := w.messageIndexes[chanID]; ok {
+			messageIndex, ok := w.messageIndexes[chanID]
+			if ok && !messageIndex.IsEmpty() {
```
wkalt, line 467: this seems ok to me, I agree current behavior is likely oversight.

One thing that is unclear to me is, how should we behave if a channel record shows up in a chunk, with no messages?


Another side-note in case it's helpful - I think the extra tooling around messageIndex (Add, IsEmpty, Entries, Reset) can probably be refactored out in favor of a slice of MessageIndexEntry maintained in the writer for each chunk. The extra structure was originally required when MessageIndex was maintaining a sort on the records, but it no longer does that. No need to act on this - just possibly useful context.

wkalt (approved):

go/mcap/writer.go
james-rms: > how should we behave if a channel record shows up in a chunk, with no messages?

I don't feel like this is unclear to me - the message index does not index channel records, so the presence of channel records shouldn't affect the content or presence of message index records.

PR #874, opened 2023-03-30
Create a flake.nix that packages the mcap-cli

therishidesai:
### Public-Facing Changes
None

### Description

Added a nix flake so you can use the [nix package manager](https://nixos.org/) to get the mcap-cli.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ site/
 
 *.db3-shm
 *.db3-wal
+
+result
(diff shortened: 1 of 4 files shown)


defunctzombie:
I like the idea of it living in the upstream nixpkgs repo rather than this repo. What would be a good change for this repo is adding a CI step similar to https://github.com/foxglove/mcap/blob/main/.github/workflows/ci.yml#L434 which would update the nixpkgs when we make new releases. We've found that this level of automation is good for keeping our releases updated in the respective package ecosystems.

defunctzombie:
Closing in favor of the upstream nixpkgs PR

Issue #723, opened 2022-11-15
specification: user-defined statistics


wkalt:
Related to https://github.com/foxglove/mcap/issues/384

It would be nice if there were a way for writers to define and use custom statistics in their mcap files and have them surfaced by the "info" subcommand. I think this could be implemented with a new record type like "custom statistic", which could minimally contain "name" and "value" fields. Better IMO would be to provide a little more structure, like supporting "channel statistic" and "file statistic" variants, or even "channel statistic", "file statistic", "attachment statistic", "metadata statistic", "chunk statistic". The channel statistics would have channel IDs associated with them, file statistics would be whole-file, and metadata/attachment/chunk statistics could contain references to the relevant record offset. I am imagining all these records would be written to the summary section somewhere, I guess in a new statistics section. The "info" command could then display file and channel statistics, and the list attachments/chunks/metadata/channels commands could show the relevant type of statistic as well.

Some additional things to think about:
* what about within-chunk channel statistics?

james-rms:
1. This seems like a reasonably clean way to head off a category of statistic addition requests.
2. One awkward point here is that we have no floating-point value type in the MCAP spec as it currently stands.
3. It's also a little awkward that we need to add a new opcode for every concept that one might want to have a statistic about: Channel, Chunk, File, Attachment, Metadatat etc.

wkalt:
regarding 2 - I think we would probably just introduce one. The representation in use in file formats I have encountered is IEEE 754.

about 3, I agree - however I'm not sure the alternatives are better. I can imagine two ways to handle it. We could go with a scheme that looks like this:

```
Statistic
Type: channel | attachment | chunk | metadata | ...
Name: string
Value: float64
Reference: channelID | attachment offset | chunk offset | ??? | ...
```

or split out different record types as the OP suggests. I don't know with the previous scheme how we encode "chunk channel statistics" if that is something we would want to provide, and the implied dispatch on user-supplied strings to interpret the "reference" field seems kind of brittle to me. But I think at the end of the day you could use make either approach work.

wkalt:
Another interesting application for custom statistics would be accelerating text search using trigrams. Consider that in typical robotics data, text values within specific fields are extremely conserved: they will generally derive from a finite number of error strings that exist in code somewhere. Typical searches will be for rare strings like "ERROR", not common strings like "success".

[...]

Issue #384, opened 2022-05-06
store per-topic size statistics


wkalt:
This would be a nice feature. May be easy to do backward compatibly.

PR #876, opened 2023-04-03
custom statistics proposal

james-rms:

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -312,6 +312,53 @@ A Data End record indicates the end of the data section.
 | --- | --- | --- | --- |
 | 4 | data_section_crc | uint32 | CRC32 of all bytes from the beginning of the file up to the DataEnd record. A value of 0 indicates the CRC32 is not available. |
 
+### Custom Statistic (op=0x10)
+
+Records some statistic about the MCAP content not covered by the [Statistics](#statistics-op0x0b) record.
+
+| Bytes | Name           | Type    | Description                               |
+| ----- | -------------- | ------- | ----------------------------------------- |
+| 4 + N | name           | string  | the human-readable name of the statistic. |
+| 8     | value          | float64 | The statistic value.                      |
+| 4 + N | subject        | string  | The subject of this statistic.            |
+| 4 + N | display_format | string  | how this statistic should be displayed.   |
+
+#### Custom statistic subjects
+
+Custom Statistics specify a subject, which indicates what this statistic is about. Some subjects are well-known:
+
+| Subject | Description |
+| --- | --- |
+| `messages` | a statistics about all messages |
+| `channels` | a statistic about all channels |
+| `schemas` | a statistic about all schemas |
+| `metadatas` | a statistic about all metadata records |
+| `attachments` | a statistic about all attachment records |
+| `chunks` | a statistic about all chunks |
+| `this_channel` | a statistic about the channel record immediately preceding |
+| `this_schema` | a statistic about the schema record immediately preceding |
+| `this_metadata` | a statistic about a metadata record in the Data section, indicated by the MetadataIndex record immediately preceding |
+| `this_attachment` | a statistic about a attachment record in the Data section, indicated by the AttachmentIndex record immediately preceding |
+| `this_chunk` | a statistic about a chunk record in the Data section, indicated by the ChunkIndex record immediately preceding |
+
+All other subjects not prefixed with `_` are reserved. The empty string `""` is not a valid subject.
+
+#### Custom statistic display format
+
+Custom statistics include a `display_format` string which indicates how the statistic should be presented to human readers. SI units should be used where applicable.
+
+TODO fully specify format syntax
+
+##### Examples
+
+| `display_format` | Displayed result | Description |
+| --- | --- | --- |
+| `""` | `1234.567890000` | Default display |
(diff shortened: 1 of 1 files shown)


docs/specification/README.md
```diff
@@ -312,6 +312,53 @@ A Data End record indicates the end of the data section.
...
+
+| Subject | Description |
+| --- | --- |
+| `messages` | a statistics about all messages |
+| `channels` | a statistic about all channels |
+| `schemas` | a statistic about all schemas |
+| `metadatas` | a statistic about all metadata records |
+| `attachments` | a statistic about all attachment records |
+| `chunks` | a statistic about all chunks |
+| `this_channel` | a statistic about the channel record immediately preceding |
```
wkalt, line 338: Few thoughts on this -
1. I do not like that this introduces a new naming scheme with "this_*". I think we can avoid that.

2. I don't get how these will get recorded. I would expect the custom statistics records to get recorded into a custom statistics portion of the summary section (i.e these are not intermixed with the other portions of the summary section - there is a summary offset record for custom statistic, pointing at a contiguous run of these).

So what's the "immediately preceding" channel? Likewise for metadata/attachment/chunk.

3. Is it necessary to have statistics for both "this channel" and "channels"? If you are recording the "this channel" stats, would you be able to synthesize what you want for "channels"? And would tooling be able to do generic handling based on types (e.g for numeric types, compute the averages, distributions, etc maybe) to do a good enough job of covering everything?

If we can do that, we avoid forcing users to record both "this_chunk: message_count" and "chunks: average message count".

wkalt: Couple alternatives to throw out -

Single record: downside - "id" and opcode are abused.
```
Custom Statistic:
name string
value float64
op OpCode
id uint64

// Legal OpCodes are Channel and Chunk, as well as 0x00 which denotes a full-file statistic. 
```

Separate records
```
FileStatistic
name string
value float64

ChannelStatistic
name string
value float64
channelID uint16

ChunkStatistic
name string
value float64
offset uint64

```

In either case, we'd use the existing index structure. So for each record type getting introduced, there would be a summary offset record with that ID, pointing up at a contiguous run of those records in the summary section.

we can include display format on these as you have - haven't thought that through but seems like a reasonable idea to me.

docs/specification/README.md
```diff
@@ -312,6 +312,53 @@ A Data End record indicates the end of the data section.
...
+| `this_attachment` | a statistic about a attachment record in the Data section, indicated by the AttachmentIndex record immediately preceding |
+| `this_chunk` | a statistic about a chunk record in the Data section, indicated by the ChunkIndex record immediately preceding |
+
+All other subjects not prefixed with `_` are reserved. The empty string `""` is not a valid subject.
+
+#### Custom statistic display format
+
+Custom statistics include a `display_format` string which indicates how the statistic should be presented to human readers. SI units should be used where applicable.
+
+TODO fully specify format syntax
```
wkalt, line 350: I'm not sure if we need to specify a syntax. Another option would be to define an enum for regular kinds of types, and use that single byte value in the field. Then implementations can render the units in whatever way makes sense. This would probably lower the amount of custom parsing required for implementers.

docs/specification/README.md
```diff
@@ -312,6 +312,53 @@ A Data End record indicates the end of the data section.
...
+| Bytes | Name           | Type    | Description                               |
+| ----- | -------------- | ------- | ----------------------------------------- |
+| 4 + N | name           | string  | the human-readable name of the statistic. |
+| 8     | value          | float64 | The statistic value.                      |
+| 4 + N | subject        | string  | The subject of this statistic.            |
+| 4 + N | display_format | string  | how this statistic should be displayed.   |
+
+#### Custom statistic subjects
+
+Custom Statistics specify a subject, which indicates what this statistic is about. Some subjects are well-known:
```
wkalt, line 328: specification of a subject is very similar to what summary offset records do, with their opcode field. it would be ideal if we could use the same mechanism here IMO

jtbandes:
Where can I read more about the motivation/backstory behind this proposal?

wkalt:
@jtbandes context here: https://github.com/foxglove/mcap/issues/723

jtbandes:
I don't see any examples there of what this might be used for. Do we have a list of examples we or our users are thinking about using this for?

wkalt:
There's a ticket linked in the linked ticket with one example: https://github.com/foxglove/mcap/issues/384, which is about per-topic size statistics. We have also gotten requests for statistics on per-topic compression ratios (which I doubt would go into the official writers but could be implemented with this proposal). Separate thread recently raised per-topic coverage.

In general the idea with this record type would be to head off future requests for spec extensions with new statistics, by instead supporting them generically.

docs/specification/README.md
james-rms: > If you are recording the "this channel" stats, would you be able to synthesize what you want for "channels"?

I considered allowing the Custom Specification record to specify how the reader should aggregate when more than one is concerned - would be neccessary for making "all channels" stats from individual stats as you say. It would also be neccessary to preserve custom stats across merges. The problem is that there are several ways for this to work (weighted average, min, max, sum, etc.) which is hard to specify ahead of time. i also considered embedding a little DSL to do this maths (something like `bc` notation or APL). In the end I don't think it's worth the API surface and implementation burden for the reader.

james-rms:
Closing this for now as i'm not planning to think about this for the next few weeks. May reopen with more interest.

PR #877, opened 2023-04-04
Fix bug in ros1msg parser

wkalt:
Prior to this commit, if a ROS 1 message definition contained both qualified and unqualified complex types in a particular order, it would incorrectly infer the parent package of a succeeding unqualified type, resulting in an erroneous failure to find a submessage definition. For a concrete example, a message definition like this:

```
        Foo foo
        ===
        MSG: Foo
        other_package/Bar bar
        FooBar foobar
        ===
        MSG: other_package/Bar
        string data
        ===
        MSG: original_package/FooBar
        string data
```

would result in the parser attempting to look up other_package/FooBar for FooBar's type definition, instead of looking for it in the parent package of the overall type.

### DIFF
--- a/go/ros/ros1msg/ros1msg_parser.go
+++ b/go/ros/ros1msg/ros1msg_parser.go
@@ -77,9 +77,13 @@ func resolveDependentFields(
 			// type with its parent package.
 			// 3. The type may be "Header". This is a special case that needs to
 			// translate to std_msgs/Header.
+
+			// Start by assuming the parent package of the field is that of the overall type.
+			fieldParentPackage := parentPackage
+
 			typeIsQualified := strings.Contains(fieldType, "/")
 			if typeIsQualified {
-				parentPackage = strings.Split(fieldType, "/")[0]
+				fieldParentPackage = strings.Split(fieldType, "/")[0]
 			}
 			subdefinition, typeIsPresent := dependencies[fieldType]
 			switch {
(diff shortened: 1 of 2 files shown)


go/ros/ros1msg/ros1msg_parser.go
```diff
@@ -77,9 +77,13 @@ func resolveDependentFields(
 			// type with its parent package.
 			// 3. The type may be "Header". This is a special case that needs to
 			// translate to std_msgs/Header.
+
+			// Start by assuming the parent package of the field is that of the overall type.
+			fieldParentPackage := parentPackage
+
 			typeIsQualified := strings.Contains(fieldType, "/")
 			if typeIsQualified {
-				parentPackage = strings.Split(fieldType, "/")[0]
```
wkalt, line 82: the bug was, previously this would result in "parentPackage" getting modified for the next field as well - which is independent of this one.

jtbandes:
Under what circumstances would we ever see `MSG: FooBar` unqualified? My impression was that the ROS 1 msg generation tooling never generates that.

wkalt:
@jtbandes sorry, I have mischaracterized the bug in the commit message. The test properly captures it. It's not

```
MSG: FooBar
```

it's

```
MSG: original_package/FooBar
```

where original_package is the package of the top-level type. The "unqualified" FooBar in the field definition then attempts to look up other_package/FooBar instead of original_package/FooBar. I will update the commit message.

james-rms (approved):

PR #883, opened 2023-04-14
python/mcap: add API to decode messages while reading

james-rms:
### Public-Facing Changes

Add option to include a dictionary of message decoders when instantiating a reader. 

### Description

This is a rework of https://github.com/foxglove/py-data-platform/pull/62 , pushing this functionality into the core McapReader class.


defunctzombie (commented):
Thoughts from my first read through:

1. Do we need decoders _and_ schemaless_decoders or can we find a nice way for users to specify only decoders?
2. Do we need `iter_decoded_messages`? Or can we make `decode` an option to `iter_messages` and you get back decoded messages alongside the raw message data? @foxglove/rosbag behaves in such a way. Is this better or worse than the separate `iter`?
3. Is it worth going one layer "up" with this and allowing the user to specify schema factories instead of decoder functions? i.e. I would give you a function to call when you encounter a new schema and it would return something that is used to decode message records that are for that schema record? Basically handle the logic of dealing with tracking channels and schemas for the user as well?
4. The tests show a basic example with json decoding, but don't get into the nuances of protobuf where you might need to track the Schema and have some mapping to actually lookup the right decoder. This makes the test example code look simple but for binary data users I think it would be more complex in practice?

python/mcap-protobuf-support/mcap_protobuf/reader.py
```diff
@@ -1,17 +1,27 @@
...
-:py:class:`mcap_protobuf.decoder.Decoder` and :py:class:`mcap.reader.McapReader` objects
-directly.
+"""the :py:mod:`mcap_protobuf.reader` module is deprecated. For similar functionality,
+instantiate the :py:class:`mcap.reader.McapReader` with a
+:py:class:`mcap_protobuf.decoder.DecoderFactory` instance, eg:
+
+>>> from mcap.reader import make_reader
+>>> from mcap_protobuf.decoder import DecoderFactory
+>>> reader = make_reader(open("proto.mcap", "rb"), decoder_factories=[DecoderFactory()])
+>>> for schema_, channel_, message_, proto_msg in reader.iter_decoded_messages():
```
defunctzombie, line 8: What is `proto_msg`? Is that something that would have a meaning to python users?

Why are `schema_, channel_, message_,` variables part of the returned set of variables for each iteration? Could all of these instead by in the "message event" and within that you have the `log_time, publish_time` etc?

python/mcap-protobuf-support/mcap_protobuf/decoder.py
```diff
@@ -76,22 +78,16 @@ def _add_file(file_descriptor: FileDescriptorProto):
...
-        :type message: mcap.records.Message
-        :raises McapProtobufDecodeError: if the content could not be decoded as a protobuf message
-            with the given schema.
-        :return: The decoded message content.
-        """
-        if schema.encoding != SchemaEncoding.Protobuf:
-            raise McapProtobufDecodeError(
-                f"can't decode schema with encoding {schema.encoding}"
-            )
+    def decoder_for(
```
defunctzombie, line 81: Missing documentation?

python/mcap-protobuf-support/mcap_protobuf/decoder.py
```diff
@@ -14,7 +16,7 @@ class McapProtobufDecodeError(McapError):
     pass
 
 
-class Decoder:
+class DecoderFactory(McapDecoderFactory):
```
defunctzombie, line 19: Missing documentation?

defunctzombie (approved):
👍 Makes sense. I'm not a python snake charmer so I've added @emersonknapp to the review to see if he has additional thoughts on naming or interface boundaries.

One thing that I wondered about is why the argument to `McapReader` is called `decoder_factories` rather than `decoders`? As a caller I would provide my list of instantiated decoders where each decoder is for some schema/message-encoding pairing? Factories don't seem to add a useful layer of indirection here.

python/mcap-protobuf-support/mcap_protobuf/reader.py
james-rms: > What is proto_msg? Is that something that would have a meaning to python users?

It's the protobuf message, or decoded message content in other words.

> Why are schema_, channel_, message_, variables part of the returned set of variables for each iteration? Could all of these instead by in the "message event" and within that you have the log_time, publish_time etc?

There's information in the schema, channel, and message records which can be useful to the caller on every iteration. Hypothetically we could yield a single object here which contains some fields from each, but that requires either just including all fields of all records or selecting which fields might be interesting to all users. This seems like extra effort for little gain, also slightly less performant because we need to construct a new object and copy a bunch of fields instead of just yielding references to the underlying records.

emersonknapp (approved):
I'm not too familiar with this codebase yet so I don't have my smell-detector fully calibrated in terms of the interface boundary changes. That said, this seems fine to me.

I've left some small things, but I'll call them optional and approve.

python/mcap/tests/test_reader.py
```diff
@@ -37,7 +41,7 @@ def test_make_not_seeking(pipe: Tuple[IO[bytes], IO[bytes]]):
 
 
 @pytest.mark.parametrize("reader_cls", [SeekingReader, NonSeekingReader])
-def test_all_messages(reader_cls: Type[McapReader]):
+def test_all_messages(reader_cls: Union[Type[SeekingReader], Type[NonSeekingReader]]):
```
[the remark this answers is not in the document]
james-rms: This is to satisfy the type checker, which is otherwise unhappy with me constructing the class with a stream argument. The base class constructor does not take a stream.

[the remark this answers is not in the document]
emersonknapp: I see - is it possible to keep "unimplemented" version of the same constructor signature on the base class to avoid that type checker issue?

Again none of this is blocking i'm just curious to get to know this library better.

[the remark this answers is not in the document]
james-rms: I didn't want to do that because it's not a requirement of future McapReader subclasses to take the same constructor arguments - maybe they could take some other input options, not sure. The constructor is not part of the abstract interface.

PR #890, opened 2023-05-15
Fix mcap conversion for bzip2 bags

wkalt:
Prior to this commit, the bzip2 decompression was broken on bags with more than one chunk. The chunk reader was getting set to a basic reader on the underlying chunk data, rather than a bz2 decompressor (on chunks after the first).

This fixes that and also does a minor refactor to remove the concept of a resettable reader. This was only applicable to the lz4 decompressor previously (out of three - lz4, none, and bz2) and existed to fake generic support for resetting readers when underlying readers didn't support that. Rather than doing that, special case lz4 and use the other two reader types directly.


jtbandes (commented):
Can we add a test with a small .bz2-compressed bag file (or generate one on the fly in the tests)?

wkalt:
yeah, I'll add one. it'll need to be fixture data because annoyingly go doesn't contain a bz2 compressor, only decompressor

Issue #887, opened 2023-05-09
ParseHeader error in McapReader


smithbrow:
header->library is wrong in ParseHeader function of McapReader

- Version: 1.0.0
- Platform: cpp

```
--- a/cpp/mcap/include/mcap/reader.inl
+++ b/cpp/mcap/include/mcap/reader.inl
@@ -725,10 +725,14 @@ Status McapReader::ParseHeader(const Record& record, Header* header) {
       !status.ok()) {
     return status;
   }
-  const uint64_t maxSize = record.dataSize - 4 - header->profile.size();
-  if (auto status = internal::ParseString(record.data, maxSize, &header->library); !status.ok()) {
+  const uint64_t offset = 4 + header->profile.size();
+  const uint64_t maxSize = record.dataSize - offset;
+  if (auto status = internal::ParseString(record.data + offset, maxSize, &header->library);
+      !status.ok()) {
     return status;
   }
   return StatusCode::Success;
 }
```

PR #892, opened 2023-05-15
cpp: fix header parsing

james-rms:
### Public-Facing Changes

Header records will now be parsed correctly, instead of duplicating the profile into both profile and library fields.

Fixes https://github.com/foxglove/mcap/issues/887

### DIFF
--- a/cpp/mcap/include/mcap/reader.inl
+++ b/cpp/mcap/include/mcap/reader.inl
@@ -725,8 +725,11 @@ Status McapReader::ParseHeader(const Record& record, Header* header) {
       !status.ok()) {
     return status;
   }
-  const uint64_t maxSize = record.dataSize - 4 - header->profile.size();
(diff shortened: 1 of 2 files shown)


cpp/mcap/include/mcap/reader.inl
```diff
@@ -725,8 +725,11 @@ Status McapReader::ParseHeader(const Record& record, Header* header) {
       !status.ok()) {
     return status;
   }
-  const uint64_t maxSize = record.dataSize - 4 - header->profile.size();
-  if (auto status = internal::ParseString(record.data, maxSize, &header->library); !status.ok()) {
+  const size_t libraryOffset = 4 + header->profile.size();
+  const std::byte* libraryData = &(record.data[libraryOffset]);
```
jtbandes, line 729: nit
```suggestion
  const std::byte* libraryData = record.data + libraryOffset;
```

jtbandes (approved):
Seems like we should have conformance tests that include this as well.

PR #893, opened 2023-05-16
CLI: remove MCAP doctor on chunks with no messages, schemaless channels

james-rms:
### Public-Facing Changes

* Doctor will no longer raise an error if a chunk is present with no messages.
* Doctor will no longer raise an error when a schemaless channel record is present in a chunk. This was previously updated to not report an error, but that change only applied to channels outside of chunks.
* `mcap.Writer` now correctly sets the chunkEndTime to 0 for chunks with no messages, as specified in the spec: https://mcap.dev/specification/index.html#chunk-op0x06 .

### DIFF
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -197,21 +197,29 @@ func (doctor *mcapDoctor) examineChunk(chunk *mcap.Chunk) {
 		}
 	}
 
-	if minLogTime != chunk.MessageStartTime && chunkMessageCount != 0 {
-		doctor.error("Chunk.message_start_time %d does not match the earliest message log time %d",
-			chunk.MessageStartTime, minLogTime)
-	}
+	if chunkMessageCount != 0 {
+		if minLogTime != chunk.MessageStartTime {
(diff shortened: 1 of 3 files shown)


wkalt:
+1 for the doctor changes - for various reasons you might get a channel record in a chunk followed by a file close without additional messages, so that shouldn't be flagged by doctor. I don't really understand the second part (if schemaless channels are valid, why would we ever flag them?)

For the writer change, does this bring the go writer in line with the behavior of the other implementations? Seems like this could use a conformance test and/or a test for the writer.

james-rms:
@wkalt thanks, updated the description. The schemaless channel error was previously thought to be fixed, but it was only fixed for channels outside of chunks.

> For the writer change, does this bring the go writer in line with the behavior of the other implementations? Seems like this could use a conformance test and/or a test for the writer.

I found that the go writer would write different values for chunkStartTime and chunkEndTime to the chunk vs. the chunk index. I agree we should add a conformance test for this - if it's easy enough I'll try to add it in this PR.

defunctzombie (approved):

PR #894, opened 2023-05-16
specification: add omgidl to appendix

james-rms:

### DIFF
--- a/docs/specification/appendix.md
+++ b/docs/specification/appendix.md
@@ -107,6 +107,28 @@ The IDL definition of the type specified by `name` along with all dependent type
 - a line containing exactly 80 `=` characters, then
 - A line containing only `IDL: <package resource name>` for that definition. The space between `IDL:` and the package resource name is mandatory. The package resource name does not include a file extension.
 
+### omgidl
+
+The `omgidl` schema encoding stores [OMG IDL](https://www.omg.org/spec/IDL/4.2/About-IDL) definitions for (X)CDR-encoded messages.
+
+- `name`: The globally-scoped name of the message type, eg. `mycompany::mymodule::MyType`
+- `encoding`: `omgidl`
+- `data`: valid OMG IDL definition file contents with a definition for the message type and all referenced types.
+
+#### Producing `schema.data` for the `omgidl` schema encoding
+
+Most commonly, IDL definitions for separate types are stored in separate files, and use C++ preprocessor `#include` statements to include definitions for types referenced in that definition.
+
+For the `omgidl` schema encoding, the `data` field needs to contain all of those definitions in a single block of bytes. One way to produce this is to use a C++ compiler to resolve those `#include` directives:
+
+```bash
+# Run from the root of your IDL definition tree, or use additional `-I` arguments to indicate your
+# full include path to the `cpp` utility.
+# The `cpp` output may contain debugging lines such as "# 1 "mycompany/mymodule/MyType.idl",
+# the grep command here is intended to filter those out.
+$ cpp -I. mycompany/mymodule/MyType.idl | grep -v "^# \d" > all_definitions.idl
+```
+
 ### jsonschema
 
 - `name`: May contain any value


docs/specification/appendix.md
```diff
@@ -107,6 +107,28 @@ The IDL definition of the type specified by `name` along with all dependent type
...
+
+- `name`: The globally-scoped name of the message type, eg. `my_company::my_module::MyType`
+- `encoding`: `omgidl`
+- `data`: valid OMG IDL definition file contents with a definition for the message type and all referenced types.
+
+#### Producing `schema.data` for the `omgidl` schema encoding
+
+Most commonly, IDL definitions for separate types are stored in separate files, and use C++ preprocessor `#include` statements to include definitions for types referenced in that definition.
+
+For the `omgidl` schema encoding, the `data` field needs to contain all of those definitions in a single block of bytes. One way to produce this is to use a C++ compiler to resolve those `#include` directives:
```
jtbandes, line 122:
```suggestion
For the `omgidl` schema encoding, the `data` field needs to contain all of those definitions. One way to produce this is to use a C++ compiler to resolve those `#include` directives:
```

jtbandes (approved):

docs/specification/appendix.md
```diff
@@ -107,6 +107,28 @@ The IDL definition of the type specified by `name` along with all dependent type
 - a line containing exactly 80 `=` characters, then
 - A line containing only `IDL: <package resource name>` for that definition. The space between `IDL:` and the package resource name is mandatory. The package resource name does not include a file extension.
 
+### omgidl
+
+The `omgidl` schema encoding stores [OMG IDL](https://www.omg.org/spec/IDL/4.2/About-IDL) definitions for (X)CDR-encoded messages.
+
+- `name`: The globally-scoped name of the message type, eg. `my_company::my_module::MyType`
```
defunctzombie, line 114: Is the `my_company` part in the idl schemas as a module?

docs/specification/appendix.md
defunctzombie, line 122: I'd put "One way to produce..." on a separate paragraph and add a bit more description to the paragraph describing the data field as one where you need to resolve all the referenced types into one single "source file".

defunctzombie:
Overall seems good - lets circulate and see if we can get some example files to prove this out.

docs/specification/appendix.md
james-rms: yes - not sure how to make that more obvious.

docs/specification/appendix.md
```diff
@@ -107,6 +107,28 @@ The IDL definition of the type specified by `name` along with all dependent type
...
+- `encoding`: `omgidl`
+- `data`: valid OMG IDL definition file contents with a definition for the message type and all referenced types.
+
+#### Producing `schema.data` for the `omgidl` schema encoding
```
amacneil, line 118: Would using the exact same format as `ros2idl` not work? Why do these differ?

james-rms: people not using ROS 2 don't have package resource names for their types, which the ros2idl spec depends on.

defunctzombie: I'd also add that if we can avoid the "ros2idl" format is required for ROS2 messages since there's no native way to have multiple messages in a single "schema". In IDL, to our knowledge, it is valid to have multiple modules and structs in a single file - so the approach proposed in this PR results in a schema file that is still a valid IDL schema file and requires no additional "separator" scheme.

docs/specification/appendix.md
snosenzo: Should this specify that this globally-scoped name should be in the `data` exactly and will be used as the root schema when decoding the message? Or was there another way we wanted to specify that?

docs/specification/appendix.md
```diff
@@ -107,6 +107,28 @@ The IDL definition of the type specified by `name` along with all dependent type
...
+
+For the `omgidl` schema encoding, the `data` field needs to contain all of those definitions in a single block of bytes. One way to produce this is to use a C++ compiler to resolve those `#include` directives:
+
+```bash
+# Run from the root of your IDL definition tree, or use additional `-I` arguments to indicate your
+# full include path to the `cpp` utility.
+# The `cpp` output may contain debugging lines such as "# 1 "my_company/my_module/MyType.idl",
+# the grep command here is intended to filter those out.
+$ cpp -I. my_company/my_module/MyType.idl | grep -v "^# \d" > all_definitions.idl
+```
```
snosenzo, line 130: example input and output as well?

docs/specification/appendix.md
emersonknapp: Maybe `my_company_global_module::my_specific_module::MyType`...?

website/docs/spec/registry.md
```diff
@@ -108,6 +108,28 @@ The IDL definition of the type specified by `name` along with all dependent type
...
+
+The `omgidl` schema encoding stores [OMG IDL](https://www.omg.org/spec/IDL/4.2/About-IDL) definitions for (X)CDR-encoded messages.
+
+- `name`: The globally-scoped name of the message type, eg. `my_company::my_module::MyType`
+- `encoding`: `omgidl`
+- `data`: valid OMG IDL definition file contents with a definition for the message type and all referenced types.
+
+#### Producing `schema.data` for the `omgidl` schema encoding
+
+Most commonly, IDL definitions for separate types are stored in separate files, and use C++ preprocessor `#include` statements to include definitions for types referenced in that definition.
```
emersonknapp, line 121:
```suggestion
Most commonly, IDL definitions for separate types are stored in separate files, and use the preprocessor directive `#include` to include definitions for types referenced in that definition.
```

For full precision I think this language is more correct. The IDL spec never mentions "C++" in the context of the preprocessor, it just defines what a "preprocessor directive" looks like for IDL. You could call them "C++-style" if you wanted to make the connection for users, but I don't think it's necessary - though I'd also note in that case that it's really just "C-style".

website/docs/spec/registry.md
```diff
@@ -108,6 +108,28 @@ The IDL definition of the type specified by `name` along with all dependent type
...
+
+For the `omgidl` schema encoding, the `data` field needs to contain all of those definitions in a single block of bytes. One way to produce this is to use a C++ compiler to resolve those `#include` directives:
```
emersonknapp, line 123:
```suggestion
For the `omgidl` schema encoding, the `data` field needs to contain all of those definitions in a single block of bytes. One way to produce this is to use a C preprocessor to resolve those `#include` directives:
```

Noting that `cpp` is not a compiler, it's the standalone gnu C preprocessor program. If a user tried `g++` or `c++` or  `clang++` as the command for above, it would not work because the compiler would try to read the `.idl` as  a linker file. Not sure how you would run this on windows, but that's probably fine.

emersonknapp (approved):
Some language nitpicks for clarity but looks good

rsanchez15 (approved):
LGTM

website/docs/spec/registry.md
james-rms: I believe the IDL spec specifically calls out C++, so i'm mirroring that here.
> IDL shall be preprocessed according to the specification of the preprocessor in ISO/IEC 14882:2003.
https://www.iso.org/standard/38110.html

emersonknapp: Ah - thanks for the info! I glossed over the ISO number when I read that section of the IDL spec.

defunctzombie:
👍 I'd like us to get an end-to-end working example with Studio prior to merging this.

willcbaker:
> +1 I'd like us to get an end-to-end working example with Studio prior to merging this.

@defunctzombie is there any guide to incorporating omgidl syntax in foxglove?

I have some xcdr encoded mcap files that I would like to view in studio and am willing to poke at this.

snosenzo:
> I have some xcdr encoded mcap files that I would like to view in studio and am willing to poke at this.

@willcbaker I'm currently working on getting `omgidl` support in studio and it should exist in a basic form by the end of the week. If you want to share your files (either via slack or email) I'd be happy to test against them and try to make sure they're supported. If sharing isn't possible feel free to let us know where we lack support after release.

PR #897, opened 2023-05-23
Website fixes

amacneil:
### Description
- Add a redirect for historical evaluation pdf
- Fix robot icon on homepage


jtbandes (approved):
Why 308 as opposed to 301 for all these?

amacneil:
> Why 308 as opposed to 301 for all these?

No real reason, they do the same thing. 308 requires clients to use the same http verb, but that doesn't make any difference in this case.

PR #898, opened 2023-05-29
CPP: add getter for metadata indices

emersonknapp:
So that I can browse metadata from indices in C++ client

### DIFF
--- a/cpp/mcap/include/mcap/reader.hpp
+++ b/cpp/mcap/include/mcap/reader.hpp
@@ -428,6 +428,12 @@ public:
    */
   const std::vector<ChunkIndex>& chunkIndexes() const;
 
+  /**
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

cpp/mcap/include/mcap/reader.hpp
```diff
@@ -428,6 +428,12 @@ class MCAP_PUBLIC McapReader final {
    */
   const std::vector<ChunkIndex>& chunkIndexes() const;
 
+  /**
+   * @brief Returns all of the parsed MetadataIndex records. Call `readSummary()`
+   * first to fully populate this data structure.
```
jtbandes, line 433: Include something to explain what the map keys represent?

Issue #899, opened 2023-05-30
"File Format Specification" link in top-level README is broken.


jkuszmaul:
**Description**

https://github.com/foxglove/mcap/blob/main/README.md?plain=1#L7 is wrong.

PR #900, opened 2023-05-30
fix link

james-rms:
### Description

Fixes https://github.com/foxglove/mcap/issues/899


amacneil:
“Motivation” link in the readme looks wrong too. 

Might be better to link to the website rather than GitHub for this. Has document outline etc.

emersonknapp (approved):

PR #904, opened 2023-06-06
Docs: rearrange record ordering more logically

wkalt:
In the description of the summary section, swaps the location of "metadata index" and "statistics". Files can be written in either order technically, but this is what our libraries do. Metadata indexes are also more logically grouped with attachment indexes.

In the listing of records, arranges the "data portion" records above the summary section records.


james-rms (approved):

emersonknapp (approved):
The reason for the initial ordering is increasing opcode, right? If this makes more sense for comprehension of file structure then it seems fine. The sidebar has TOC in order of opcode so it's easily discoverable in that way, correct?

PR #913, opened 2023-06-15
Add flake8-import-order linter

emersonknapp:
The specific style is a matter of preference - but I like to have this linter plugged in for some consistency/readability improvement. Open to feedback on exact configuration.

No functional change in the diff, just formatting.

### DIFF
--- a/python/.flake8
+++ b/python/.flake8
@@ -1,5 +1,6 @@
 [flake8]
 max-line-length = 100
+import-order-style = google
 exclude =
   **/build/**,
   examples/protobuf/*pb2.py,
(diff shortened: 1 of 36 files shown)


emersonknapp:
Matter of personal taste - I don't agree, I find it easier to find and modify within a single alphabetic list, rather than thinking or caring whether I `import X` or `from X import ...` - however there is a setting that does it the way you're requesting and I'm happy in the end to do whatever as long as it's consistent. Maybe we need a third opinion?

jtbandes:
Just to throw another idea out there — I've had good experiences using `black` as a formatter, which is highly opinionated and has very few configuration options.

emersonknapp:
`black` is already enabled for this repo, but from what I understand it doesn't have opinions about import ordering. There is `isort` which has a `black` profile https://pycqa.github.io/isort/docs/configuration/black_compatibility.html. That seems an acceptable alternative - I'm seeing what that output looks like.

emersonknapp:
Closing in favor of #914

PR #914, opened 2023-06-15
Add isort formatter for import ordering

emersonknapp:
Alternative to https://github.com/foxglove/mcap/pull/913

Maybe we like this standard better, it does the `from X import Y` statements after `import X`, and a couple other differences

More configuration options available at https://pycqa.github.io/isort/docs/configuration/options.html - I just did the default `black` profile here.


defunctzombie (approved):
such sort order 👍

PR #915, opened 2023-06-15
Support fast summary section rewrites

wkalt:
Changes the implementation of "mcap add attachment" and "mcap add metadata" to append to the data section in place, then overwrite an updated summary section with all offsets and CRCs updated.

This makes adding attachments and metadata very fast, and constant time with respect to size of the file. Prior to this commit, each one did an overwrite.

### DIFF
--- a/go/cli/mcap/utils/mcap_amendment.go
+++ b/go/cli/mcap/utils/mcap_amendment.go
@@ -1,0 +1,454 @@
+package utils
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"hash/crc32"
+	"io"
+
+	"github.com/foxglove/mcap/go/mcap"
+)
+
+const (
+	SizeRecordLength = 8
+	SizeOpcode       = 1
+	SizeDataEnd      = 4
+	SizeFooter       = 8 + 8 + 4
+	SizeMagic        = 8
+)
+
+// AmendMCAP adds attachment and metadata records to the end of the
+// data section of an existing MCAP, in place. It then writes a new summary
+// section consisting of the existing summary section, plus new attachment and
+// metadata index records as applicable, with all offsets and CRCs updated to
+// account for the new data.
+func AmendMCAP(
+	wsc io.ReadWriteSeeker,
+	attachments []*mcap.Attachment,
+	metadata []*mcap.Metadata,
+) error {
+	lexer, err := mcap.NewLexer(wsc, &mcap.LexerOptions{
+		SkipMagic: true,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to build lexer: %w", err)
+	}
+	defer lexer.Close()
+
+	oldFooter, err := readFooter(wsc)
+	if err != nil {
+		return fmt.Errorf("failed to read footer: %w", err)
+	}
+	oldSummaryStart := oldFooter.SummaryStart
+
+	// seek to the opcode of the data end record
+	oldDataEndOffset := int64(oldSummaryStart - SizeDataEnd - SizeRecordLength - SizeOpcode)
+	oldDataEnd, err := readDataEnd(wsc, oldDataEndOffset)
+	if err != nil {
+		return fmt.Errorf("failed to read data end: %w", err)
+	}
+
+	// Now we are at the old summary start. Parse from here to EOF into a
+	// summarySection structure.
+	summary, err := readSummarySection(wsc)
+	if err != nil {
+		return fmt.Errorf("failed to read summary section: %w", err)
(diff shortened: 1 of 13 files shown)


wkalt:
I have done this only in the CLI and kept changes to the writer minimal, rather than attempting to add more generic "append" functionality to the SDK. I think generic append is possible but definitely will require more invasive writer changes, and will also introduce questions around whether we should support it in all SDKs - which I don't think we want to bite off yet.

emersonknapp:
> constant time with respect to size of the file.

I got confused by this phrasing. Do you mean that it used to be linear to size of the file, but now it's a constant-time operation?

wkalt:
@emersonknapp yeah exactly. I can clarify the wording, when I say "overwrite" I think that's also not too clear. Previously doing these operations fully rewrote the file, including all the data. Now it just rewrites the summary section and updates all the accounting accordingly.

go/mcap/writer.go
```diff
@@ -520,7 +520,7 @@ func makePrefixedMap(m map[string]string) []byte {
 	return buf
 }
 
-func (w *Writer) writeChunkIndex(idx *ChunkIndex) error {
+func (w *Writer) WriteChunkIndex(idx *ChunkIndex) error {
```
james-rms, line 523: lets document the function now that it's public.

go/cli/mcap/utils/mcap_amendment.go
```diff
@@ -0,0 +1,454 @@
...
+		oldDataEnd.DataSectionCRC,
+		attachments,
+		metadata,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to build extra data: %w", err)
+	}
+
+	// the new summary offset is the old data end offset plus the length just written.
+	newSummaryOffset := oldDataEndOffset + int64(extraDataLength)
```
james-rms, line 77: i may answer this for myself as I read more - does writeExtraData include a data end record? if not, shouldn't this be:
```suggestion
	newSummaryOffset := oldDataEndOffset + int64(extraDataLength) + SIZE_DATA_END + SIZE_RECORD_LENGTH + SIZE_OPCODE
```

go/cli/mcap/utils/mcap_amendment.go
```diff
@@ -0,0 +1,454 @@
...
+		summary.AttachmentIndexes = append(summary.AttachmentIndexes, attachmentIndex)
+		summary.Statistics.AttachmentCount++
+	}
+	for _, metadataIndex := range metadataIndexes {
+		summary.MetadataIndexes = append(summary.MetadataIndexes, metadataIndex)
+		summary.Statistics.MetadataCount++
+	}
+
+	// Write summary out into a well-formed summary section, with all offsets and CRCs updated.
+	newSummaryData, err := buildSummaryBytes(*summary, newSummaryOffset)
```
james-rms, line 90: why do we build the summary in a buffer first then write it? wouldn't it be simpler to pass `wsc` to buildSummary~~Bytes~~?

go/cli/mcap/utils/mcap_amendment.go
james-rms: ok, looks like it does.

go/cli/mcap/utils/mcap_amendment.go
```diff
@@ -0,0 +1,454 @@
...
+	"encoding/binary"
+	"fmt"
+	"hash/crc32"
+	"io"
+
+	"github.com/foxglove/mcap/go/mcap"
+)
+
+const (
+	SIZE_RECORD_LENGTH = 8
```
james-rms, line 14:
```suggestion
	SizeRecordLength = 8
```
seems more in the go idiom for package-level constants?

go/cli/mcap/utils/mcap_amendment.go
```diff
@@ -0,0 +1,454 @@
...
+		return fmt.Errorf("failed to read summary section: %w", err)
+	}
+
+	// Seek to the location of the old data end record. We will write the new data section records starting here.
+	_, err = wsc.Seek(oldDataEndOffset, io.SeekStart)
+	if err != nil {
+		return fmt.Errorf("failed to seek to data end: %w", err)
+	}
+
+	extraDataLength, attachmentIndexes, metadataIndexes, err := writeExtraData(
```
james-rms, line 65: I might have named this `extendDataSection`, which feels more like it would add a DataEnd record than `writeExtraData`. I defer to your taste though.

james-rms (approved):
Looks great!

wkalt:
comments addressed

PR #916, opened 2023-06-15
demo: show camera immediately when enabled

jtbandes:
### Public-Facing Changes

Camera feed now appears immediately when "Camera" is toggled on, even if recording has not started yet.


narasaka (approved):
noice 🫡

PR #919, opened 2023-07-05
Tighten python protobuf dependency to `<=4.21.11` to fix potential segfault

achim-k:
### Public-Facing Changes

Tighten python protobuf dependency to `<=4.21.11` to fix potential segfault

### Description
Reading an mcap file with `mcap-protobuf-support` can lead to a segfault which is very likely related to https://github.com/protocolbuffers/protobuf/issues/12047 (see https://github.com/protocolbuffers/protobuf/issues/12047#issuecomment-1622405787 for reproduction steps). This patch tightens the python protobuf dependency to `<=4.21.11` as the segfault apprently only happens from `4.21.11` onwards.

Fixes FG-3792

### DIFF
--- a/python/mcap-protobuf-support/setup.cfg
+++ b/python/mcap-protobuf-support/setup.cfg
@@ -13,7 +13,7 @@ classifiers =
 [options]
 install_requires = 
     mcap>=0.0.14
-    protobuf<=4.21.11
+    protobuf>=3.8,<=4.21.11
 install_package_data = True
 packages = find:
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
Should we also add `>3` or something to ensure a minimum compatible version?

achim-k:
> Should we also add `>3` or something to ensure a minimum compatible version?

Done, `3.8.0` is the lowest working version.

PR #922, opened 2023-07-10
spec: add channel UUID

james-rms:
### Public-Facing Changes

* Resolves ambiguity around extending records with new fields. Records may be extended by changes to the specification only.
* Defines zero values for all field types, to be used when a field is missing from a serialized record.
* Adds a UUID field to the Channel record. This allows readers to safely treat channels in different files as the same logical channel.

### DIFF
--- a/website/docs/spec/index.md
+++ b/website/docs/spec/index.md
@@ -111,7 +111,7 @@ All MCAP records are serialized as follows:
 
 Record type is a single byte opcode, and record content length is a uint64 value.
 
-Records may be extended by adding new fields at the end of existing fields. Readers should ignore any unknown fields.
+Future changes to this specification may extend records by adding new fields at the end of existing fields. Readers should ignore any unknown fields.
 
 > The Footer and Message records will not be extended, since their formats do not allow for backward-compatible size changes.
 
--- a/website/docs/spec/index.md
+++ b/website/docs/spec/index.md
@@ -117,6 +117,8 @@ Future changes to this specification may extend records by adding new fields at 
 
 Each record definition below contains a `Type` column. See the [Serialization](#serialization) section on how to serialize each type.
 
+Record content may end before all known fields have been serialized. Readers should treat a missing field as having that field type's [zero value](#zero-values).
+
 ### Header (op=0x01)
 
 | Bytes | Name    | Type   | Description                                                                                                                                                                                                                                                                                                              |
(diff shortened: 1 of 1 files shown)


website/docs/spec/index.md
```diff
@@ -155,13 +157,17 @@ A Channel record defines an encoded stream of messages on a topic.
...
+| 2     | schema_id        | uint16                | The schema for messages on this channel. A schema_id of 0 indicates there is no schema for this channel.                                                                       |
+| 4 + N | topic            | String                | The channel topic.                                                                                                                                                             |
+| 4 + N | message_encoding | String                | Encoding for messages on this channel. The [well-known message encodings][message_encodings] are preferred.                                                                    |
+| 4 + N | metadata         | `Map<string, string>` | Metadata about this channel                                                                                                                                                    |
+| 16    | uuid             | uuid                  | A globally unique identifier for this channel. A [nil UUID](https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.7) indicates no globally unique identifier is available. |
```
amacneil, line 170: seems like this could be solved by using channel metadata. is there something that makes this special enough to be an explicitly named field?

website/docs/spec/index.md
```diff
@@ -111,12 +111,14 @@ All MCAP records are serialized as follows:
 
 Record type is a single byte opcode, and record content length is a uint64 value.
 
-Records may be extended by adding new fields at the end of existing fields. Readers should ignore any unknown fields.
+Future changes to this specification may extend records by adding new fields at the end of existing fields. Readers should ignore any unknown fields.
```
amacneil, line 114: if we are introducing this language it would be good to do in a separate PR, and ensure we have conformance tests for readers well ahead of introducing any new fields.

website/docs/spec/index.md
```diff
@@ -111,12 +111,14 @@ All MCAP records are serialized as follows:
...
 
 > The Footer and Message records will not be extended, since their formats do not allow for backward-compatible size changes.
 
 Each record definition below contains a `Type` column. See the [Serialization](#serialization) section on how to serialize each type.
 
+Record content may end before all known fields have been serialized. Readers should treat a missing field as having that field type's [zero value](#zero-values).
```
amacneil, line 120: if we are going to make this statement, we should probably be explicit that records must still be fully formed, even if they are missing fields (i.e. records must match their stated length). otherwise the file should be considered corrupt.

website/docs/spec/index.md
```diff
@@ -319,22 +325,33 @@ A Summary Offset record contains the location of records within the summary sect
...
 
+The [zero value](#zero-values) of all integer types is 0.
+
 ### String
 
 Strings are serialized using a `uint32` byte length followed by the string data, which should be valid [UTF-8](https://en.wikipedia.org/wiki/UTF-8).
 
     <byte length><utf-8 bytes>
 
+The [zero value](#zero-values) of a string is the empty string.
```
amacneil, line 345:
```suggestion
The [zero value](#zero-values) of a string is an empty string (i.e. a single `0x00` byte).
```

website/docs/spec/index.md
```diff
@@ -353,6 +370,8 @@ Example `Tuple<uint16, string>`:
 
     <uint16><uint32><utf-8 bytes>
 
+The [zero value](#zero-values) of a tuple is the first value type's zero value followed by the second value type's zero value.
```
amacneil, line 373: do tuples ever occur outside of an array? I don't think this case is needed.

website/docs/spec/index.md
```diff
@@ -365,6 +384,8 @@ An array of uint64 is specified as `Array<uint64>` and serialized as:
 
 > Since arrays use a `uint32` byte length prefix, the maximum size of the serialized array elements cannot exceed 4,294,967,295 bytes.
 
+The [zero value](#zero-values) of an array is the empty array.
```
amacneil, line 387:
```suggestion
The [zero value](#zero-values) of an array is an array with length zero (i.e. `0x00000000`).
```

website/docs/spec/index.md
defunctzombie: Being a first class citizen means that tooling can rely on this field. Metadata is better read as "user-data" that is unspecified.

amacneil: Do we have information about such a use case? What if I use something other than UUID to identify channels? 

I think we should have hard coded fields where there is a demonstrated need for cross-tool interoperability. If this is only to support needs of individual companies then using a named metadata field would be sufficient.

defunctzombie: I am not convinced that UUID is the appropriate term or requirement. I would describe it is an identifier that is stable across multiple separate mcap files (possibly via rolling recording).

I think this PR needs to be split up into separate proposals that have clear problem statements. Right now its not in a state that makes a case on what problems it is solving.

PR #928, opened 2023-07-21
fix python reader support for mcap without summary

omegacoleman:
When attempting to "proxy" a generator function in python, using `return gen()` is wrong. `return val` constructs a StopIteration exception with `val` and raise it, not doing what you want.

This PR fix this by using `yield from`, also adds a test to secure the no summary mcap processing logic.

### DIFF
--- a/python/mcap/tests/test_reader.py
+++ b/python/mcap/tests/test_reader.py
@@ -234,5 +234,7 @@ def test_no_summary_not_seeking(tmpdir: Path):
 
     with open(filepath, "rb") as f:
         assert len(list(NonSeekingReader(f).iter_messages())) == 200
+    with open(filepath, "rb") as f:
         assert len(list(NonSeekingReader(f).iter_attachments())) == 1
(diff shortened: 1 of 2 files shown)


python/mcap/tests/test_reader.py
```diff
@@ -195,3 +195,35 @@ def test_non_seeking_used_once():
...
+        writer.finish()
+
+
+@pytest.mark.parametrize("reader_cls", READER_SUBCLASSES)
+def test_no_summary(reader_cls: AnyReaderSubclass, tmpdir: Path):
+    filepath = tmpdir / "no_summary.mcap"
+    write_no_summary_mcap(filepath)
+
+    with open(filepath, "rb") as f:
+        reader: McapReader = SeekingReader(f)
```
jtbandes, line 226:
```suggestion
        reader: McapReader = reader_cls(f)
```

jtbandes (approved):
Thank you! Looks great

PR #930, opened 2023-07-25
Log time0

### LINKED ISSUE #924: rust writer does not track log_time 0 in message_start_time
Due to this code, any messages with log_time 0 will not be accounted for in the message_start_time of the chunk: https://github.com/foxglove/mcap/blob/6ffcfa0ba19155d425d034ca92d9f8ca973c2137/rust/src/write.rs#L283-L286

mrkline:
Fixes #924, removes some duplicated stats tracking data in the Chunk and overall MCAP writer while we're at it.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -7,7 +7,7 @@ categories = [ "science::robotics", "compression" ]
 repository = "https://github.com/foxglove/mcap"
 documentation = "https://docs.rs/mcap"
 readme = "README.md"
-version = "0.7.0"
+version = "0.7.1"
 edition = "2021"
 license = "MIT"
 
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -172,10 +172,12 @@ pub struct Writer<'a, W: Write + Seek> {
     options: WriteOptions,
     schemas: HashMap<Schema<'a>, u16>,
     channels: HashMap<Channel<'a>, u16>,
-    stats: records::Statistics,
     chunk_indexes: Vec<records::ChunkIndex>,
     attachment_indexes: Vec<records::AttachmentIndex>,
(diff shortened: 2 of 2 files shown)


defunctzombie:
Not sure how the tests are organized but is there an existing test for checking this behavior? Or is this code tested with the conformance tests?

defunctzombie (approved):

rust/Cargo.toml
```diff
@@ -7,7 +7,7 @@ categories = [ "science::robotics", "compression" ]
 repository = "https://github.com/foxglove/mcap"
 documentation = "https://docs.rs/mcap"
 readme = "README.md"
-version = "0.7.0"
+version = "0.7.1"
```
defunctzombie, line 10: Will we need to do anything to release this version to crates?

defunctzombie: The only previous instance of a "release" I've found is this tag: https://github.com/foxglove/mcap/releases/tag/releases%2Frust%2Fv0.6.0

mrkline: @james-rms is the owner of the crate (see https://crates.io/crates/mcap) so he'll have to push an update with [`cargo publish`](https://doc.rust-lang.org/cargo/reference/publishing.html) unless someone has his credentials sitting around. :stuck_out_tongue:

mrkline:
The existing unit tests should verify that file contents for the test outputs haven't changed (see https://github.com/foxglove/mcap/tree/main/rust/tests/data). A specific test that confirms #924 is gone wouldn't be a bad idea; I can add one later tonight.

mrkline:
Test added. Looks like there's some spurious failure about an API token in the `docs-deploy` CI step.

defunctzombie (approved):

PR #940, opened 2023-08-02
Add og-image for mcap.dev site

esthersweon:
### Public-Facing Changes

- Add og-image for mcap.dev site


bryfox (approved):
Looks correct. Presumably twitter et al post-process this as needed, so we could use a higher-res version with less compression if we wanted?

PR #948, opened 2023-08-22
cpp: fix warning in examples on msvc

pezy:
### Public-Facing Changes

None

### Description

- remove unused include
- use explicit type conversion
- ignore C4267 and C4244 for 3rd_party protobuf generate codes
- prefer constexpr

### DIFF
--- a/cpp/examples/bag2mcap.cpp
+++ b/cpp/examples/bag2mcap.cpp
@@ -30,8 +30,8 @@ int main() {
   mcap::Channel topic("/chatter", "ros1", stdMsgsString.id);
   writer.addChannel(topic);
 
-  std::array<std::byte, 4 + 13> payload;
-  const uint32_t length = 13;
+  std::array<std::byte, 4 + 13> payload{};
+  constexpr uint32_t length = 13;
   std::memcpy(payload.data(), &length, 4);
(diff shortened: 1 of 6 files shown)


cpp/examples/bag2mcap.cpp
```diff
@@ -30,8 +30,8 @@ int main() {
   mcap::Channel topic("/chatter", "ros1", stdMsgsString.id);
   writer.addChannel(topic);
 
-  std::array<std::byte, 4 + 13> payload;
-  const uint32_t length = 13;
+  std::array<std::byte, 4 + 13> payload{};
```
james-rms, line 33: does this line change fix a warning?

pezy: > does this line change fix a warning?

Specifically, this is not a warning, but according to the [cpp core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Pro-type-memberinit) recommendation: Don't use a variable before it has been initialized

james-rms (approved):
happy to merge. I'm noting here that support for these examples on Windows is not guaranteed and is best-effort, because we don't test them in CI.

PR #950, opened 2023-08-23
Go: Allocate less in indexed message chunk loading

wkalt:
Prior to this commit, when reading a chunk into memory we would read it (internally, in io.ReadAll) into a resizing buffer and need to resize it several times up to final size. Since we know the final size in advance we can preallocate and save those allocations/copies.

### DIFF
--- a/go/mcap/indexed_message_iterator.go
+++ b/go/mcap/indexed_message_iterator.go
@@ -162,18 +162,14 @@ func (it *indexedMessageIterator) loadChunk(chunkIndex *ChunkIndex) error {
 	case CompressionNone:
 		chunkData = parsedChunk.Records
 	case CompressionZSTD:
-		var err error
-		if it.zstdDecoder == nil {
-			it.zstdDecoder, err = zstd.NewReader(bytes.NewReader(parsedChunk.Records))
-		} else {
-			err = it.zstdDecoder.Reset(bytes.NewReader(parsedChunk.Records))
-		}
+		it.zstdDecoder, err = zstd.NewReader(nil)
 		if err != nil {
 			return fmt.Errorf("failed to read zstd chunk: %w", err)
 		}
-		chunkData, err = io.ReadAll(it.zstdDecoder)
+		chunkData = make([]byte, 0, parsedChunk.UncompressedSize)
+		chunkData, err = it.zstdDecoder.DecodeAll(parsedChunk.Records, chunkData)
 		if err != nil {
-			return fmt.Errorf("failed to decompress zstd chunk: %w", err)
+			return fmt.Errorf("failed to decode chunk data: %w", err)
 		}
 	case CompressionLZ4:
 		if it.lz4Reader == nil {
--- a/go/mcap/indexed_message_iterator.go
+++ b/go/mcap/indexed_message_iterator.go
@@ -34,6 +34,8 @@ type indexedMessageIterator struct {
 	zstdDecoder           *zstd.Decoder
 	lz4Reader             *lz4.Reader
 	hasReadSummarySection bool
+
+	compressedChunk []byte
 }
 
 // parseIndexSection parses the index section of the file and populates the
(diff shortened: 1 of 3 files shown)


wkalt:
simple timings

main
```
[~/work/mcap/go/cli/mcap] (main) $ time mcap cat cal_loop.mcap > /dev/null
real    0m5.047s
user    0m4.719s
sys     0m1.083s
```

patch
```
[~/work/mcap/go/cli/mcap] (ticket/fg-4715/reader-less-allocs-go) $ time mcap cat cal_loop.mcap > /dev/null

real    0m3.084s
user    0m2.716s
sys     0m0.532s

```

40% faster

wkalt:
lz4 numbers

main
```
[~/work/mcap/go/cli/mcap] (main) $ time mcap cat cal_loop.lz4.mcap > /dev/null

real    0m3.802s
user    0m3.563s
sys     0m0.622s
```

patch

```
[~/work/mcap/go/cli/mcap] (ticket/fg-4715/reader-less-allocs-go) $ time mcap cat cal_loop.lz4.mcap > /dev/null

real    0m2.372s
user    0m2.151s
sys     0m0.354s
```

wkalt:
it seems like there would be further savings if we could reuse the chunk buffers, but I think this is too complicated by the heap. I also noticed the Next() method still takes a buffer which is now unused (I think originally that was passed to the lexer's Next method to conserve allocations) but I think resurrecting it is complicated now. Maybe we can sunset it, or maybe it actually is used for the unindexed iterator (need to check).

wkalt:
last commit shaves a bit more off

```
[~/work/mcap/go/cli/mcap] (ticket/fg-4715/reader-less-allocs-go) $ time mcap cat cal_loop.mcap > /dev/null

real    0m2.734s
user    0m2.527s
sys     0m0.286s
```

wkalt:
with final commit, I got a
```
[~/work/mcap/go/cli/mcap] (ticket/fg-4715/reader-less-allocs-go) $ time mcap cat cal_loop.mcap > /dev/null

real    0m2.515s
user    0m2.374s
sys     0m0.219s
```

which is close to 50% better than the first observation

go/mcap/indexed_message_iterator.go
```diff
@@ -162,34 +168,33 @@ func (it *indexedMessageIterator) loadChunk(chunkIndex *ChunkIndex) error {
...
+		_, err = io.ReadFull(it.lz4Reader, chunkData)
 		if err != nil {
 			return fmt.Errorf("failed to decompress lz4 chunk: %w", err)
 		}
 	default:
 		return fmt.Errorf("unsupported compression %s", parsedChunk.Compression)
 	}
 	// use the message index to find the messages we want from the chunk
-	messageIndexSection := chunk[chunkIndex.ChunkLength:]
+	messageIndexSection := it.compressedChunk[chunkIndex.ChunkLength:compressedChunkLength]
```
wkalt, line 197: the naming is very confusing here - maybe we should change "compressedChunkLength" to "compressedChunkPlusIndexSectionLength".

wkalt: also clearly "compressedChunk" is not the compressed chunk record, it's the chunk plus the after-chunk data

go/mcap/indexed_message_iterator.go
```diff
@@ -147,12 +149,16 @@ func (it *indexedMessageIterator) loadChunk(chunkIndex *ChunkIndex) error {
 	if err != nil {
 		return err
 	}
-	chunk := make([]byte, chunkIndex.ChunkLength+chunkIndex.MessageIndexLength)
-	_, err = io.ReadFull(it.rs, chunk)
+
+	compressedChunkLength := chunkIndex.ChunkLength + chunkIndex.MessageIndexLength
+	if len(it.compressedChunk) < int(compressedChunkLength) {
+		it.compressedChunk = make([]byte, int(float64(compressedChunkLength)*1.2))
```
wkalt, line 155: .2 for luck

wkalt: should pull this out into a constant

jtbandes (approved):

wkalt:
there is still some preexisting issue with very large inputs, where reading becomes sluggish after a while. I need to debug it further to understand what is happening. Won't merge this til I understand better.

wkalt:
ok I think ^^ is just my terminal emulator, I verified with pv there's no actual slowdown and if I use xterm it appears to read at a consistent speed too. Merging.

PR #960, opened 2023-08-30
swift: flush writer buffer when closing a chunk

jtbandes:
### Public-Facing Changes

Swift: data is now properly sent to the writer each time a chunk is closed.

### DIFF
--- a/swift/mcap/MCAPWriter.swift
+++ b/swift/mcap/MCAPWriter.swift
@@ -161,6 +161,9 @@ public final class MCAPWriter {
       }
     } else {
       message.serialize(to: &buffer)
+      if buffer.count >= options.chunkSize {
+        await _flush()
+      }
     }
   }
 
@@ -219,6 +222,7 @@ public final class MCAPWriter {
 
     let chunkStartOffset = _position()
     chunk.serialize(to: &buffer)
+    await _flush()
     let messageIndexStartOffset = _position()
 
     var messageIndexOffsets: [ChannelID: UInt64] = [:]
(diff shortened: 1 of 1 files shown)


wkalt (approved):

swift/mcap/MCAPWriter.swift
```diff
@@ -161,6 +161,9 @@ public final class MCAPWriter {
       }
     } else {
       message.serialize(to: &buffer)
+      if buffer.count >= options.chunkSize {
```
jtbandes, line 164: Deceptively, this is actually for un-chunked writing – I thought I might as well use the chunkSize variable to decide when to flush, since it was there. Another option would be that we expose `flush` as a public method.

wkalt: I have been resistant to exposing a public flush method in the go lib. I think we may in rust. I don't really understand the connection here though. If this feature is about unchunked writing why not just do `buffer.count >= 0`? caveat: I haven't properly

wkalt: read the code... finished writing the sentence...?

jtbandes: Then it would be flushing after every message - is that desirable? Maybe it's fine?

jtbandes: To be clear, the [other line](https://github.com/foxglove/mcap/pull/960/files#diff-99aedf66b1cf439c9026d4dde15f49199f62897fda27c8d74a9658e13369ffa6R225) is necessary to ensure we flush when closing a chunk – this part is additionally adding a flush for non-chunked mode

PR #961, opened 2023-09-01
[Rust - bugfix] read_record_from_slice checking wrong size of buffer wrong

dmweis:
### Public-Facing Changes

There should be no changes in public API caused by this change.

### Description

`read_record_from_slice` was checking that buffer size is bigger than 5 but was trying to read a `u8` and `u64` which are in total 9 bytes. This was causing panics when reading malformed mcap files or otherwise malformed streams.

I added some simple tests as well but they aren't very robust (I don't check parsing of records).


jtbandes (approved):
Looks correct to me, adding @james-rms for a second sanity check.

james-rms (approved):

PR #964, opened 2023-09-04
Fix ROS 2 message equality check

### LINKED ISSUE #959: ROS2 decoded messages `__eq__` always returns true
**Description**
When checking the equality two messages decoded by `mcap_ros2.decoder.DecoderFactory`, the messages will compare equal even if their contents are different. I think this is because `__eq__` is comparing `__dict__` which is empty for these decoded messages. They use `__slots__` instead.

- Version: mcap 1.1.0; mcap-ros2-support 0.5.0
- Platform: Ubuntu 20.04

**Steps To Reproduce**

Place this in `mcap/python/mcap-ros2-support/tests/test_ros2_decoder.py` and run with pytest:
```python
def test_ros2_decoder_msg_eq():
    with generate_sample_data() as m:
        reader = make_reader(m, decoder_factories=[DecoderFactory()])

        decoded_messages = reader.iter_decoded_messages("/chatter")
        _, _, _, msg0 = next(decoded_messages)
        _, _, _, msg1 = next(decoded_messages)
        assert msg0.data == "string message 0"
        assert msg1.data == "string message 1"
        print("__dict__:", msg0.__dict__)
        print("__slots__:", msg0.__slots__)

        # Uncomment to show that __eq__ is comparing __dict__
        # msg0.__dict__["foo"] = ["bar"]

        assert msg0 != msg1
```

**Expected Behavior**
The test passes. Here the test does not pass because `msg0 == msg`.

achim-k:
### Public-Facing Changes

Fix ROS 2 message equality check

### Description
Use a custom comparison function that compares messages by their `__slots__` fields.

Fixes #959
Resolves FG-4775


jtbandes (approved):
Seems fine. I am not an expert in `__slots__`. For reference, I believe this is the original implementation which this dynamic generation code was adapted from:

https://github.com/ros/genpy/blob/f70658f2200dfb7c98e99e9cf15be61cfaa1a432/src/genpy/message.py#L419

PR #968, opened 2023-09-08
Go: add support for custom compression

wkalt:
Adds support for bringing custom compressors and decompressors to the go lexer and writer.

For the writer, a ResettableWriteCloser is accepted. For the reader, a map of compression format to ResettableReader is accepted. If the reader implements io.Closer we'll call that on file close.

### DIFF
--- a/go/mcap/Makefile
+++ b/go/mcap/Makefile
@@ -1,5 +1,5 @@
 test:
-	go test -cover ./...
+	go test ./...
 
 bench:
 	go test -benchmem -run=^$$ -count 5 -bench ^BenchmarkLexer/demo -memprofile mem.out -cpuprofile cpu.out
--- a/go/mcap/Makefile
+++ b/go/mcap/Makefile
@@ -1,5 +1,5 @@
 test:
-	go test ./...
+	go test -cover ./...
 
 bench:
 	go test -benchmem -run=^$$ -count 5 -bench ^BenchmarkLexer/demo -memprofile mem.out -cpuprofile cpu.out
--- a/go/mcap/resettable.go
+++ b/go/mcap/resettable.go
@@ -1,0 +1,15 @@
+package mcap
+
+import "io"
+
+// ResettableWriteCloser implements io.WriteCloser and adds a Reset method.
+type ResettableWriteCloser interface {
+	io.WriteCloser
+	Reset(io.Writer)
+}
+
+// ResettableReadCloser implements io.ReadCloser and adds a Reset method.
+type ResettableReader interface {
+	io.Reader
+	Reset(io.Reader)
+}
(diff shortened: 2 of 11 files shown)


go/mcap/Makefile
```diff
@@ -1,5 +1,5 @@
 test:
-	go test ./...
+	go test -cover ./...
```
james-rms, line 2: remove from this PR?

wkalt: done

go/mcap/resettable.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "io"
+
+// ResettableWriteCloser implements io.WriteCloser and adds a Reset method.
+type ResettableWriteCloser interface {
+	io.WriteCloser
+	Reset(io.Writer)
+}
+
+// ResettableReadCloser implements io.ReadCloser and adds a Reset method.
```
wkalt, line 11: io.Reader

go/mcap/resettable.go
```diff
@@ -0,0 +1,15 @@
+package mcap
+
+import "io"
+
+// ResettableWriteCloser implements io.WriteCloser and adds a Reset method.
+type ResettableWriteCloser interface {
+	io.WriteCloser
+	Reset(io.Writer)
```
wkalt, line 8: perhaps Reset should return error? lz4's does not, but zstd's does.

go/mcap/writer.go
```diff
@@ -833,20 +837,26 @@ func NewWriter(w io.Writer, opts *WriterOptions) (*Writer, error) {
 	compressed := bytes.Buffer{}
 	var compressedWriter *countingCRCWriter
 	if opts.Chunked {
-		switch opts.Compression {
-		case CompressionZSTD:
+		switch {
+		case opts.Compressor != nil: // must be top
+			if opts.Compression == "" {
```
bryfox, line 842: My first reaction is that the interface feels a little strange with 'compressor' and 'compression' decoupled. I don't see where `Compression` is used in the case of a custom compressor, but presumably an lz4 writer should always be accompanied by type LZ4. Is `Compression` required if a compressor is passed? If so, would it make sense to have `Compressor` contain the compression, and embed the ResettableWriteCloser type?

bryfox (commented):
Just a small question / food for thought, but this seems good to me.

go/mcap/writer.go
wkalt: it's an option, but what I don't really like about it is it forces all consumers of this to implement their own Compressor type (to expose a `Compression() string` method), whereas the current interface has a chance of matching the compressors out of the box - although I think this will only be true of one of lz4 and zstd, depending on whether or not Reset() is chosen to send an error out.

wkalt: but we can't make that change here unfortunately

james-rms (approved):
I haven't looked closely at the implementation but I approve of this configuration approach.

james-rms (commented):
Don't forget to bump the library version unless you plan to handle this in a follow up PR.

PR #969, opened 2023-09-08
fix: LinearMessageView begin does not compare equal to begin

james-rms:
### Public-Facing Changes

Multiple iterators created from LinearMessageView::begin() will compare to `true`.

### DIFF
--- a/cpp/mcap/include/mcap/reader.inl
+++ b/cpp/mcap/include/mcap/reader.inl
@@ -1588,8 +1588,7 @@ LinearMessageView::Iterator LinearMessageView::begin() {
   if (dataStart_ == dataEnd_ || !mcapReader_.dataSource()) {
     return end();
   }
-  return LinearMessageView::Iterator{mcapReader_, dataStart_, dataEnd_, readMessageOptions_,
-                                     onProblem_};
+  return LinearMessageView::Iterator{*this};
 }
 
 LinearMessageView::Iterator LinearMessageView::end() {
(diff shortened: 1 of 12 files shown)


cpp/test/unit_tests.cpp
```diff
@@ -476,6 +476,37 @@ TEST_CASE("McapReader::readMessages()", "[reader]") {
...
+    WriteMsg(writer, channel.id, 0, 2, 1, data);
+    WriteMsg(writer, channel.id, 1, 4, 3, data);
+    writer.close();
+
+    mcap::McapReader reader;
+    requireOk(reader.open(buffer));
+
+    auto view = reader.readMessages();
+    auto it = view.begin();
+    REQUIRE(it != view.end());
```
jtbandes, line 499: Could we add another test with an empty mcap and check that `view.begin() == view.end()`?

cpp/mcap/include/mcap/reader.inl
```diff
@@ -1752,7 +1742,17 @@ void LinearMessageView::Iterator::operator++(int) {
...
+    // special case for Iterator::end() == Iterator::end()
+    return a.impl_ == b.impl_;
+  }
+  if (!a.begun_ && !b.begun_) {
+    // special case for Iterator::begin() == Iterator::begin()
+    // comparing iterators to the beginning of the same view should return true.
+    return &(a.impl_->view_) == &(b.impl_->view_);
+  }
+  // all other cases return false.
+  return false;
```
jtbandes, line 1755: Does this mean an iterator does not `==` itself (not reflexive) at least in the case of `begin() != begin()`? That seems potentially problematic

jtbandes (approved):

PR #970, opened 2023-09-10
Rust: enable write conformance tests

james-rms:


defunctzombie:
Has this PR gone stale? Do we want to keep it as reference or revisit when we plan to do this?

james-rms:
Very stale - we need to re-vamp the writer so that it can write data section CRCs before doing this. I've left this PR as a reminder, will close it and write up a linear project for this.

PR #971, opened 2023-09-11
Python: Detect invalid magic when instantiating SeekingReader

achim-k:
### Public-Facing Changes

Python: Detect invalid magic when instantiating `SeekingReader`

### Description
Attempts to read the magic bytes when instantiating a `SeekingReader`, leading to an exception when the stream is not from a valid mcap file

Resolves FG-4787


python/mcap/tests/test_reader.py
```diff
@@ -238,3 +238,17 @@ def test_no_summary_not_seeking(tmpdir: Path):
...
+    with open(filepath, "w") as f:
+        f.write("some bytes longer than the initial magic bytes")
+
+    with open(filepath, "rb") as f:
+        with pytest.raises(InvalidMagic):
+            SeekingReader(f)
+
+    with open(filepath, "rb") as f:
+        with pytest.raises(InvalidMagic):
+            NonSeekingReader(f).get_header()
```
jtbandes, line 254: What if you use a NonSeekingReader but don’t explicitly ask for the header?

jtbandes (approved):

python/mcap/tests/test_reader.py
achim-k: The magic check only happens when [iterating over the records](https://github.com/foxglove/mcap/blob/8ac076a0bcd6a35ca7cdfc9dbddc345a2d5191d8/python/mcap/mcap/stream_reader.py#L104-L109) (which is what `get_header()` does). So just instantiating would not raise the invalid magic exception. Do we want to change this?

jtbandes: I think that's fine, as long as it does happen sometime

PR #973, opened 2023-09-14
Add progress bar for `mcap convert`

narasaka:
### Public-Facing Changes

This adds a progress bar when running `mcap convert ...`


go/ros/bag2mcap.go
```diff
@@ -225,7 +232,7 @@ func channelIDForConnection(connID uint32) (uint16, error) {
 	return uint16(connID), nil
 }
 
-func Bag2MCAP(w io.Writer, r io.Reader, opts *mcap.WriterOptions) error {
+func Bag2MCAP(w io.Writer, r io.Reader, opts *mcap.WriterOptions, progressBar *progressbar.ProgressBar) error {
```
narasaka, line 235: I feel like i should just put the progress bar in this function instead of injecting it 🤷

wkalt:
This looks great generally. The thing that's not clear to me is this:

```
 100% |████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| (1606/1606, 4395 it/s)

```

the 1606 needs some units, and I have no idea what "it/s" is. I guess that's the conversion rate? We can probably either remove that, or change it to a MB/sec computation.

go/ros/bag2mcap.go
wkalt: that does seem nicer, but you'd just need to pass in a messageCount.

Maybe the cleanest option would be to pass in the `*Info` struct, and then construct the progress bar in here? You need to call Info() outside this function (as you do) or else you would need to pass this a ReadSeeker, which you're not in position to do.

go/ros/ros2db3_to_mcap.go
```diff
@@ -263,6 +264,12 @@ func DB3ToMCAP(w io.Writer, db *sql.DB, opts *mcap.WriterOptions, searchdirs []s
 	if err != nil {
 		return err
 	}
+
+	progressBar := progressbar.NewOptions64(
+		int64(len(topics)),
```
wkalt, line 269: what is len(topics) doing here?

narasaka: isn't number of topics the denominator of the progress?

go/ros/bag2mcap.go
```diff
@@ -173,6 +173,10 @@ func processBag(
 			}
 		}
 
+		if err != nil {
+			return err
+		}
```
wkalt, line 178: remove

go/ros/ros2db3_to_mcap.go
```diff
@@ -313,6 +319,10 @@ func DB3ToMCAP(w io.Writer, db *sql.DB, opts *mcap.WriterOptions, searchdirs []s
 		if err != nil {
 			return fmt.Errorf("failed to write channel info: %w", err)
 		}
+
+		if err != nil {
+			return fmt.Errorf("failed to increment progress bar: %w", err)
+		}
```
wkalt, line 325: remove

go/go.work
```diff
@@ -1,4 +1,6 @@
-go 1.18
+go 1.21
```
wkalt, line 1: 1.18

wkalt (approved):

PR #975, opened 2023-09-18
Python: Handle ros2 uint8[] field as str without ValueError

pezy:
### Public-Facing Changes

None

### Description

When calling the `write_message` method, the message argument often originates from some other serialized form, like `Protobuf` or `JSON`. However, bytes fields can become mismatched during these serialization processes. For instance, even though a Protobuf message may contain bytes fields, using `MessageToDict` to convert it to a Python dictionary will base64 encode those bytes, as it adheres to the [Proto3 JSON mapping guidelines](https://protobuf.dev/programming-guides/proto3/#json).

This leads to a confusing situation for the user. They might assume that the field is still in `bytes` and defined as `uint8[]` in the ROS2 message. But when invoking `write_message`, they encounter the following cryptic error:
```
ValueError: Field "data" is not an array (<class 'str'>) but has array type "uint8[]"
```
While the error is technically correct, it's confusing for users who are not aware of the limitations of JSON or the transformations performed by `MessageToDict`.

This pull request aims to alleviate this issue by adding an additional check to see if the problematic field is base64-encoded. If it is, we decode it and proceed without errors, as demonstrated in `test_write_bytes_object_from_json_deserialization`. If decoding fails, the function will still raise the original error.

### DIFF
--- a/python/mcap-ros2-support/mcap_ros2/_dynamic.py
+++ b/python/mcap-ros2-support/mcap_ros2/_dynamic.py
@@ -2,6 +2,8 @@
 
 import os
 import re
+import base64
+import binascii
 from io import BytesIO
 from types import SimpleNamespace
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


james-rms (commented):
Thanks for the PR. I believe this change might hide encoding bugs that a user would rather fix in their client code, so I'm against merging this upstream. As an aside, since this is the mcap-ros2-support package, we don’t want any protobuf-specific logic in there and we don’t expect/assume users are using protobuf.

Since this change helps you, you're free to keep it in your own fork of mcap-ros2-support or move it to a helper function in your code that processes messages before calling writer.write_message.

PR #982, opened 2023-09-28
CLI/merge: Avoid duplication of identical schemas

achim-k:
### Public-Facing Changes

CLI/merge: Avoid duplication of identical schemas

### Description
- Removes some unused members of the McapMerger struct
- Avoid duplication of identical schemas when merging mcap files. Schemas are considered identical if their md5 hash sum over the schema name, encoding and data is equal.

### DIFF
--- a/go/cli/mcap/cmd/merge.go
+++ b/go/cli/mcap/cmd/merge.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"container/heap"
+	"crypto/md5"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
@@ -43,6 +45,7 @@ type channelID struct {
 type mcapMerger struct {
 	schemaIDs  map[schemaID]uint16
 	channelIDs map[channelID]uint16
+	schemaIdByHash map[string]uint16
 
 	nextChannelID uint16
 	nextSchemaID  uint16
@@ -53,6 +56,7 @@ func newMCAPMerger(opts mergeOpts) *mcapMerger {
 	return &mcapMerger{
 		schemaIDs:            make(map[schemaID]uint16),
 		channelIDs:           make(map[channelID]uint16),
+		schemaIdByHash:       make(map[string]uint16),
 		nextChannelID:        1,
 		nextSchemaID:         1,
 		opts:                 opts,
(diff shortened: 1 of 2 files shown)


go/cli/mcap/cmd/merge.go
```diff
@@ -157,6 +169,13 @@ func (m *mcapMerger) mergeInputs(w io.Writer, inputs []namedReader) error {
 	profiles := make([]string, len(inputs))
 	pq := utils.NewPriorityQueue(nil)
 
+	// Reset struct members
+	m.schemaIdByHash = make(map[string]uint16)
+	m.schemaIDs = make(map[schemaID]uint16)
+	m.channelIDs = make(map[channelID]uint16)
+	m.nextChannelID = 1
+	m.nextSchemaID = 1
```
achim-k, line 177: Here I have to reset the `schemaIdByHash` as we might have a new writer where the schemas haven't been written to yet. 
In general, I don't fully understand the purpose of the mcapMerger struct. Wouldn't a single function be sufficient? I don't think users are supposed to call `mcapMerger.addSchema` themselves? But maybe I don't fully understand Go's concept of private members yet.

wkalt: I think this could definitely be written as a single function. There are no users of mcapMerger since it's private. I don't remember the details about mcapMerger but I'm pretty sure it only exists to assist with whatever remapping of channel/schema IDs are required, and you could write that either way.

wkalt: are you doing this resetting because you anticipate mergeInputs will be called multiple times?

wkalt: ^^ this seems reasonable to me (though I don't think it actually will be called that way), as would writing a single function -- maybe that's why you were asking.

Patch looks good to me generally, thanks!

wkalt (approved):

go/cli/mcap/cmd/merge.go
achim-k: > are you doing this resetting because you anticipate mergeInputs will be called multiple times?

It's [called multiple times](https://github.com/foxglove/mcap/blob/e422d7f9917be88644c401f89368c117bf105eda/go/cli/mcap/cmd/merge_test.go#L145-L153) in the tests. The resetting was required to make it succeed.

PR #984, opened 2023-09-30
fix negative timestamp in cpp protobuf example

wirthual:
### Public-Facing Changes

None

### Description

Example code generates negative nanosecond values in the timestamp.

Before:
Screenshot from 2023-09-30 07-56-12

After:
Screenshot from 2023-09-30 07-54-56


james-rms (approved):

jtbandes:
Thanks!

PR #985, opened 2023-10-02
go/cli: Coalesce identical channels by default

achim-k:
### Public-Facing Changes
go/cli: Coalesce identical channels by default

### Description
Changes the default behavior of `mcap merge` to automatically coalesce identical channels. Channels are considered identical when their topic, schema, message encoding and metadata equals. One can force coalescing of channels with different metadata with `--coalesce-channels force`. Channel coalescing can also be completely disabled with `--coalesce-channels none`.

### DIFF
--- a/go/cli/mcap/cmd/merge.go
+++ b/go/cli/mcap/cmd/merge.go
@@ -66,7 +66,7 @@ type HashSum = [md5.Size]byte
 type mcapMerger struct {
 	schemaIDs       map[schemaID]uint16
 	channelIDs      map[channelID]uint16
-	schemaIDByHash  map[string]uint16
+	schemaIDByHash  map[HashSum]uint16
 	channelIDByHash map[HashSum]uint16
 	metadataHashes  map[string]bool
 	metadataNames   map[string]bool
@@ -85,7 +85,7 @@ func newMCAPMerger(opts mergeOpts) *mcapMerger {
 	return &mcapMerger{
 		schemaIDs:       make(map[schemaID]uint16),
 		channelIDs:      make(map[channelID]uint16),
-		schemaIDByHash:  make(map[string]uint16),
+		schemaIDByHash:  make(map[HashSum]uint16),
 		channelIDByHash: make(map[HashSum]uint16),
 		metadataHashes:  make(map[string]bool),
 		metadataNames:   make(map[string]bool),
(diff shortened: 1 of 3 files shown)


jtbandes (commented):
👍  The idea seems fine to me but I don't think I'm the best person to review the code changes.

defunctzombie:
@wkalt @james-rms thoughts?

achim-k:
Friendly ping @wkalt / @james-rms / @narasaka

go/cli/mcap/cmd/merge.go
```diff
@@ -361,4 +409,11 @@ func init() {
 		true,
 		"chunk the output file",
 	)
+	mergeCmd.PersistentFlags().StringVarP(
+		&coalesceChannels,
+		"coalesce-channels",
+		"",
+		"auto",
+		"channel coalescing behavior (supported: auto, force, none)",
```
james-rms, line 417: Can we explain the behavior of each further in the help command output?

go/cli/mcap/cmd/merge.go
```diff
@@ -82,6 +93,30 @@ func (m *mcapMerger) outputSchemaID(inputID int, inputSchemaID uint16) (uint16,
...
+			hasher.Write([]byte(value))
+		}
+	case ForceCoalescing: // Channel metadata is not included in hash
+		break
+	default:
+		die("Invalid value for --coalesce-channels: %s\n", coalesceChannels)
+	}
+
+	hash := hasher.Sum(nil)
+	return hex.EncodeToString(hash)
```
james-rms, line 117: seems odd to me that we store the hash as a string rather than a fixed size value - won't block the PR, just seems odd.

james-rms: I feel like channelIdByHash should key off the result of `hasher.Sum()`.

go/cli/mcap/cmd/merge.go
achim-k: Done. Shows up like this now:
image

go/cli/mcap/cmd/merge.go
achim-k: Done. Should I do the same for the `schemaIDByHash` map ?

james-rms: Yep, that would be good.

james-rms (approved):

PR #988, opened 2023-10-04
Use Python 3.8 for CI builds

achim-k:
### Public-Facing Changes

Use Python 3.8 for CI builds 

### Description
Ensures that Python3.8 is used for CI. This was not the case so far as `pipenv` was "resetting" the python version:

```
make lint
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.7.17/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.7.17/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.7.17/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.7.17/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.7.17/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.7.17/x64/lib
SETUPTOOLS_ENABLE_FEATURES="legacy-editable" pipenv install --dev --deploy
Creating a virtualenv for this project...
Pipfile: /home/runner/work/mcap/mcap/python/Pipfile
Using /usr/bin/python3 (3.10.12) to create virtualenv...
created virtual environment CPython3.10.12.final.0-64 in 636ms
```

Resolves [FG-5220](https://linear.app/foxglove/issue/FG-5220/community-pr-mcap-compatibility-with-older-python-versions)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,7 +248,7 @@ jobs:
           lfs: true
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.7
+          python-version: 3.8
       - run: pip install pipenv==2023.2.4
       - run: make lint
       - run: make test
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -55,7 +55,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.7
+          python-version: 3.8
 
       - run: pip install pipenv
 
--- a/python/Pipfile
+++ b/python/Pipfile
@@ -35,4 +35,4 @@ typing_extensions = "*"
 isort = "*"
 
 [requires]
-python_version = "3"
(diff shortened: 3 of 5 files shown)


jtbandes (approved):
Looks good, thanks for doing this.

Why 3.8 instead of 3.7?

Also why is there a new lint error? 🤔

achim-k:
> Why 3.8 instead of 3.7?

I initially tried with 3.7 but that resulted in various errors due to a [deprecated endpoint](https://stackoverflow.com/a/73932581). (the library mentioned in that answer also requires Python 3.8 now)

> 
> Also why is there a new lint error? 🤔

Good question, I'll see if I can find out why these errors (there are more after fixing that one) suddenly appear

PR #989, opened 2023-10-06
Rust: make mcap-rs wasm compatible

JiangengDong:
### Public-Facing Changes

* Dependency is changed from [lz4](https://github.com/10xGenomics/lz4-rs) to [lz4_flex](https://github.com/PSeitz/lz4_flex)
* A new feature flag `lz4` is introduced and enabled by default

### Description

I am trying to use `mcap-rs` in an application that targets multiple platforms, including native and wasm. I modified the source code to make it wasm compatible, and think this may be also useful for others. 

The changes include: 

* Replace [lz4](https://github.com/10xGenomics/lz4-rs) with [lz4_flex](https://github.com/PSeitz/lz4_flex), because the former one depends on `lz4-sys`, which is not easy to port to wasm. 
* Add a feature flag `lz4` so the library user can choose to opt-out of the dependency.
* ~~Turn off the default features and~~ turn on the `wasm` feature for [zstd](https://github.com/gyscos/zstd-rs) when the target platform is wasm. 
* Add Clippy checking in the CI for all combinations of features.
* Minor fixes for Clippy linter warnings.

I am not quite sure if it would be a good idea to change the `lz4` library. It is also possible to still use [lz4](https://github.com/10xGenomics/lz4-rs) library and opt-out of it when the platform is wasm. 

I haven't raised an issue yet, because I am not sure if this belongs to a bug fix or feature request. (Feature request always leads me to the "discussion" page. Is that the correct place?) Please let me know where I should create it.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -25,9 +25,7 @@ thiserror = "1.0"
 lz4_flex = { version = "0.11.1", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-zstd = { version = "0.11", default-features = false, features = [
-    "wasm",
-], optional = true }
+zstd = { version = "0.11", features = ["wasm"], optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 zstd = { version = "0.11", features = ["zstdmt"], optional = true }
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -2,8 +2,8 @@
 name = "mcap"
 description = "A library for reading and writing MCAP files"
 homepage = "https://mcap.dev"
-keywords = ["foxglove", "mcap"]
-categories = ["science::robotics", "compression"]
+keywords = [ "foxglove", "mcap" ]
+categories = [ "science::robotics", "compression" ]
 repository = "https://github.com/foxglove/mcap"
 documentation = "https://docs.rs/mcap"
 readme = "README.md"
@@ -39,7 +39,7 @@ lz4 = ["dep:lz4_flex"]
 anyhow = "1"
 atty = "0.2"
 camino = "1.0"
-clap = { version = "3.2", features = ["derive"] }
+clap = { version = "3.2", features = ["derive"]}
 itertools = "0.10"
 memmap = "0.7"
 rayon = "1.5"
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -2,8 +2,8 @@
 name = "mcap"
 description = "A library for reading and writing MCAP files"
 homepage = "https://mcap.dev"
-keywords = [ "foxglove", "mcap" ]
-categories = [ "science::robotics", "compression" ]
+keywords = ["foxglove", "mcap"]
+categories = ["science::robotics", "compression"]
 repository = "https://github.com/foxglove/mcap"
 documentation = "https://docs.rs/mcap"
 readme = "README.md"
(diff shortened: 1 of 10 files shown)


defunctzombie:
Thanks for opening this effort and the writeup; this is definitely a feature that I think makes the rust library more desirable and we should add this capability.

Sharing some of my thoughts:

> * A new feature flag `lz4` is introduced and enabled by default

Is there a feature flag for zstd? If not - why does only lz4 get a feature flag?

> I am not quite sure if it would be a good idea to change the `lz4` library.

If the new library supports writing and reading lz4 in a way that the existing library also supports then I don't see a problem. I don't know if anything specific went into picking `lz4` over the other options when the rust code was first added.

> * Turn off the default features and turn on the `wasm` feature for [zstd](https://github.com/gyscos/zstd-rs) when the target platform is wasm.

Do we need to turn off the default features? It isn't clear to me if turning off the default is a requirement to turning wasm on.

cc @mrkline

JiangengDong:
> Is there a feature flag for `zstd`?

Yes, and it is already on main. https://github.com/foxglove/mcap/blob/c687014acc0ea5aec4201d7141183e030d2602ce/rust/Cargo.toml#L30

> If the new library supports writing and reading lz4 in a way that the existing library also supports then I don't see a problem.

I generated several mcap files and tried without a problem. Let me know if we should add a test for this. `lz4` is still the most popular lz4 compression library in Rust, and I think that's why it was chosen at the beginning.

> Do we need to turn off the default features? (zstd)

Good point. I copied the code from another post and did not verify by myself. Just tried again and confirmed we don't need to turn off the default features. Pushed a commit to fix this.

rust/Cargo.toml
```diff
@@ -2,8 +2,8 @@
 name = "mcap"
 description = "A library for reading and writing MCAP files"
 homepage = "https://mcap.dev"
-keywords = [ "foxglove", "mcap" ]
-categories = [ "science::robotics", "compression" ]
+keywords = ["foxglove", "mcap"]
+categories = ["science::robotics", "compression"]
```
mrkline, line 6: Was this some auto-formatter?

rust/Cargo.toml
```diff
@@ -19,21 +19,27 @@ byteorder = "1.4"
...
-lz4 = "1.0"
 num_cpus = "1.13"
 paste = "1.0"
 thiserror = "1.0"
+lz4_flex = { version = "0.11.1", optional = true }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+zstd = { version = "0.11", features = ["wasm"], optional = true }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
```
mrkline, line 30: Nice - appreciate you only pulling in wasm support when it's the target!

mrkline (approved):
Nice!

As a nit I'd say that it would be nicer to have cleanup of unrelated code (e.g., reformatting, `self: &Self -> &self`, etc.) as a seperate PR, but wider support for compression is great in my book!

FWIW, I don't think there was any deep reason for chosing `lz4` other than that it's a popular crate. :sweat_smile:

rust/Cargo.toml
JiangengDong: Yes. I can revert it.

JiangengDong:
Pushed two commits to revert all the irrelevant changes made by the linter and formatter. Let me know if this looks good to you.

defunctzombie (approved):

PR #990, opened 2023-10-08
go/mcap: Fix scenario where schemaID equals 0

StarCsu:
### Public-Facing Changes

mcap specification  explains schema id that it must be not zero,and go cli has zero-check,The picture is as follows:

img_v2_8730037e-0fbf-4014-8df4-14b4113821eg

### Description

schema's documnet of website is [https://mcap.dev/spec](https://mcap.dev/spec);

image

so i add this judgment that schema can't not zero,but I'm not sure if this merging operation is reasonable；


StarCsu:
i hvae signed cla,but i don't know why it still has cla errors;

PR #992, opened 2023-10-11
go/mcap: Fix scenario where schemaID equals 0

StarCsu:
### Public-Facing Changes

mcap specification  explains schema id that it must be not zero,and go cli has zero-check,The picture is as follows:

img_v2_8730037e-0fbf-4014-8df4-14b4113821eg

### Description

schema's documnet of website is [https://mcap.dev/spec](https://mcap.dev/spec);

image

so i add this judgment that schema can't not zero,but I'm not sure if this merging operation is reasonable；


wkalt:
@StarCsu thanks, I think we should take this patch. However, are all the changes in go.work.sum necessary?

StarCsu:
> @StarCsu thanks, I think we should take this patch. However, are all the changes in go.work.sum necessary?

yes,go.work.sum is used for project version verification, which needs to be updated, and the go test pipeline has been verified and passed;

wkalt:
@StarCsu What did you run to update the file? I understand it won't cause tests to fail or anything, and if this is something we just should have run previously I'm fine to commit it. But changes to that file can also come in from local modifications in the user's environment, so I just want to make sure those changes are actually relevant to this project.

StarCsu:
> @StarCsu What did you run to update the file? I understand it won't cause tests to fail or anything, and if this is something we just should have run previously I'm fine to commit it. But changes to that file can also come in from local modifications in the user's environment, so I just want to make sure those changes are actually relevant to this project.

i understand your means,go.sum's changes is not my local user's environment,if everyone use this cmd that is "go mod tidy",it will become current content;

Considering that I am not completely dependent on this go.work.sum, I rolled back this file

defunctzombie (approved):
@james-rms any other issues with this PR. It LGTM. If you are 👍 please merge.

PR #994, opened 2023-10-23
Python: Fix ROS2 message definition parsing on Windows

olavsr:
### Public-Facing Changes

None

### Description

Fixes issue on Windows, where empty lines in the ROS2 message definitions would be replaced by `\r\n`.
The regex-patterns don't handle this, which led to an `InvalidFieldDefinition` exception later on in `parse_message_string()`.
See relevant documentation on `os.linesep` here: https://docs.python.org/3/library/os.html#os.linesep

### DIFF
--- a/python/mcap-ros2-support/mcap_ros2/_dynamic.py
+++ b/python/mcap-ros2-support/mcap_ros2/_dynamic.py
@@ -1,6 +1,5 @@
 """ROS2 message definition parsing and message deserialization."""
 
-import os
 import re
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
Looks straightforward to me. Thanks for the fix!

jtbandes:
This is now published in version 0.5.3

olavsr:
Great! Thanks for releasing it so quickly.

PR #998, opened 2023-10-24
Fix `mcap cat` SIGSEGV when no schema present

b-camacho:
### Public-Facing Changes
`mcap cat` used to crash on some valid, but uncommon files. This fixes the crash.

### Description
Mcap spec [allows for a missing schema](https://mcap.dev/spec#channel-op0x04).
Currently, `mcap cat` crashes with a nil pointer dereference on such a file (unless you pass `--json`).
After this change, `mcap cat` will print `[no schema]` instead of a schema for schemaless channels.

Also added another test to `cat_test.go`, and added tests for `mcap info` in `info_test.go`

### DIFF
--- a/go/cli/mcap/cmd/info_test.go
+++ b/go/cli/mcap/cmd/info_test.go
@@ -1,0 +1,94 @@
+package cmd
+
+import (
+	"bytes"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/foxglove/mcap/go/mcap"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInfo(t *testing.T) {
+	cases := []struct {
+		assertion string
+		inputfile string
+		expected  string
+	}{
+		{
+			"OneMessage",
+			"../../../../tests/conformance/data/OneMessage/OneMessage-ch-chx-mx-pad-rch-rsh-st-sum.mcap",
+			`library:
+profile:
+messages:  1
+duration:  0s
+start:     0.000000002
+end:       0.000000002
+compression:
+	: [1/1 chunks] [115.00 B/115.00 B (0.00%)]
+channels:
+	(1) example  1 msgs (+Inf Hz)   : Example [c]
+attachments: 0
+metadata: 0`,
+		},
+		{
+			"OneSchemalessMessage",
+			"../../../../tests/conformance/data/OneSchemalessMessage/OneSchemalessMessage-ch-chx-mx-pad-rch-st.mcap",
+			`library:
+profile:
+messages:  1
+duration:  0s
+start:     0.000000002
+end:       0.000000002
+compression:
+	: [1/1 chunks] [70.00 B/70.00 B (0.00%)]
+channels:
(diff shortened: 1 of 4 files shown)


jtbandes:
Hmm, how was this missed in https://github.com/foxglove/mcap/pull/857? Are there more tests we need to add?

b-camacho:
Likely, I tried the CLI on the schemaless file that's checked in for conformance testing and found another crash
```
./mcap info ./tests/conformance/data/OneSchemalessMessage/OneSchemalessMessage.mcap
goroutine 1 [running]:
github.com/foxglove/mcap/go/cli/mcap/cmd.printInfo({0xf806e0, 0xc00011e028}, 0xc000420460)
	/home/bmc/sources/mcap/go/cli/mcap/cmd/info.go:134 +0x1f2d
```
I'll throw some more tests in

b-camacho:
Added more tests like you suggested @jtbandes

james-rms (approved):

go/cli/mcap/cmd/info_test.go
```diff
@@ -0,0 +1,94 @@
...
+metadata: 0`,
+		},
+		{
+			"OneSchemalessMessage_NoChannels",
+			"../../../../tests/conformance/data/OneSchemalessMessage/OneSchemalessMessage.mcap",
+			`library:
+profile:
+channels:
+attachments: unknown
+metadata: unknown`,
```
wkalt, line 58: I'm a bit nervous this will become annoying to update as the info command gains new functionality. This text display isn't really stable - for that we should probably introduce some kind of specification for the output.

Would it be sufficient to test that printInfo simply doesn't crash the program? If you're interested in asserting on the name of the spoofed schema, you should be able to do that by making an assertion against the info structure's Schemas field.

wkalt: I wonder whether if we wrote that test to loop over all of the conformance files, we'd find more bugs or if this is the only one. That seems like it would be a stronger test to me, but I don't want to obligate you to address whatever else it might show.

b-camacho: I think both of these are good points!
I made the test run on every `.mcap` in the conformance folder and only found one new crash, so I fixed it

wkalt (approved):
@b-camacho thanks for the patch and for the new test coverage! looks good to me now & I will merge it.

I will say that although we support these "schemaless" channels in the spec, it's one area I don't feel too good about. Schemaless channels can either be described "implicitly" with a zero-valued schema ID, or explicitly with a Schema record that has encoding="" and data="". IMO supporting the "implicit" version has more costs than benefits. It means all consumers need to be prepared for a nil-valued schema to appear, or else they get bugs like this, and it complicates the modeling of mcap files in a relational DB, by introducing nulls and left joins. If we instead force the writer to handle writing an empty schema, all downstream code is simplified.

Anyway, sorry for the digression... that's mostly for your curiosity & in case you didn't know of the other way. Since it's in the spec we need to support it, so thanks again for the patch.

PR #1005, opened 2023-11-01
Update CLI install instructions on mcap.dev

james-rms:


jtbandes (commented):
The code block no longer matches the paragraph above. Also, I would like to avoid having to update this each time we make a new release

PR #1008, opened 2023-11-02
C++: Make LZ4 and ZSTD dependencies optional + clang-format

asherikov:
### Public-Facing Changes

New defines: MCAP_COMPRESSION_NO_ZSTD, MCAP_COMPRESSION_NO_LZ4

### Description

Make LZ4 and ZSTD dependencies optional

### DIFF
--- a/cpp/mcap/include/mcap/reader.inl
+++ b/cpp/mcap/include/mcap/reader.inl
@@ -1,9 +1,13 @@
 #include "internal.hpp"
 #include <algorithm>
 #include <cassert>
-#include <lz4frame.h>
-#include <zstd.h>
-#include <zstd_errors.h>
+#ifndef MCAP_COMPRESSION_NO_LZ4
+#  include <lz4frame.h>
+#endif
+#ifndef MCAP_COMPRESSION_NO_ZSTD
+#  include <zstd.h>
+#  include <zstd_errors.h>
+#endif
 
 namespace mcap {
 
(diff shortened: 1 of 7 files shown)


cpp/mcap/include/mcap/reader.inl
```diff
@@ -1251,10 +1259,24 @@ TypedChunkReader::TypedChunkReader()
...
+      break;
+#ifndef MCAP_COMPRESSION_NO_LZ4
+    case Compression::Lz4:
+      decompressor = static_cast<ICompressedReader*>(&lz4Reader_);
+      break;
+#endif
+#ifndef MCAP_COMPRESSION_NO_ZSTD
+    case Compression::Zstd:
+      decompressor = static_cast<ICompressedReader*>(&zstdReader_);
+      break;
```
james-rms, line 1277: An MCAP chunk containing an unrecognized compression string will cause a reader to defereference a null pointer 
here. Before this patch we would have assumed any unrecognized encoding was `zstd`, which would have caused some confusing error message, but likely not a segfault.

Would you be able to add an early return here in case of an unrecognized compression method?

asherikov: You mean `decompressor` pointer? It is initialized to uncompressed reader in the default switch case above.

asherikov: I've modified it slightly to make behavior the same as before when ZSTD is enabled.

james-rms: I don't think it makes sense to default to zstd here, we should set `status_` and return.

When both lz4 and zstd are built into the reader, this switch case is exhaustive. This is why the existing code works: we detected unrecognised compression strings in `ParseCompression` before ever getting here.

We're introducing a new error condition here for when a compression string is recognised but not supported. I think it makes sense to add a new error in `errors.hpp` for this. We could then go through all the supported compression values in this switch statement, and in the `default:` case set `status_ = ` and return early.

asherikov: Ok, now ParseCompression should fail if compression is not compiled in.

james-rms:
Thanks for the PR! I'm continuing with your changes here: https://github.com/foxglove/mcap/pull/1013

PR #1009, opened 2023-11-02
Add subcommand for sorting a file

wkalt:
Adds a "sort" subcommand to the mcap CLI tool. This will rewrite the messages into a new file, physically sorted on time.

### DIFF
--- a/go/cli/mcap/cmd/sort.go
+++ b/go/cli/mcap/cmd/sort.go
@@ -14,6 +14,19 @@ var (
 	sortOutputFile string
 )
 
+type errUnindexedFile struct {
+	err error
+}
+
+func (e errUnindexedFile) Error() string {
+	return e.err.Error()
+}
+
+func (e errUnindexedFile) Is(tgt error) bool {
+	_, ok := tgt.(errUnindexedFile)
+	return ok
+}
+
 func sortFile(w io.Writer, r io.ReadSeeker) error {
 	reader, err := mcap.NewReader(r)
 	if err != nil {
@@ -29,7 +42,7 @@ func sortFile(w io.Writer, r io.ReadSeeker) error {
 	}
 	info, err := reader.Info()
 	if err != nil {
-		return fmt.Errorf("failed to read info: %w", err)
+		return errUnindexedFile{err}
 	}
 
 	err = writer.WriteHeader(info.Header)
@@ -120,6 +133,9 @@ var sortCmd = &cobra.Command{
 		}
 		err = sortFile(output, f)
 		if err != nil {
+			if errors.Is(err, errUnindexedFile{}) {
+				die("Error reading file index: %s. You may need to run `mcap recover` if the file is corrupt.", err)
+			}
 			die("failed to sort file: %s", err)
 		}
 	},
--- a/go/cli/mcap/cmd/sort.go
+++ b/go/cli/mcap/cmd/sort.go
@@ -45,16 +45,14 @@ func fileHasNoMessages(r io.ReadSeeker) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	for {
-		_, _, _, err := it.Next(nil)
-		if err != nil {
-			if errors.Is(err, io.EOF) {
-				return true, nil
-			}
-			return false, err
+	_, _, _, err = it.Next(nil)
+	if err != nil {
+		if errors.Is(err, io.EOF) {
+			return true, nil
 		}
-		return false, nil
+		return false, err
 	}
+	return false, nil
 }
 
 func sortFile(w io.Writer, r io.ReadSeeker) error {
--- a/go/cli/mcap/cmd/sort.go
+++ b/go/cli/mcap/cmd/sort.go
@@ -175,7 +175,7 @@ var sortCmd = &cobra.Command{
 		if err != nil {
 			if errors.Is(err, errUnindexedFile{}) {
 				die("Error reading file index: %s. "+
-					"You may need to run `mcap recover` if the file is corrupt or unchunked.", err)
+					"You may need to run `mcap recover` if the file is corrupt or not chunk indexed.", err)
 			}
 			die("failed to sort file: %s", err)
 		}
--- a/go/cli/mcap/cmd/sort.go
+++ b/go/cli/mcap/cmd/sort.go
@@ -11,7 +11,11 @@ import (
 )
 
 var (
-	sortOutputFile string
+	sortOutputFile  string
+	sortChunkSize   int64
+	sortCompression string
+	sortIncludeCRC  bool
+	sortChunked     bool
 )
 
 type errUnindexedFile struct {
@@ -33,9 +37,10 @@ func sortFile(w io.Writer, r io.ReadSeeker) error {
 		return fmt.Errorf("failed to create reader: %w", err)
 	}
 	writer, err := mcap.NewWriter(w, &mcap.WriterOptions{
-		Chunked:     true,
-		Compression: mcap.CompressionZSTD,
-		ChunkSize:   4 * 1024 * 1024,
+		Chunked:     sortChunked,
+		Compression: mcap.CompressionFormat(sortCompression),
+		ChunkSize:   sortChunkSize,
+		IncludeCRC:  sortIncludeCRC,
 	})
 	if err != nil {
 		return fmt.Errorf("failed to create writer: %w", err)
@@ -43,6 +48,10 @@ func sortFile(w io.Writer, r io.ReadSeeker) error {
 	info, err := reader.Info()
 	if err != nil {
 		return errUnindexedFile{err}
+	}
+
+	if len(info.ChunkIndexes) == 0 {
+		return errUnindexedFile{errors.New("no chunk index records found")}
 	}
 
 	err = writer.WriteHeader(info.Header)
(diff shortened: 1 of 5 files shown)


go/mcap/reader.go
```diff
@@ -184,6 +184,20 @@ func (r *Reader) Info() (*Info, error) {
 	return info, nil
 }
 
+// GetAttachment returns an attachment reader located at the specific offset.
```
james-rms, line 187: changing `go/mcap` means we should update the library version number in `version.go`

wkalt: thanks, done

go/mcap/version.go
```diff
@@ -1,4 +1,4 @@
 package mcap
 
 // Version of the MCAP library.
-var Version = "v1.0.4"
+var Version = "v1.0.5"
```
james-rms, line 4:
```suggestion
var Version = "v1.1.0"
```

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,141 @@
...
+		Chunked:     true,
+		Compression: mcap.CompressionZSTD,
+		ChunkSize:   4 * 1024 * 1024,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create writer: %w", err)
+	}
+	info, err := reader.Info()
+	if err != nil {
+		return fmt.Errorf("failed to read info: %w", err)
```
james-rms, line 32: can we have a more friendly message if the input file is not indexed? the user can `mcap recover` first, it'd be good to tell them that.

go/mcap/version.go
wkalt: pushed directly

go/cli/mcap/cmd/sort.go
wkalt: updated

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,157 @@
...
+			return fmt.Errorf("failed to write message: %w", err)
+		}
+	}
+
+	return writer.Close()
+}
+
+var sortCmd = &cobra.Command{
+	Use:   "sort [file] -o output.mcap",
+	Short: "Read an MCAP file and write the messages out physically sorted on time",
```
jtbandes, line 119: specify that this means log time?

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,157 @@
...
+		defer f.Close()
+
+		output, err := os.Create(sortOutputFile)
+		if err != nil {
+			die("failed to open output: %s", err)
+		}
+		err = sortFile(output, f)
+		if err != nil {
+			if errors.Is(err, errUnindexedFile{}) {
+				die("Error reading file index: %s. You may need to run `mcap recover` if the file is corrupt.", err)
```
wkalt, line 137: linter might nail me for this

james-rms (approved):

go/cli/mcap/cmd/sort.go
jtbandes, line 137: I would expect to be able to use a `sort` command on an unindexed file. If the file is unindexed, but not corrupt, we should probably have a better message in that case.

wkalt: nope

james-rms:
make sure you tag both releases (`releases/mcap-cli/vx.y.z` and `go/mcap/v1.1.0`) after merging.

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,157 @@
...
+
+func sortFile(w io.Writer, r io.ReadSeeker) error {
+	reader, err := mcap.NewReader(r)
+	if err != nil {
+		return fmt.Errorf("failed to create reader: %w", err)
+	}
+	writer, err := mcap.NewWriter(w, &mcap.WriterOptions{
+		Chunked:     true,
+		Compression: mcap.CompressionZSTD,
+		ChunkSize:   4 * 1024 * 1024,
```
jtbandes, line 38: should we make these settings configurable? (I think they are for some of the other subcommands, but can't recall)

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return errUnindexedFile{err}
+	}
+
+	err = writer.WriteHeader(info.Header)
+	if err != nil {
+		return fmt.Errorf("failed to write header: %w", err)
+	}
+
+	// handle the attachments and metadata metadata first; physical location in
```
jtbandes, line 53:
```suggestion
	// handle the attachments and metadata first; physical location in
```

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,157 @@
...
+	// the file is irrelevant but order is preserved.
+	for _, index := range info.AttachmentIndexes {
+		ar, err := reader.GetAttachment(index.Offset)
```
jtbandes, line 56: I know that short variable names are common in this codebase but I cannot imagine what `ar` stands for.

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+	for _, index := range info.MetadataIndexes {
+		metadata, err := reader.GetMetadata(index.Offset)
+		if err != nil {
+			return fmt.Errorf("failed to read attachment: %w", err)
+		}
+		err = writer.WriteMetadata(metadata)
+		if err != nil {
+			return fmt.Errorf("failed to read attachment: %w", err)
```
jtbandes, line 79: error message is wrong

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,157 @@
...
+		err = writer.WriteAttachment(&mcap.Attachment{
+			Name:       index.Name,
+			MediaType:  index.MediaType,
+			CreateTime: index.CreateTime,
+			LogTime:    index.LogTime,
+			DataSize:   index.DataSize,
+			Data:       ar.Data(),
+		})
+		if err != nil {
+			return fmt.Errorf("failed to read attachment: %w", err)
```
jtbandes, line 69: error message is wrong

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,157 @@
...
+	schemas := make(map[uint16]*mcap.Schema)
+	channels := make(map[uint16]*mcap.Schema)
+	for {
+		schema, channel, message, err := it.Next(nil)
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+		}
+		if _, ok := schemas[schema.ID]; !ok {
```
jtbandes, line 96: Handle nil schema?

go/cli/mcap/cmd/sort.go
wkalt: it stands for AttachmentReader, which is a type that unfortunately needs to be public and different from Attachment. I should perhaps rename this function to GetAttachmentReader - I guess we might as well rip the bandaid off.

jtbandes: why not name the variable `attachmentReader` or `attReader`?

go/cli/mcap/cmd/sort.go
wkalt: I really wish we just had sane writer defaults, then we wouldn't need to even think about this.. :(

go/cli/mcap/cmd/sort.go
wkalt: sorry, missed this - was your proposal "attachmentReader := reader.GetAttachment()" or "attachmentReader := reader.GetAttachmentReader()"?

I don't really like the incongruity of the first, and the second to me doesn't make things clearer than "ar := reader.GetAttachmentReader()". Either way you need to figure out what an attachment reader is, if you don't know already.

wkalt: pushed up the change anyway

go/cli/mcap/cmd/sort.go
wkalt: now we will handle that like this:
```
[~/work/data-platform] (main) $ mcap sort gps.unchunked.mcap -o output.mcap
Error reading file index: no chunk index records. You may need to run `mcap recover` if the file is corrupt or unchunked.
```

wkalt:
@jtbandes thanks for the review, comments all addressed

go/cli/mcap/cmd/sort.go
jtbandes: yeah, but in the original code you also have to figure out what an `ar` is 😂

go/cli/mcap/cmd/sort.go
```diff
@@ -0,0 +1,196 @@
...
+
+		output, err := os.Create(sortOutputFile)
+		if err != nil {
+			die("failed to open output: %s", err)
+		}
+		err = sortFile(output, f)
+		if err != nil {
+			if errors.Is(err, errUnindexedFile{}) {
+				die("Error reading file index: %s. "+
+					"You may need to run `mcap recover` if the file is corrupt or unchunked.", err)
```
jtbandes, line 178: nit: should this be "unindexed" rather than unchunked? That would match the `errUnindexedFile` we're testing for

go/cli/mcap/cmd/sort_test.go
```diff
@@ -0,0 +1,61 @@
...
+	assert.Nil(t, writer.WriteHeader(&mcap.Header{}))
+	assert.Nil(t, writer.WriteSchema(&mcap.Schema{
+		ID:       1,
+		Name:     "foo",
+		Encoding: "ros1",
+		Data:     []byte{},
+	}))
+	assert.Nil(t, writer.WriteChannel(&mcap.Channel{
+		ID:              0,
+		SchemaID:        1,
```
jtbandes, line 26: maybe include a channel with schemaid 0 just to exercise that condition?

jtbandes (approved):

go/cli/mcap/cmd/sort.go
wkalt: unindexed doesn't seem right to me either, because we can have other kinds of indexes just not chunks. I can say chunk-indexed.

PR #1012, opened 2023-11-13
Add Mcap Appender to the Typescript Core package

alexern14:
### Public-Facing Changes

- Added data end offset information to Typescipt Mcap Indexed Reader.
- Added an Mcap Appender to the Typescript Mcap core package.

### Description

- Added data end offset information to Typescipt Mcap Indexed Reader for Mcap appending purposes.
- Added an Mcap Appender to the Typescript Mcap core package in order to append metadata and attachments to existing Mcap files via the @mcap/core package.

### DIFF
--- a/typescript/core/package.json
+++ b/typescript/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mcap/core",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "description": "MCAP file support in TypeScript",
   "license": "MIT",
   "repository": {
--- a/typescript/core/src/McapAppender.ts
+++ b/typescript/core/src/McapAppender.ts
@@ -1,0 +1,406 @@
+import { crc32Init, crc32Update, crc32Final } from "@foxglove/crc";
+
+import { IWritable } from "./IWritable";
+import { McapRecordBuilder } from "./McapRecordBuilder";
+import { Opcode } from "./constants";
+import {
+  Schema,
+  Channel,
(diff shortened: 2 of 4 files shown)


jtbandes:
Thanks for the PR, this is interesting functionality. Can I ask why you chose to make this a separate class and not a feature of the existing writer class? (e.g. "open in append mode")

alexern14:
@jtbandes I originally started this by just changing the functionality of the writer class to be able to append, and while doing some cleanup I just created a separate class and was not quite sure how you guys would want this organized. I am open and willing to changing it to be additional functionality on the writer class though!

jtbandes:
I'm not sure how easily it would slot into the existing class but I do think, from the standpoint of a user, it would be more natural to be able to use the same writer class for either creating a new file or modifying an existing one. I think it could be worth exploring this (and maybe opening a separate PR if you think it's not clear which is better).

I also think the user would rather not have to do things like `setStatistics` / `addChunkIndex` explicitly, but instead the writer/appender should populate itself by reading the file during creation. This might necessitate an async create function, like we have for the indexed reader:  https://github.com/foxglove/mcap/blob/f87352f0aaae81628bdd7ffc43f7563f6bd98547/typescript/core/src/McapIndexedReader.ts#L78

alexern14:
@jtbandes Thanks for the feedback! I just put up a different PR with that proposed implementation here https://github.com/foxglove/mcap/pull/1016

PR #1013, opened 2023-11-13
cpp: add build-time flags to make compression dependencies optional

james-rms:
### Public-Facing Changes

Adds two new build-time flags to optionally disable `zstd` and `lz4` support. These are `MCAP_COMPRESSION_NO_ZSTD` and `MCAP_COMPRESSION_NO_LZ4`.

When reading MCAP files, if the reader encounters a known compression method but is compiled without support for it, it will yield `Status::UnsupportedCompression`.

### Description

Based on https://github.com/foxglove/mcap/pull/1008

A new test binary is added to CI, which exercises these build-time flags.

### DIFF
--- a/cpp/mcap/include/mcap/errors.hpp
+++ b/cpp/mcap/include/mcap/errors.hpp
@@ -23,6 +23,7 @@ enum class StatusCode {
   DecompressionFailed,
   DecompressionSizeMismatch,
   UnrecognizedCompression,
+  UnsupportedCompression,
   OpenFailed,
   MissingStatistics,
   InvalidMessageReadOptions,
@@ -85,6 +86,9 @@ struct [[nodiscard]] Status {
         break;
       case StatusCode::UnrecognizedCompression:
         message = "unrecognized compression";
+        break;
+      case StatusCode::UnsupportedCompression:
+        message = "unsupported compression";
         break;
       case StatusCode::OpenFailed:
         message = "open failed";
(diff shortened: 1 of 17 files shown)


cpp/mcap/include/mcap/errors.hpp
```diff
@@ -23,6 +23,7 @@ enum class StatusCode {
   DecompressionFailed,
   DecompressionSizeMismatch,
   UnrecognizedCompression,
+  UnsupportedCompression,
```
jtbandes, line 26: would there perhaps be a good reason to keep new status codes monotonically increasing (not modifying any existing status codes)? I'm not sure – just flagging as something to think about

cpp/test/unit_tests.cpp
```diff
@@ -584,6 +584,7 @@ TEST_CASE("Message index records", "[writer]") {
   REQUIRE(messageIndexChannelIds[1] == channel2.id);
 }
 
+#ifndef MCAP_COMPRESSION_NO_LZ4
```
jtbandes, line 587: is there not a corresponding test case for zstd?

jtbandes (approved):
👍

cpp/mcap/include/mcap/errors.hpp
james-rms: Good idea.

cpp/test/unit_tests.cpp
james-rms: There is not.

asherikov:
Well, I was lazy with my PR, but if you are not I have a few suggestions:
- the defines should enable compression rather than disabling it;
- the defines should be located in corresponsing headers, so that the end users would need to include the headers they need instead of playing with defines;
- compression related stuff should be moved to mixin classes.

PS: oh, ok i guess too late.

PR #1014, opened 2023-11-15
Update TypeScript, Prettier, and other dev deps

jtbandes:
### Public-Facing Changes

[BREAKING] The `@mcap/core` source code now uses [private properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties). Depending on what runtime and build system you use to consume this package, you may need to add plugins or change settings to handle the new syntax. Modern JS engines and bundlers should not require any changes.

### Description

- Upgrade TypeScript, Prettier, `@foxglove/eslint-plugin` etc. and fix new lint errors.
- Use `es2022` for typescript `target` & `module` settings


snosenzo (approved):
glanced over it and it seems fine to me

defunctzombie (approved):

PR #1015, opened 2023-11-15
Add h264 recording support to mcap.dev demo

jtbandes:
### Public-Facing Changes

The recording demo on mcap.dev can now record H.264-encoded CompressedVideo, if supported by the browser.

### Description

- Use VideoEncoder to encode h264
- Detection of browser support & workarounds for some Safari issues (related to https://github.com/WebKit/WebKit/pull/15562)

Depends on #1014 (for VideoEncoder type definitions)


website/src/components/McapRecordingDemo/videoCapture.ts
```diff
@@ -66,36 +85,106 @@ export function startVideoCapture(params: VideoCaptureParams): () => void {
...
+      width: video.videoWidth,
+      height: video.videoHeight,
+      latencyMode: "realtime",
+      avc: { format: "annexb" },
+    };
+    if (typeof VideoEncoder !== "function") {
+      break setupEncoder;
+    }
+    if ((await VideoEncoder.isConfigSupported(config)).supported !== true) {
+      break setupEncoder;
```
snosenzo, line 116: What is the error message when it breaks like this? Why not throw a more specific error message?

website/src/components/McapRecordingDemo/videoCapture.ts
```diff
@@ -66,36 +85,106 @@ export function startVideoCapture(params: VideoCaptureParams): () => void {
...
-          framePromise = undefined;
         },
         "image/jpeg",
         0.8,
       );
-    });
+    }
   }, frameDurationSec * 1000);
 
   const cleanup = () => {
```
snosenzo, line 190: Should you do anything to clean up the encoder, event listeners, or framepool here?

website/src/components/McapRecordingDemo/videoCapture.ts
```diff
@@ -66,36 +85,106 @@ export function startVideoCapture(params: VideoCaptureParams): () => void {
...
+      );
+      const now = performance.now();
+      const frame = new VideoFrame(video);
+      const encodeOptions: VideoEncoderEncodeOptions = {};
+      if (
+        lastKeyframeTime == undefined ||
+        now - lastKeyframeTime >= keyframeInterval
+      ) {
+        encodeOptions.keyFrame = true;
+        lastKeyframeTime = now;
```
snosenzo, line 171: Does it matter that this is using `performance.now` instead of the timestamp from the VideoFrame? Depending on how long it takes to encode frames, you could get variable rate keyframes no?

website/src/components/McapRecordingDemo/videoCapture.ts
jtbandes: It falls back to jpeg encoding

website/src/components/McapRecordingDemo/videoCapture.ts
jtbandes: Does that matter? I’m not sure, you tell me 😄

website/src/components/McapRecordingDemo/videoCapture.ts
snosenzo: Ah thought it still went through the catch

website/src/components/McapRecordingDemo/videoCapture.ts
snosenzo: I don't really know how VideoFrame works on the video stream. I guess regardless you could still dropping `VideoFrames` between encoding frames potentially? It probably doesn't matter, better to err on writing more keyframes than not i suppose

defunctzombie (approved):

website/src/components/McapRecordingDemo/videoCapture.ts
```diff
@@ -66,40 +85,189 @@ export function startVideoCapture(params: VideoCaptureParams): () => void {
...
+          config,
+          err,
+        );
+      },
+    });
+    encoder.configure(config);
+    const bitmap = await createImageBitmap(new ImageData(1, 1));
+    const frame = new VideoFrame(bitmap, { timestamp: 0 });
+    bitmap.close();
+    encoder.encode(frame, { keyFrame: true });
```
james-rms, line 117: i guess it's simplest to encode all frames as keyframes, but if someone's using this demo to compare CompressedImage vs. VideoFrame, won't they see them as exactly the same?

james-rms (commented):
Feature request - are we able to give the user a toggle to record images or video? it'd be good to give them an idea of the compression benefit they can expect.

PR #1016, opened 2023-11-15
Add append mode to Mcap Writer in the Typescript Core package

alexern14:
### Public-Facing Changes

- Added data end offset information to Typescipt Mcap Indexed Reader.
- Added append mode to the Mcap Writer in the Typescript Mcap core package.

### Description

- Added data end offset information to Typescipt Mcap Indexed Reader for Mcap appending purposes.
- Added append mode to the Mcap Writer in the Typescript Mcap core package in order to append metadata and attachments to existing Mcap files via the @mcap/core package.

### DIFF
--- a/typescript/core/src/McapIndexedReader.ts
+++ b/typescript/core/src/McapIndexedReader.ts
@@ -18,6 +18,8 @@ type McapIndexedReaderArgs = {
   summaryOffsetsByOpcode: ReadonlyMap<number, TypedMcapRecords["SummaryOffset"]>;
   header: TypedMcapRecords["Header"];
   footer: TypedMcapRecords["Footer"];
+  dataEndOffset?: bigint;
+  dataSectionCrc?: number;
 };
 
 export class McapIndexedReader {
@@ -30,6 +32,9 @@ export class McapIndexedReader {
   readonly summaryOffsetsByOpcode: ReadonlyMap<number, TypedMcapRecords["SummaryOffset"]>;
   readonly header: TypedMcapRecords["Header"];
   readonly footer: TypedMcapRecords["Footer"];
+  // Used for appending attachments/metadata to existing MCAP files
+  readonly dataEndOffset?: bigint;
+  readonly dataSectionCrc?: number;
 
   #readable: IReadable;
   #decompressHandlers?: DecompressHandlers;
@@ -51,6 +56,8 @@ export class McapIndexedReader {
     this.summaryOffsetsByOpcode = args.summaryOffsetsByOpcode;
     this.header = args.header;
     this.footer = args.footer;
+    this.dataEndOffset = args.dataEndOffset;
+    this.dataSectionCrc = args.dataSectionCrc;
 
     for (const chunk of args.chunkIndexes) {
       if (this.#messageStartTime == undefined || chunk.messageStartTime < this.#messageStartTime) {
@@ -203,6 +210,13 @@ export class McapIndexedReader {
       throw errorWithLibrary("File is not indexed");
     }
 
+    const dataEndOffset = BigInt(
+      footer.summaryStart -
+        /* data end length */ 4n -
+        /* record content length */ 8n -
+        /* Opcode.FOOTER */ 1n,
+    );
+
     // Copy the footer prefix before reading the summary because calling readable.read() may reuse the buffer.
     const footerPrefix = new Uint8Array(
       /* Opcode.FOOTER */ 1 +
@@ -248,6 +262,7 @@ export class McapIndexedReader {
     const metadataIndexes: TypedMcapRecords["MetadataIndex"][] = [];
     const summaryOffsetsByOpcode = new Map<number, TypedMcapRecords["SummaryOffset"]>();
     let statistics: TypedMcapRecords["Statistics"] | undefined;
+    let dataSectionCrc = 0;
 
     let offset = 0;
     for (
@@ -281,6 +296,9 @@ export class McapIndexedReader {
         case "SummaryOffset":
           summaryOffsetsByOpcode.set(result.record.groupOpcode, result.record);
           break;
+        case "DataEnd":
+          dataSectionCrc = result.record.dataSectionCrc;
+          break;
         case "Header":
         case "Footer":
         case "Message":
@@ -288,7 +306,6 @@ export class McapIndexedReader {
         case "MessageIndex":
         case "Attachment":
         case "Metadata":
-        case "DataEnd":
           throw errorWithLibrary(`${result.record.type} record not allowed in index section`);
         case "Unknown":
           break;
@@ -310,6 +327,8 @@ export class McapIndexedReader {
       summaryOffsetsByOpcode,
       header,
       footer,
+      dataEndOffset,
+      dataSectionCrc,
     });
   }
 
--- a/typescript/core/src/McapWriter.test.ts
+++ b/typescript/core/src/McapWriter.test.ts
@@ -344,4 +344,310 @@ describe("McapWriter", () => {
       },
     ]);
   });
+
+  it("supports append mode", async () => {
+    const tempBuffer = new TempBuffer();
+
+    const writer = new McapWriter({ writable: tempBuffer, chunkSize: 0 });
+
+    await writer.start({ library: "", profile: "" });
+    const schemaId = await writer.registerSchema({
+      name: "test1",
+      encoding: "json",
+      data: new Uint8Array(),
+    });
+    const channelId = await writer.registerChannel({
+      topic: "test",
+      schemaId,
+      messageEncoding: "json",
+      metadata: new Map(),
+    });
+    await writer.addMessage({
+      channelId,
+      data: new Uint8Array(),
+      sequence: 0,
+      logTime: 0n,
+      publishTime: 0n,
+    });
+    await writer.end();
+
+    const writerAppendMode = await McapWriter.InitializeInAppendMode(tempBuffer);
+
+    await writerAppendMode.addAttachment({
+      name: "attachment test",
+      logTime: 0n,
+      createTime: 0n,
+      mediaType: "application/json",
+      data: new TextEncoder().encode(`{"test": "testValue"}`),
+    });
+    await writerAppendMode.addMetadata({
+      name: "metadata test",
+      metadata: new Map<string, string>([["test", "testValue"]]),
+    });
+    await writerAppendMode.addMessage({
+      channelId,
+      data: new Uint8Array(),
+      sequence: 1,
+      logTime: 1n,
+      publishTime: 1n,
+    });
+    const newChannelId = await writerAppendMode.registerChannel({
+      topic: "test/append",
+      schemaId,
+      messageEncoding: "json",
+      metadata: new Map(),
+    });
+    await writerAppendMode.addMessage({
+      channelId: newChannelId,
+      data: new Uint8Array(),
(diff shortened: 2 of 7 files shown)


typescript/core/src/IWritable.ts
```diff
@@ -7,4 +7,7 @@ export interface IWritable {
 
   // The current position in bytes from the start of the output
   position(): bigint;
+
+  // Seek the cursor to the given position
+  seek(position: bigint): Promise<void>;
```
jtbandes, line 12: I think most of our existing writers could be updated to add seeking support, but I think it would be better to scope this requirement down so it's only needed when opening in append mode. As-is this would also be a breaking change which would require a major version bump.

typescript/core/src/McapIndexedReader.ts
```diff
@@ -18,6 +18,7 @@ type McapIndexedReaderArgs = {
   summaryOffsetsByOpcode: ReadonlyMap<number, TypedMcapRecords["SummaryOffset"]>;
   header: TypedMcapRecords["Header"];
   footer: TypedMcapRecords["Footer"];
+  dataEndOffset: bigint;
```
jtbandes, line 21: Adding a new required param is a breaking change – can it be made optional?

jtbandes (commented):
This approach looks nicer. I think we could put the seek method in a separate interface to make the change backwards compatible.

jtbandes (commented):
Thanks for the continued work on this. In order to merge we will need to get some test coverage on this code. If you have some time & energy to put in, that would be appreciated – I think you should be able to follow some of the existing test patterns in McapWriter.test.ts, to create tests that write a file and then append to it with a 2nd writer instance, and demonstrating that the appended file has the correct records/indexes in it.

typescript/core/src/McapWriter.test.ts
```diff
@@ -344,4 +344,213 @@ describe("McapWriter", () => {
...
+      metadata: new Map<string, string>([["test", "testValue"]]),
+    });
+    await writerAppendMode.end();
+
+    const reader = new McapStreamReader();
+    reader.append(tempBuffer.get());
+    const records: TypedMcapRecord[] = [];
+    for (let rec; (rec = reader.nextRecord()); ) {
+      records.push(rec);
+    }
```
jtbandes, line 399: we could assert `reader.done()` here

typescript/core/src/McapIndexedReader.ts
```diff
@@ -203,6 +209,13 @@ export class McapIndexedReader {
       throw errorWithLibrary("File is not indexed");
     }
 
+    const dataEndOffset = BigInt(
+      footer.summaryStart -
+        /* data end length */ BigInt(4) -
+        /* record content length */ BigInt(8) -
+        /* Opcode.FOOTER */ BigInt(1),
```
jtbandes, line 216: nit:
```suggestion
        /* data end length */ 4n -
        /* record content length */ 8n -
        /* Opcode.FOOTER */ 1n,
```

typescript/core/src/McapIndexedReader.ts
```diff
@@ -51,6 +54,9 @@ export class McapIndexedReader {
     this.summaryOffsetsByOpcode = args.summaryOffsetsByOpcode;
     this.header = args.header;
     this.footer = args.footer;
+    if (args.dataEndOffset != undefined) {
```
jtbandes, line 57: nit: probably don't need the `if` around this, since the property is already optional

typescript/core/src/McapWriter.ts
```diff
@@ -120,7 +129,60 @@ export class McapWriter {
...
+  static async InitializeInAppendMode({
+    writable,
+    readable,
+  }: {
+    writable: IAppendWritable;
+    readable: IReadable;
+  }): Promise<McapWriter> {
+    const reader = await McapIndexedReader.Initialize({ readable });
+
+    await writable.seek(reader.dataEndOffset!);
```
jtbandes, line 144: Generally, it would be nice to avoid `!` and instead throw meaningful error messages if required fields are missing. Under what circumstances would this be undefined?

typescript/core/src/McapWriter.ts
```diff
@@ -120,7 +129,60 @@ export class McapWriter {
...
+
+    for (const channel of reader.channelsById) {
+      await writer.#registerExistingChannel(channel[1]);
+    }
+
+    for (const chunkIndex of reader.chunkIndexes) {
+      await writer.#addChunkIndex(chunkIndex);
+    }
+
+    await writer.#setStatistics(
```
jtbandes, line 171: I think it would be cleaner and less error-prone if this function could take in the whole Statistics record from the reader as one object instead of individual params.

typescript/core/src/McapWriter.ts
```diff
@@ -469,4 +533,97 @@ export class McapWriter {
     await this.#writable.write(this.#recordWriter.buffer);
     this.#recordWriter.reset();
   }
+
+  /**
+   * Add attachment index from existing MCAP file.
+   * The purpose of this is to update the information for the new summary with the existing data from the existing MCAP file.
+   */
+  async #addAttachmentIndex(attachmentIndex: AttachmentIndex): Promise<void> {
```
jtbandes, line 541: Seems like most/all of these new methods don't actually need to be `async`

typescript/core/src/McapWriter.ts
```diff
@@ -34,6 +37,7 @@ export type McapWriterOptions = {
   startChannelId?: number;
   chunkSize?: number;
   compressChunk?: (chunkData: Uint8Array) => { compression: string; compressedData: Uint8Array };
+  appendMode?: boolean;
```
jtbandes, line 40: If this isn't meant to be set to true manually, ideally it would not be exposed in the TS types. I guess TS doesn't allow you to define multiple constructors where some are private – so I wonder if it would make sense to remove this from the constructor entirely and set #appendMode later in the InitializeInAppendMode method?

typescript/core/src/McapWriter.ts
```diff
@@ -120,7 +129,60 @@ export class McapWriter {
...
+      reader.statistics!.chunkCount,
+      reader.statistics!.channelMessageCounts,
+    );
+
+    return writer;
+  }
+
   async start(header: Header): Promise<void> {
+    if (this.#appendMode) {
+      return;
```
jtbandes, line 184: Probably should throw an error here

typescript/core/src/McapWriter.ts
```diff
@@ -130,7 +192,9 @@ export class McapWriter {
   }
 
   async end(): Promise<void> {
-    await this.#finalizeChunk();
+    if (!this.#appendMode) {
+      await this.#finalizeChunk();
```
jtbandes, line 196: Why don't we want to do this in append mode? It seems like this would lead to messages not being written out.

typescript/core/src/McapWriter.test.ts
```diff
@@ -344,4 +344,213 @@ describe("McapWriter", () => {
...
+      logTime: 0n,
+      createTime: 0n,
+      mediaType: "application/json",
+      data: new TextEncoder().encode(`{"test": "testValue"}`),
+    });
+    await writerAppendMode.addMetadata({
+      name: "metadata test",
+      metadata: new Map<string, string>([["test", "testValue"]]),
+    });
+    await writerAppendMode.end();
```
jtbandes, line 392: can we also include some `addMessage` calls?

typescript/core/src/McapWriter.ts
```diff
@@ -469,4 +533,97 @@ export class McapWriter {
...
+  async #addAttachmentIndex(attachmentIndex: AttachmentIndex): Promise<void> {
+    if (this.statistics) {
+      ++this.statistics.attachmentCount;
+    }
+
+    if (this.#attachmentIndices) {
+      this.#attachmentIndices.push(attachmentIndex);
+    }
+
+    this.#dataSectionCrc = crc32Update(this.#dataSectionCrc, this.#recordWriter.buffer);
```
jtbandes, line 550: I haven't dug in too closely but I suspect something is wrong with the CRC calculation, because I don't see the CRC being initialized in append mode.

It looks like _some_ of our readers (but not all) will verify this value when reading a mcap file (two examples I found are [python](https://github.com/foxglove/mcap/blob/main/python/mcap/mcap/stream_reader.py#L126) & [rust](https://github.com/foxglove/mcap/blob/main/rust/src/read.rs#L684)) – so you could lean on those libraries to validate the resulting crc is correct. I'm not sure if `mcap doctor` checks it.

You can also refer to the Go implementation: https://github.com/foxglove/mcap/blob/main/go/cli/mcap/utils/mcap_amendment.go

typescript/core/src/McapWriter.ts
```diff
@@ -120,7 +129,60 @@ export class McapWriter {
...
 
+  /**
+   * Initializes a new McapWriter in append mode.
+   */
+  static async InitializeInAppendMode({
+    writable,
+    readable,
+  }: {
+    writable: IAppendWritable;
+    readable: IReadable;
```
jtbandes, line 140: I'm not sure if there is specific prior art for this sort of API, but you could consider accepting these as a single object (if you are okay with forcing the caller to implement both interfaces on the same object).

You could also update the recently-added nodejs `FileHandleReadable`/`FileHandleWritable` class(es) so they can be used with this API.

typescript/core/src/IAppendWritable.ts
```diff
@@ -0,0 +1,9 @@
+import { IWritable } from "./IWritable";
+
+/**
+ * IAppendWritable describes a writer interface with append abilities.
+ */
+export interface IAppendWritable extends IWritable {
```
jtbandes, line 6: naming-wise, I'm not sure about "AppendWritable" – since being able to seek isn't necessarily related to the concept of appending. Some other ideas:

- `ISeekable`
- `ISeekableWriter`  (SeekableWritable sounds bad, but would be more consistent 😓)

typescript/core/src/McapWriter.ts
alexern14: I'm not quite sure what you mean by this. Do you have an example?

typescript/core/src/McapWriter.ts
alexern14: I added CRC initialization for append mode, but am not quite sure what to set it to. Do you have a quick answer? If not, I can continue to look into the Go implementation and figure it out.

typescript/core/src/McapWriter.ts
jtbandes: I meant like `IReadable & IAppendWritable` – since I'm not sure if I can imagine situations where the caller would actually want the reader and writer to be different objects?

alexern14:
@jtbandes Thanks for the feedback! Addressed all your comments.

jtbandes:
Sorry for the delay, this is on my radar but it might take a little while before I get more time to review it fully.

jtbandes:
Still planning on reviewing this but we've had a few other high-priority things come up. Thanks for your patience!

jtbandes:
Made various updates & fixes; opened a new PR here: https://github.com/foxglove/mcap/pull/1060

jtbandes:
@alexern14 Thanks for getting this started and being patient! Now released in v2.1.0 🎉  https://mcap.dev/docs/typescript/classes/_mcap_core.McapWriter#InitializeForAppending

PR #1017, opened 2023-11-16
Relax version restrictions of protobuf dependency

achim-k:
### Public-Facing Changes

Relax version restrictions of protobuf dependency

### Description
Changed the protobuf version restrictions to exclude the 4.22-4.24 which could lead to segmentation faults when reading a mcap file. We previously tightened it in https://github.com/foxglove/mcap/pull/919


jtbandes (approved):
This is the most advanced dependency version requirement I've ever seen 😆

PR #1018, opened 2023-11-16
Fix metadata rendering issue

narasaka:
### Description

Prior to this commit, the printed metadata was wrapped/formatted incorrectly when running `mcap list metadata`. This commit fixes it by removing the pretty json format – making the metadata column print only 1 line.

### DIFF
--- a/go/cli/mcap/cmd/metadata.go
+++ b/go/cli/mcap/cmd/metadata.go
@@ -25,13 +25,12 @@ var (
 
 func printMetadata(w io.Writer, r io.ReadSeeker, info *mcap.Info) error {
 	rows := make([][]string, 0, len(info.MetadataIndexes))
-	headers := []string{
+	rows = append(rows, []string{
 		"name",
 		"offset",
 		"length",
 		"metadata",
-	}
-	rows = append(rows, headers)
+	})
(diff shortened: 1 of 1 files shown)


go/cli/mcap/cmd/metadata.go
```diff
@@ -25,12 +25,16 @@ var (
...
 		"name",
 		"offset",
 		"length",
 		"metadata",
-	})
+	}
+	rows = append(rows, headers)
+	// jsonRowSeparator is used to replace the newline characters
+	// so that the metadata displayed is formatted and wrapped correctly
+	jsonRowSeparator := "\n" + strings.Repeat("\t", len(headers)-1) + strings.Repeat(" ", len(headers)-1)
```
narasaka, line 37: @wkalt this assumes that `metadata` will be the right-most column.

wkalt:
@narasaka this still looks misaligned:

```

[~/data/mcap] $ mcap list metadata gps.mcap
name         offset   length  metadata
my-metadata  1255462  56      {
                             "biz": "baz",
                             "foo": "bar"
                           }
```

what do you think about just putting it all on one line like in `mcap list channels`?

Issue #1002, opened 2023-11-01
Missing documentation for `offered_qos_profiles`


jhurliman:
**Description**

The ROS 2 section of MCAP's well-known profiles documentation (https://mcap.dev/spec/registry#ros2) says `offered_qos_profiles` is a required metadata field, but no documentation or example is provided for what this field should contain. Is it JSON-serialized? What are the fields?

EDIT: I found <https://github.com/ros2/rosbag2/issues/609>, it looks like the serialization format is YAML. I'm not sure if there is authoritative documentation that could be linked to, or if all of the fields should be documented on the MCAP site. Even saying "consult the ROS 2 source code for documentation on this field" would be a start.

jhurliman:
Related: https://github.com/ros2/rosbag2/issues/1420

defunctzombie:
@MichaelOrlov @emersonknapp Reading the referenced rosbag issue its not clear to me how to document this or whether the mcap spec is even the correct place to document this.

Options that come to mind:

* Update the spec to say that the field is a string but the string is ROS distro/release specific?
* Remove the documentation of the field from the mcap spec altogether though arguably that makes creating tooling for ROS 2 mcap files harder?
* Be more specific and say that the string is valid yaml? (is it?)

I also noticed that the mcap storage plugin also provides a type hash field now: https://github.com/ros2/rosbag2/blob/rolling/rosbag2_storage_mcap/src/mcap_storage.cpp#L820 which did not exist when we first wrote this profile spec.

MichaelOrlov:
@defunctzombie @jhurliman Can we refer to the section about QoS profile overriding in the Rosbag2 readme file https://github.com/ros2/rosbag2#overriding-qos-profiles and official [How-to Overriding QoS Policies in rosbag2 guide](https://docs.ros.org/en/rolling/How-To-Guides/Overriding-QoS-Policies-For-Recording-And-Playback.html)?

jhurliman:
@MichaelOrlov the only tricky part about those docs is the required format is _slightly_ different. The top-level `/topic_name:` key cannot be present, only the key-values below that. That's what the rosbag2 recorder is writing and expecting to read, anyways. The following "schema" from https://docs.ros.org/en/rolling/How-To-Guides/Overriding-QoS-Policies-For-Recording-And-Playback.html is exactly the documentation that should be provided or linked to, somehow:

```
history: [keep_all, keep_last]
depth: int
reliability: [system_default, reliable, best_effort, unknown]
[...]

PR #1020, opened 2023-11-17
Add link to ROS QoS Policy docs to spec registry

defunctzombie:
Update the `ros2` profile section of the registry with more details about the expectations for `offered_qos_profiles` and link to the official ROS docs.

Fixes: #1002

---
I do wonder if it would be more appropriate for the ROS docs to define these aspects of the profile and the MCAP registry links to those docs for details on the profile.


defunctzombie:
@MichaelOrlov @jhurliman fyi - if you have feedback.

achim-k (approved):

MichaelOrlov:
@defunctzombie My apologies, a bit busy and don't have time to consider this PR.  
However, I can get back to it after Thanksgiving holidays.

jhurliman:
LGTM

MichaelOrlov (commented):
@defunctzombie The wording and links looks good to me.

However, there are some subtle specifics that in theory could be more than one QoS profile stored in that field and it is actually a sequence of such serialized QoS profiles in yaml format. This is true for the case when exists different publishers with different QoS profiles on the same topic.
See for details https://github.com/ros2/rosbag2/blob/9855072566f8fb46a2bdd20cc217d1a0c2b03e82/rosbag2_storage_mcap/src/mcap_storage.cpp#L823-L825
```
    channel.metadata.emplace(
      "offered_qos_profiles",
      rosbag2_storage::serialize_rclcpp_qos_vector(topic_info.topic_metadata.offered_qos_profiles));
```
Maybe reword the first sentence `The sequence of the QoS policy values for the topic in yaml format`?

defunctzombie:
> Maybe reword the first sentence `The sequence of the QoS policy values for the topic in yaml format`?

👍 sounds good. Do you know what the format of this sequence would look like? Would it be a typical yaml sequence of the existing QoS "objects"? The above code makes it look like it will always be a sequence even with 1 element. Is that correct or does the serialization of the vector special case single element vectors?

MichaelOrlov:
It is going to be always a YAML sequence.
The implementation is https://github.com/ros2/rosbag2/blob/542bda34358d03acf3c15228a60859e62f5b69ba/rosbag2_storage/src/rosbag2_storage/qos.cpp#L252-L259
```
Node convert<std::vector<rclcpp::QoS>>::encode(const std::vector<rclcpp::QoS> & rhs, int version)
{
  Node node{NodeType::Sequence};
  for (const auto & value : rhs) {
    node.push_back(convert<rclcpp::QoS>::encode(value, version));
  }
  return node;
}
```

PR #1029, opened 2023-11-24
ci.yml: use OIDC connect to publish pypi packages

james-rms:
### Public-Facing Changes

None.

### Description
I previously configured OIDC in pypi and test.pypi for all of our MCAP packages, see for example https://pypi.org/manage/project/mcap-protobuf-support/settings/publishing/

This PR switches CI over to using that auth method.

Once this is done, I can invalidate our API tokens, and we can remove the PYPI_TOKEN github secrets from this repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -283,6 +283,8 @@ jobs:
     defaults:
       run:
         working-directory: python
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
👍

PR #1030, opened 2023-12-04
Update TypeScript READMEs

jtbandes:
### Public-Facing Changes

Minor updates to READMEs

### Description

`@mcap/support/nodejs` was an outdated package name. This stuff is not in the `@mcap/support` package anymore but in a separate `@mcap/nodejs` package.

Resolves FG-5991

### DIFF
--- a/typescript/nodejs/README.md
+++ b/typescript/nodejs/README.md
@@ -6,7 +6,7 @@ The `@mcap/nodejs` package provides utilities for working with MCAP files from N
 
 ## Usage examples
 
-### Reading MCAP files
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):
As someone who copy-pasted the example - it was not obvious I needed to call `.start() and then `.end()`.

jtbandes:
I agree that is a documentation gap but I don't think it is the @mcap/support or @mcap/nodejs package's responsibility to explain it – they are just showing how to _create_ an McapWriter, not how to use it

PR #1031, opened 2023-12-05
Do not create a new TextDecoder for every record

achim-k:
### Public-Facing Changes

Do not create a new TextDecoder for every record

### Description
Massively improves read performance by avoiding unnecessary creation / cleanups of TextDecoders.

Before:
```
 [Summary]:
   ticks  total  nonlib   name
   6600    7.2%   46.1%  JavaScript
   7693    8.4%   53.8%  C++
  21411   23.3%  149.6%  GC
  77489   84.4%          Shared libraries
     16    0.0%          Unaccounted
```

After (Notice the drop in GC time):
```
 [Summary]:
   ticks  total  nonlib   name
   3950   10.1%   60.7%  JavaScript
   2555    6.6%   39.2%  C++
    549    1.4%    8.4%  GC
  32426   83.3%          Shared libraries
      5    0.0%          Unaccounted
```

### DIFF
--- a/typescript/core/src/Reader.ts
+++ b/typescript/core/src/Reader.ts
@@ -1,5 +1,8 @@
 import { getBigUint64 } from "./getBigUint64";
 
+// For performance reasons we use a single TextDecoder instance whose internal state is merely
+// the encoding (defaults to UTF-8). This means that a TextDecoder.decode() call is not affected
+// be previous calls.
 const textDecoder = new TextDecoder();
 
 export default class Reader {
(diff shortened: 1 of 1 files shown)


typescript/core/src/Reader.ts
```diff
@@ -1,9 +1,10 @@
 import { getBigUint64 } from "./getBigUint64";
 
+const textDecoder = new TextDecoder();
```
defunctzombie, line 3: We should add a comment about why we are using a global decoder.

I'm not aware of any issue with doing this (I've long hated that the API requires a new instance each time) but it could be that the decoder holds onto some internal state for longer than we'd want?

achim-k: added a comment. The encoding seems to be the TextDecoder's only internal state

jtbandes (approved):

PR #1036, opened 2023-12-08
@mcap/core: Avoid heap operations when chunks are ordered

achim-k:
### Public-Facing Changes

@mcap/core: Avoid heap operations when chunks are ordered

### Description
When chunks are ordered and not overlapping, we do not need to re-evaluate the heap after every message and can instead keep on reading from the current chunk.


jtbandes (approved):
Makes sense to me.

Do we have tests that catch this? i.e. if you incorrectly set `chunksOrdered = true` will a test fail?

achim-k:
> Do we have tests that catch this? i.e. if you incorrectly set `chunksOrdered = true` will a test fail?

Yes, there is a test failure then

PR #1037, opened 2023-12-08
@mcap/core: Replace message index heap with sorted list

achim-k:
### Public-Facing Changes

@mcap/core: Replace message index heap with sorted list

### Description
Instead of using a heap for storing message offsets, we read all message offsets into a sorted array (sorted by logTime) which allows for faster retrieval of next messages.

### DIFF
--- a/typescript/core/src/sortedLastIndex.ts
+++ b/typescript/core/src/sortedLastIndex.ts
@@ -2,7 +2,6 @@
  * Return the lowest index of `array` where an element can be inserted and maintain its sorted
  * order. This is a specialization of lodash's sortedIndexBy().
  */
-
 export function sortedLastIndexBy(
   array: [bigint, bigint][],
   value: bigint,


typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +150,78 @@ export class ChunkCursor {
...
+  }
+
+  #getSortTime(): bigint {
+    if (
+      this.#orderedMessageOffset != undefined &&
+      this.#orderedMessageOffset.length > 0 &&
+      this.#messageOffsetIndex < this.#orderedMessageOffset.length
+    ) {
+      return this.#orderedMessageOffset[this.#messageOffsetIndex]![0];
+    } else {
```
defunctzombie, line 223: nit: You can drop this _else_ since the above _if_ returns. Consider adding a comment above the _if_ to help clarify a rather complex conditional expression.

typescript/core/src/ChunkCursor.ts
```diff
@@ -36,7 +25,8 @@ export class ChunkCursor {
   #endTime: bigint | undefined;
   #reverse: boolean;
 
-  #messageIndexCursors?: Heap<MessageIndexCursor>;
+  #orderedMessageOffset?: [logTime: bigint, offset: bigint][];
+  #messageOffsetIndex = 0;
```
defunctzombie, line 29: Lets add a comment on this and how it is used - my understanding of this from the implementation is that it is the next message offset index that we need to use to produce/read the next message?

defunctzombie (commented):
Do you have some performance numbers (cpu or memory) that you can share showing how this helps file reading?

typescript/core/src/ChunkCursor.ts
achim-k: I don't understand this comment. If message indices haven't been loaded yet then `this.#orderedMessageOffset` is undefined and the `else` branch is taken.

typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +150,78 @@ export class ChunkCursor {
...
 
-    return cursor.records[cursor.index]![0];
+    const [logTimeLastMessage] = this.#orderedMessageOffset[this.#orderedMessageOffset.length - 1]!;
+    if (logTimeLastMessage > this.chunkIndex.messageEndTime) {
+      throw new Error(
+        `Chunk contains a message with logTime with logTime (${logTimeLastMessage}) later than chunk messageEndTime (${this.chunkIndex.messageEndTime}) in chunk at offset ${this.chunkIndex.chunkStartOffset}`,
+      );
+    }
+
+    // Remove message offsets whose log time is not within [startTime, endTime];
```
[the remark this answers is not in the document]
achim-k: To determine the ChunkCursor's first mesage logTime which is returned in `#getSortTime` and used for ordering ChunkCursors in the heap

typescript/core/src/ChunkCursor.ts
defunctzombie: You don't have to change it if you prefer this - my comment was a code style nit that because the _if_ condition returns, you can drop the _else_ block and put its return under the _if_. To me that makes the code read more like "here's the normal return path" but if these things are true we might return earlier.

```
    if (
      this.#orderedMessageOffset != undefined &&
      this.#orderedMessageOffset.length > 0 &&
      this.#messageOffsetIndex < this.#orderedMessageOffset.length
    ) {
      return this.#orderedMessageOffset[this.#messageOffsetIndex]![0];
    {

    return this.#reverse ? this.chunkIndex.messageEndTime : this.chunkIndex.messageStartTime;
```

typescript/core/src/ChunkCursor.ts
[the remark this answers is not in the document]
achim-k: I changed this part now a bit to make it clearer. Instead of pruning the message offset list from unused offsets (which are not in the range of [startTime, endTime]) I only determine the start & endIndex which might also be a bit more performant.

achim-k:
> Do you have some performance numbers (cpu or memory) that you can share showing how this helps file reading?

I tested this on a couple of different mcap files and the speed improvement is about 5-10%. I didn't measure memory consumption, but I think it should be the same or slightly less than the heap variant.

typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +160,75 @@ export class ChunkCursor {
...
 
-  #getSortTime(): bigint {
-    if (!this.#messageIndexCursors) {
-      return this.#reverse ? this.chunkIndex.messageEndTime : this.chunkIndex.messageStartTime;
+    if (this.#orderedMessageOffset.length === 0) {
+      return;
+    }
+
+    this.#orderedMessageOffset.sort(([logTimeA], [logTimeB]) => Number(logTimeA - logTimeB));
+    if (reverse) {
```
james-rms, line 177: This is awkward - the message indexes in an MCAP are not specified as being in the same order as the messages, so you can't guarantee after sorting by timestamp that the messages with the same timestamp will be in file-offset order. This is also true in reverse.

I think you need to compare by log time and file offset when sorting.

typescript/core/src/ChunkCursor.ts
```diff
@@ -36,7 +25,12 @@ export class ChunkCursor {
   #endTime: bigint | undefined;
   #reverse: boolean;
 
-  #messageIndexCursors?: Heap<MessageIndexCursor>;
+  // List of message offsets (across all channels) sorted by logTime.
+  #orderedMessageOffset?: [logTime: bigint, offset: bigint][];
```
jtbandes, line 29:
```suggestion
  #orderedMessageOffsets?: [logTime: bigint, offset: bigint][];
```

typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +160,75 @@ export class ChunkCursor {
...
+    const startTime = reverse ? this.#endTime : this.#startTime;
+    const endTime = reverse ? this.#startTime : this.#endTime;
+    const iteratee = (logTime: bigint) => (reverse ? -logTime : logTime);
+    if (startTime != undefined) {
+      this.#nextMessageOffsetIndex = sortedIndexBy(this.#orderedMessageOffset, startTime, iteratee);
+    }
+    if (endTime != undefined) {
+      this.#messageOffsetEndIndex = sortedIndexBy(this.#orderedMessageOffset, endTime, iteratee);
+      // sortedIndexBy returns the minimum index but for the end index we actually want the highest index since
+      // endTime is inclusive. So we count up the end index manually until we reach a logTime that is not included anymore.
```
jtbandes, line 208: It appears the lodash implementation (on which our sortedIndexBy is based) includes a flag for whether the highest value should be returned: https://github.com/lodash/lodash/blob/aa18212085c52fc106d075319637b8729e0f179f/src/.internal/baseSortedIndexBy.ts#L16

They use this to implement `sortedLastIndexBy`: https://github.com/lodash/lodash/blob/aa18212085c52fc106d075319637b8729e0f179f/src/sortedLastIndexBy.ts#L23

Seems like it might be slightly nicer (and slightly faster in cases when there are many messages with the same timestamp) to adopt this approach.

On the other hand it's probably not a very big impact.

typescript/core/src/ChunkCursor.ts
jtbandes: This seems like a good callout, but I don't think the behavior is changing here. It looks like the [heap we were previously using](https://github.com/foxglove/mcap/pull/1037/files#diff-cc829f0c49b6423567981a8ce9084d8d3d282c1faff58b0cd69830e8c43c7b59L143) didn't have a fallback comparator for file offset order.  On the other hand, we do have this in `compare()` already (for the higher-level heap across chunks): https://github.com/foxglove/mcap/blob/6986d8b32e9840fed5e7c18de7a20ce2fcd0a1c1/typescript/core/src/ChunkCursor.ts#L70

typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +160,75 @@ export class ChunkCursor {
...
+      // If we used `logTimeB - logTimeA` as the comparator for reverse iteration, messages with
+      // the same timestamp would not be in reverse order. To avoid this problem we use reverse()
+      // instead.
+      this.#orderedMessageOffset.reverse();
+    }
+
+    const [logTimeFirstMessage] = this.#orderedMessageOffset[0]!;
+    if (logTimeFirstMessage < this.chunkIndex.messageStartTime) {
+      throw new Error(
+        `Chunk contains a message with logTime (${logTimeFirstMessage}) earlier than chunk messageStartTime (${this.chunkIndex.messageStartTime}) in chunk at offset ${this.chunkIndex.chunkStartOffset}`,
```
jtbandes, line 187: wording nit: these errors are mentioning the chunk twice

```suggestion
        `Chunk at offset ${this.chunkIndex.chunkStartOffset} contains a message with logTime (${logTimeFirstMessage}) earlier than chunk messageStartTime (${this.chunkIndex.messageStartTime})`,
```

typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +160,75 @@ export class ChunkCursor {
...
-          continue;
-        }
-      }
-
-      this.#messageIndexCursors.push({
-        index: startIndex,
-        channelId: result.record.channelId,
-        records: result.record.records,
-      });
+      this.#orderedMessageOffset = this.#orderedMessageOffset.concat(result.record.records);
```
jtbandes, line 163: Consider using `push` rather than `concat` for better performance.

It's possible that using `.flat()` on an array of arrays would be even faster depending on how it is implemented.

typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +160,75 @@ export class ChunkCursor {
...
       throw new Error(
-        `Unexpected empty cursor for chunk at offset ${this.chunkIndex.chunkStartOffset}`,
+        `Chunk contains a message with logTime with logTime (${logTimeLastMessage}) later than chunk messageEndTime (${this.chunkIndex.messageEndTime}) in chunk at offset ${this.chunkIndex.chunkStartOffset}`,
       );
     }
 
-    return cursor.records[cursor.index]![0];
+    // Determine the indexes corresponding to the start and end time.
+    const startTime = reverse ? this.#endTime : this.#startTime;
+    const endTime = reverse ? this.#startTime : this.#endTime;
```
jtbandes, line 200: Did you try a version where you filter out entries outside of [startTime,endTime] before putting them into the array and sorting? That would certainly reduce memory usage a little bit, and reduce CPU since not as many items are being sorted – maybe not significant enough though? It would also mean you can do less bookkeeping around the endIndex.

typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +160,75 @@ export class ChunkCursor {
...
+    const iteratee = (logTime: bigint) => (reverse ? -logTime : logTime);
```
jtbandes, line 201: might be faster: `reverse ? ((logTime: bigint) => -logTime) : ((logTime: bigint) => logTime)`

typescript/core/src/ChunkCursor.ts
achim-k: Before https://github.com/foxglove/mcap/pull/1037/commits/ef37a836761fc223faaf965462e9e6a03c4f6a29, I was filtering out entries that are not within the desired time range. Maybe I should go back to that and do as you say and already filter out elements of each message index before adding them to the final array.

typescript/core/src/ChunkCursor.ts
```diff
@@ -26,11 +27,9 @@ export class ChunkCursor {
   #reverse: boolean;
 
   // List of message offsets (across all channels) sorted by logTime.
-  #orderedMessageOffset?: [logTime: bigint, offset: bigint][];
+  orderedMessageOffsets?: [logTime: bigint, offset: bigint][];
```
defunctzombie, line 30: Did you mean to make public by omitting the `#`?

achim-k: ah, no! good catch

achim-k:
I tried to figure out in theory with O(N) notation if the sorted array implementation is superior over the previous heap implementation. Not sure if I got it right, but the result (and the experimental results) leans towards the sorted array implementation.

---


**M** : Number of message indexes
**X** : Number of messages in a message index
**N** : Total number of messages (= M * X)

Previous heap implementation:
- Sorting **M** arrays of size **X**: `O(M * X * log X) = O(N * log X)`
- Building heap of size **M**: `O(M)`
- Replacing **N** elements from heap of size **M**: `O(N * log M)`

Sorted array implementation:
- Sorting array of size **N**: `O (N * log N)`
- Getting next message: `O(1)`

Changing the implementation to a sorted array makes only sense if `O(sorted array) < O(heap implementation)`. And that is the case:

```
N * log N < N * log X + M + N * log M
N * log N < N * (log X + log M) + M
N * log N < N * log (X * M) + M
N * log N < N * log N + M
        0 < M
```

typescript/core/src/ChunkCursor.ts
```diff
@@ -211,80 +153,87 @@ export class ChunkCursor {
...
+    if (
+      this.#orderedMessageOffsets != undefined &&
+      this.#orderedMessageOffsets.length > 0 &&
+      this.#nextMessageOffsetIndex < this.#orderedMessageOffsets.length
+    ) {
+      return this.#orderedMessageOffsets[this.#nextMessageOffsetIndex]![0];
+    }
+
+    // Fall back to the chunk index' start time or end time.
+    return this.#reverse ? this.chunkIndex.messageEndTime : this.chunkIndex.messageStartTime;
```
defunctzombie, line 237: This should use the first or last item of `orderedMessageOffsets` since that is the actual available message time to read. The old logic used peek to look at the heap and I think to mirror that you would use your ordered offsets array.

typescript/core/src/sortedLastIndex.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Return the lowest index of `array` where an element can be inserted and maintain its sorted
+ * order. This is a specialization of lodash's sortedIndexBy().
+ */
+
```
defunctzombie, line 5:
```suggestion
```

defunctzombie (approved):
One more question on the `getSortTime` return value and I think this is good.

I guess no test had issue with the return value or maybe there's a gap in the testing.

wkalt:
I think for this problem the real cheat is figuring out a way for the writer to convey to the reader, through the EOF chunk index, that a chunk is fully-ordered. That is the case with 99%+ of chunks in my guess, and would allow readers to fully skip all the heap stuff, skip message indexes, and just read chunks in linear order.

defunctzombie:
> allow readers to fully skip all the heap stuff, skip message indexes, and just read chunks in linear order

Any reader that wants to read specific topics will need to consult the message index rather than reading the chunk in message order.

PR #1041, opened 2023-12-12
Website: Add Slack link to header

amacneil:
### Public-Facing Changes
Add Slack link to website header.

Rationale: Make it easier to find.


jtbandes (approved):
Is there a way to make a slack link that goes to a specific channel?

amacneil:
> Is there a way to make a slack link that goes to a specific channel?

You can link directly to a channel, but I'm not aware of any ability to combine that with invite links.

PR #1042, opened 2023-12-14
Fix internal links in spec index.md

jameskuszmaul-brt:
### Public-Facing Changes

Fixes internal links in https://mcap.dev/spec

### Description

URLs like https://mcap.dev/spec#data-end-op0x0F are incorrect; they should be lower-case (like https://mcap.dev/spec#data-end-op0x0f).

Noticed on the live website.


defunctzombie (approved):
Thank you!

PR #1047, opened 2024-01-02
docs: prioritize converting db3 files with ros2 bag convert

james-rms:
### Public-Facing Changes

Moves the `ros2 bag convert` instructions for converting ROS 2 bags to the top. This is the better-supported method and we should be pointing users to it first.

### DIFF
--- a/website/docs/guides/getting-started/ros-2.md
+++ b/website/docs/guides/getting-started/ros-2.md
@@ -30,29 +30,9 @@ $ ros2 bag record -s mcap --all --storage-config-file my_storage_config.yaml
 
 ## Convert to MCAP
 
-### Using the `mcap` CLI tool
-
-To convert your existing ROS 2 db3 files into MCAP files, [install the mcap CLI tool](https://github.com/foxglove/mcap/tree/main/go/cli/mcap#installing) and run the following command:
-
-```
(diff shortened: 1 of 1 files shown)


website/docs/guides/getting-started/ros-2.md
```diff
@@ -30,15 +30,31 @@ $ ros2 bag record -s mcap --all --storage-config-file my_storage_config.yaml
...
+    storage_id: mcap
+    all: true
+EOF
+$ ros2 bag convert -i ros2_input.db3 -o convert.yaml
+```
+
 ### Using the `mcap` CLI tool
 
-To convert your existing ROS 2 db3 files into MCAP files, [install the mcap CLI tool](https://github.com/foxglove/mcap/tree/main/go/cli/mcap#installing) and run the following command:
+The [`mcap` CLI tool](https://github.com/foxglove/mcap/tree/main/go/cli/mcap#installing) also supports converting ROS 2 `.db3` files into MCAP files directly.
```
amacneil, line 49: We should probably call out that if you have difficulty getting the mcap cli to discover message definitions, then try the rosbag2 method as preferred approach.

amacneil (approved):

PR #1049, opened 2024-01-05
CLI: Fix schema deduplication in mcap convert

wkalt:
Prior to this commit, if mcap convert got two ROS1 connections with the same schema but different topic, it would write two schema records.

In this situation only one schema record is needed, as long as the type and the md5 sum are aligned.

### DIFF
--- a/go/ros/bag2mcap.go
+++ b/go/ros/bag2mcap.go
@@ -260,7 +260,7 @@ func Bag2MCAP(w io.Writer, r io.Reader, opts *mcap.WriterOptions, messageCallbac
 			msgdef := connectionDataHeader["message_definition"]
 			delete(connectionDataHeader, "message_definition")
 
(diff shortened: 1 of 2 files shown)


go/ros/bag2mcap.go
```diff
@@ -260,7 +260,7 @@ func Bag2MCAP(w io.Writer, r io.Reader, opts *mcap.WriterOptions, messageCallbac
 			msgdef := connectionDataHeader["message_definition"]
 			delete(connectionDataHeader, "message_definition")
 
-			key := fmt.Sprintf("%s/%s", topic, connectionDataHeader["md5sum"])
+			key := fmt.Sprintf("%s/%s", typ, connectionDataHeader["md5sum"])
```
wkalt, line 263: encoding would also be required, except it's always ros1msg for ros bags.

james-rms (approved):

PR #1050, opened 2024-01-08
Go: do not implicitly close custom decompressor in lexer

wkalt:
This removes some implicit behavior when using the lexer with a custom decompressor. Prior to this commit, if your custom decompressor exposed a Close() method it would get closed on close of the lexer. After this commit, the caller is responsible for that. The benefit of this is they will be better able to reuse their decompressor if desired.


wkalt:
this is not an interface change but it will change publicly documented behavior. I suspect that we (foxglove) are the only consumers of this functionality, however.

james-rms (approved):

PR #1052, opened 2024-01-11
CLI: Append attachments to merged mcap files

bryfox:
This updates `mcap merge` to include attachments from input files.

Attachments are appended in the order they are encountered, and may be out of order with respect to log time. This procedure works differently than metadata because of the way we process attachment opcodes. 

- If an input file contains a summary statistics record and attachment indexes, we scan any attachments from the those indexes
- If a summary statistics record contains an attachmentCount of 0, no work is done
- Otherwise, for unindexed inputs, this scans the input file a second time and processes attachments using the lexer's attachment callback. This avoids changing any public interface NewReader.

We considered modifying the NewReader interface to accept an attachment callback option, which would simplify the merging code (and avoid a second scan on unindexed files), but deemed it better to leave the interface alone.

### DIFF
--- a/go/cli/mcap/cmd/merge.go
+++ b/go/cli/mcap/cmd/merge.go
@@ -369,40 +369,53 @@ func (m *mcapMerger) mergeInputs(w io.Writer, inputs []namedReader) error {
 		heap.Push(pq, utils.NewTaggedMessage(msg.InputID, newMessage))
 	}
 
-	// merge any indexed attachments
+	// scan the input again for any attachments
 	for _, input := range inputs {
 		_, err := input.reader.Seek(0, io.SeekStart)
 		if err != nil {
 			return fmt.Errorf("failed to seek: %w", err)
 		}
-		reader, err := mcap.NewReader(input.reader)
+		lexer, err := mcap.NewLexer(input.reader, &mcap.LexerOptions{
+			ComputeAttachmentCRCs: true,
+			AttachmentCallback: func(attReader *mcap.AttachmentReader) error {
+				err := writer.WriteAttachment(&mcap.Attachment{
+					LogTime:    attReader.LogTime,
+					CreateTime: attReader.CreateTime,
+					Name:       attReader.Name,
+					MediaType:  attReader.MediaType,
+					DataSize:   attReader.DataSize,
+					Data:       attReader.Data(),
+				})
+				if err != nil {
+					return fmt.Errorf("failed to write attachment: %w", err)
+				}
+				computed, err := attReader.ComputedCRC()
+				if err != nil {
+					return fmt.Errorf("failed to compute CRC: %w", err)
+				}
+				parsed, err := attReader.ParsedCRC()
+				if err != nil {
+					return fmt.Errorf("failed to parse CRC: %w", err)
+				}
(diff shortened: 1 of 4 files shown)


go/mcap/reader.go
```diff
@@ -222,14 +222,25 @@ func (r *Reader) Close() {
...
+	options := ReaderOptions{}
+	for _, opt := range opts {
+		err := opt(&options)
+		if err != nil {
+			return nil, err
+		}
+	}
+	var lexerOpts *LexerOptions
+	if options.LexerOpts != nil {
+		lexerOpts = options.LexerOpts
```
wkalt, line 239: would this cause a problem if options were supplied that didn't include emit chunks?

go/mcap/reader_options.go
```diff
@@ -4,6 +4,19 @@ import (
...
+}
+
+type ReaderOpt func(*ReaderOptions) error
+
+func WithLexerOptions(lexerOpts *LexerOptions) ReaderOpt {
+	return func(ro *ReaderOptions) error {
+		ro.LexerOpts = lexerOpts
+		return nil
+	}
+}
```
wkalt, line 18: I'm not totally sure this is the best config option to publicly expose. The reader uses a lexer in two places I believe. This config is for one of them. But I'm not sure that detail is good for us to expose to the user. If we need to add an option, something explicit like "AttachmentCallback" seems better/more robust to change IMO.

I'm not totally sold that putting this functionality into the Reader is the way to go. I think we could even get by without reader changes, if in the CLI we iterate over the attachment indexes in each Reader (obtained with the Info) call, and then use the `reader.GetAttachment` function to access them and stream them into the output writer.

bryfox: > iterate over the attachment indexes

I like this idea, but I assume this requires an indexed mcap (which `merge` does not today)?  Same problem if I set up an attachment callback (similar to metadataCallback) in ReadOpt, at least with the current lexer implementation. Currently, the only hook into OpAttachment is the lexer's callback.

I pushed a commit that explores this; maybe we can discuss based on that.

bryfox: Discussed offline: a reasonable compromise is to make a second pass over the file just using a lexer, and let its attachment callback process the attachments.

go/cli/mcap/cmd/merge.go
```diff
@@ -367,12 +368,124 @@ func (m *mcapMerger) mergeInputs(w io.Writer, inputs []namedReader) error {
...
+		}
+
+		reader, err := mcap.NewReader(input.reader)
+		if err != nil {
+			return fmt.Errorf("failed to create attachment reader: %w", err)
+		}
+		defer reader.Close() //nolint:gocritic // we actually want these deferred in the loop.
+		info, err := reader.Info()
+		if err == nil && info != nil && info.Statistics != nil {
+			if info.Statistics.AttachmentCount > 0 {
```
james-rms, line 387: we should be checking for `info.Statistics.AttachmentCount != 0 && info.Statistics.AttachmentCount == len(info.AttachmentIndexes)`. This accounts for mcaps that have a statistics record but no attachment indexes.

james-rms (approved):

PR #1053, opened 2024-01-11
Go: Fixes bug in metadata handling for indexed iteration

wkalt:
Prior to this commit, when using the "metadata callback" feature of the indexed message iterator, the reader would mistakenly interpret a byte as a "Token" rather than an "OpCode", causing the parsing of record types to get confused.

This wasn't surfaced in the CLI metadata merging support because that uses the unindexed iterator, which is not affected.

This commit fixes the bug and adds a test for both cases.

### DIFF
--- a/go/mcap/reader_test.go
+++ b/go/mcap/reader_test.go
@@ -552,6 +552,53 @@ func find[T any](items []T, f func(T) bool) (val T, err error) {
 	return val, fmt.Errorf("not found")
 }
 
+func TestReaderMetadataCallback(t *testing.T) {
+	cases := []struct {
+		assertion string
+		useIndex  bool
+	}{
+		{
+			"using index",
+			true,
+		},
+		{
+			"without index",
+			false,
+		},
+	}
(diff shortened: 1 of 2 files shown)


go/mcap/reader_test.go
```diff
@@ -394,150 +394,49 @@ func TestReaderCounting(t *testing.T) {
...
-		schemas     []*Schema
-		channels    []*Channel
-		messages    []*Message
-		metadata    []*Metadata
-		attachments []*Attachment
+		assertion string
+		useIndex  bool
 	}{
 		{
-			"no metadata or attachments",
```
bryfox, line 407: Hard to tell what's being removed in this test — I assume this is unneeded for metadata testing, but are we losing coverage for something else (since this test used to be called something else?)

bryfox (approved):

go/mcap/reader_test.go
wkalt: weird - I wasn't exepecting anything removed

PR #1058, opened 2024-01-31
mcap convert: Fix duplicate type definition in ROS2 schemas

achim-k:
### Public-Facing Changes
FIx `mcap convert` producing ROS2 schemas with duplicate type definitions

### Description
It could happen that `mcap convert` produced ROS2 schemas with duplicated subtype definitions if the subtype was included multiple times by various `.msg` files. This PR fixes this by only processing subtypes that haven't been processed before.

Resolves FG-6369

### DIFF
--- a/go/ros/ros2db3_to_mcap.go
+++ b/go/ros/ros2db3_to_mcap.go
@@ -71,7 +71,7 @@ func getSchemas(directories []string, types []string) (map[string][]byte, error)
 		}{
 			{parentPackage: rosPackage, rosType: rosType, schema: schema},
 		}
-		seenSubtypes := make(map[string]struct{})
+		seenSubtypes := map[string]struct{}{rosType: struct{}{}}
 		first := true
 		for len(subdefinitions) > 0 {
 			subdefinition := subdefinitions[0]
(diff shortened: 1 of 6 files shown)


go/ros/ros2db3_to_mcap.go
```diff
@@ -71,6 +71,7 @@ func getSchemas(directories []string, types []string) (map[string][]byte, error)
 		}{
 			{parentPackage: rosPackage, rosType: rosType, schema: schema},
 		}
+		seenSubtypes := make(map[string]struct{})
```
jtbandes, line 74: the magic of go...

do we need to start with an initial entry in this set? or maybe it would only matter if recursive types were allowed?

jtbandes (approved):

go/ros/ros2db3_to_mcap.go
achim-k: > or maybe it would only matter if recursive types were allowed?

Are they allowed? I guess we would have run in infinite loops / max. call stack exceeded issues if this would be the case.

jtbandes: I don't think they are allowed in ROS. I just wanted to think through if there are any other cases we should test to make sure this works

jtbandes: https://github.com/ros2/ros2/issues/1445

achim-k: Then I better add it now, doesn't hurt

PR #1059, opened 2024-02-13
go: reader: fall back to linear scan if statistics show messages outside chunks

james-rms:
### Public-Facing Changes

* the go `mcap.Reader` interface now detects unindexed MCAPs and reads messages from them with a linear scan.

### Description

Previously when reading a seekable MCAP file with the go library Reader, it would assume that the MCAP was indexed and return no messages if not. After this PR, if an MCAP is not indexed but contains a statistics record, the reader will use the statistics to determine if there are un-indexed messages in the file.

Fixes https://linear.app/foxglove/issue/FG-6441/mcap-cat-doesnt-display-anything-for-unchunked-file

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,8 @@
 // -*- jsonc -*-
 {
-  "editor.codeActionsOnSave": { "source.fixAll.eslint": true },
+  "editor.codeActionsOnSave": {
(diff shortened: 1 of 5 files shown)


jtbandes (approved):
Seems like a sensible small change.

bryfox (approved):

PR #1060, opened 2024-02-13
typescript: add McapWriter.InitializeForAppending()

jtbandes:
### Public-Facing Changes

- Added `InitializeForAppending()` to `McapWriter`.
- `McapIndexedReader` now exposes `dataEndOffset` and `dataSectionCrc`.

### Description

Append mode works by using `McapIndexedReader` to read the summary section, loading all indexes, channels, etc. from the footer into memory, then chopping off everything from DataEnd onwards and continuing to write from there.

This does not work if the `DataEnd` record contains extra data or padding. This seems to be a fundamental flaw with the spec; a separate PR will be raised to change the spec to disallow this.

See also: https://github.com/foxglove/mcap/blob/4d62967abe95e755b91f16cce390f6560d18e07e/go/cli/mcap/utils/mcap_amendment.go

Supersedes / closes https://github.com/foxglove/mcap/pull/1016
Resolves FG-5821


typescript/core/src/McapIndexedReader.test.ts
```diff
@@ -142,15 +143,15 @@ describe("McapIndexedReader", () => {
...
+        ...uint64LE(BigInt(summaryStart)), // summary start
+        ...uint64LE(0n), // summary offset start
         ...uint32LE(crc32(new Uint8Array([42]))), // summary crc
       ]),
       ...MCAP_MAGIC,
     );
     const readable = makeReadable(new Uint8Array(data));
     await expect(McapIndexedReader.Initialize({ readable })).rejects.toThrow(
-      "Incorrect summary CRC 491514153 (expected 163128923)",
+      "Incorrect summary CRC 1656343536 (expected 163128923)",
```
achim-k, line 154: this changed because of the additional `DataEnd` record?

typescript/core/src/McapIndexedReader.test.ts
```diff
@@ -130,6 +130,7 @@ describe("McapIndexedReader", () => {
         ...string(""), // profile
         ...string(""), // library
       ]),
+      ...record(Opcode.DATA_END, [...uint32LE(0)]),
```
achim-k, line 133: Previously the reader wouldn't detect a missing `DataEnd` record and now it does?

typescript/core/src/McapWriter.test.ts
```diff
@@ -344,4 +355,483 @@ describe("McapWriter", () => {
       },
     ]);
   });
+
+  it("supports append mode", async () => {
```
achim-k, line 359: this is a super long test. Maybe we could split them up in multiple `can append <schema,channel,message,...>`?

achim-k (approved):

defunctzombie (approved):
I'm 👍 on this once you and @achim-k have settled on any of the outstanding feedback items you have

typescript/core/src/McapIndexedReader.ts
```diff
@@ -218,14 +226,27 @@ export class McapIndexedReader {
...
+    const dataEndAndSummarySection = await readable.read(
+      dataEndOffset,
+      footerOffset - dataEndOffset,
     );
     if (footer.summaryCrc !== 0) {
       let summaryCrc = crc32Init();
-      summaryCrc = crc32Update(summaryCrc, allSummaryData);
+      summaryCrc = crc32Update(
+        summaryCrc,
+        dataEndAndSummarySection.subarray(Number(dataEndLength)),
```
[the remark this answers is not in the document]
jtbandes: It does not make a copy: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/subarray

typescript/core/src/McapIndexedReader.test.ts
jtbandes: yep!

typescript/core/src/McapIndexedReader.test.ts
jtbandes: yes, indirectly, because the summary CRC includes the CRC of `footer.summary_start + footer.summary_offset_start`

typescript/core/src/McapWriter.ts
```diff
@@ -120,23 +126,85 @@ export class McapWriter {
...
+
   async start(header: Header): Promise<void> {
+    if (this.#appendMode) {
+      throw new Error(`Cannot call start() when writer is in append mode`);
+    }
     this.#recordWriter.writeMagic();
     this.#recordWriter.writeHeader(header);
 
-    this.#dataSectionCrc = crc32Update(this.#dataSectionCrc, this.#recordWriter.buffer);
+    if (this.#dataSectionCrc != undefined) {
```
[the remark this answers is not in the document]
jtbandes: Not sure if I fully understand the question – but if I try to comment out this line I get a typescript error on the next line because `#dataSectionCrc` is `number | undefined`. In the normal non-appending mode, it is initialized to `crc32Init()` (`~0`). In append mode, if the dataSectionCrc from the original file is unknown (i.e. zero), we don't attempt to produce an updated value because it would be incorrect (unless we scan back to the beginning of the file and calculate the CRC from scratch).

typescript/core/src/McapWriter.test.ts
jtbandes: I'm not sure if splitting them up would help since each of those tests would also be quite long (if we are testing the full file contents before & after appending). One option would be to move some of the expected values into a test fixtures file and import them, or use jest `expect.toMatchSnapshot` (so they are moved into a snap file). What do you think?

typescript/core/src/McapWriter.ts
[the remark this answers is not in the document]
achim-k: My question was if we can't simply do
```
this.#dataSectionCrc = crc32Init();
```
since this function is only called in the normal non-appending mode.

typescript/core/src/McapWriter.test.ts
achim-k: Sounds like quite some work, I don't mind if it stays as it is

typescript/core/src/McapWriter.ts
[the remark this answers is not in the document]
jtbandes: 👍 I like that, will remove `crc32Init()` from the member initializer and move it here.

PR #1061, opened 2024-02-13
spec: specify that DataEnd will not be extended

jtbandes:
### Public-Facing Changes

The spec has been updated to say that DataEnd records cannot have extra data or padding in them. This will not affect most readers or writers of MCAP files, but enables workflows for appending data to existing indexed files.

### Description

This change allows readers to assume that `Footer.summary_offset - 13` is the beginning of the `DataEnd` record. Without it, this line of code is wrong: https://github.com/foxglove/mcap/blob/4d62967abe95e755b91f16cce390f6560d18e07e/go/cli/mcap/utils/mcap_amendment.go#L45-L46

See also: #1060


defunctzombie (approved):

james-rms (approved):

defunctzombie:
There's an argument here that this is not binary breaking. The spec says:

> Records may be extended by adding new fields at the end of existing fields.

However, we have not made any such changes to any records yet and the proposed change here is a tightening of the restriction on the `DataEnd` record from accepting such a change in the future. Existing readers that adhere to the prior language won't be able to make the assumption about `DataEnd` until they update to the new spec language but nothing in this change breaks the binary format.

PR #1065, opened 2024-02-21
Add CMake support to cpp module

filippobrizzi:
### Public-Facing Changes
None

### Description
Add CMake support for the top-level cpp module.

This allows to use mcap in projects that do not support/use conan. 

The CMake file generates the install config files that allow running `find_package(mcap)` in external projects. 

NOTES:
* It is possible to update `examples/` and `test/` to be part of the global CMake and stop using conan to build them.
  * If this is needed, I can do it in this PR.
* I haven't updated the README, as soon as I get greenlight for this PR I will do it.


filippobrizzi:
PTAL @jtbandes @wkalt @james-rms

indirectlylit:
Hi @filippobrizzi, thank you for the contribution! Improved support for our CMake users is important to us. It may be a little while before we get to this, and we appreciate your patience in the meantime.

One thing that will be necessary eventually is a CI step.

At a high level, we'd like to provide CMake scaffolding for folks to build off of without maintaining something for all architectures and environments. It's possible that we could even migrate from conan to cmake internally if we can do it in a way that doesn't impact existing conan users.

Any feedback would be appreciated!

cpp/Makefile
```diff
@@ -71,7 +71,7 @@ test-examples-host:
 
 .PHONY: ci-image
 ci-image: hdoc-build
-	docker build -t mcap_cpp_ci -f ci.Dockerfile --build-arg IMAGE=ubuntu:focal --build-arg CC=clang-13 --build-arg CXX=clang++-13 .
+	docker build -t mcap_cpp_ci -f ci.Dockerfile --build-arg IMAGE=ubuntu:jammy --build-arg CC=clang-13 --build-arg CXX=clang++-13 .
```
filippobrizzi, line 74: using latest ubuntu to get latest cmake version needed to create an interface library

filippobrizzi:
Hi @indirectlylit thanks for the reply. 
I have done the following as a demonstration of a possible solution:
- mcap/cpp is an header only library so there is nothing much to test here as nothing compile
- to test that the CMake is correct I updated `test` to be compiled via CMake and fix CI to work with the new paths.
- examples still work with conan, in this way in CI we make sure that conan still works and as a demonstration on how to use it.
- overall we have conan and cmake working side-by-side without any issue.

The only real design choice that I made here for the CMake solution is that `lz4` and `zstd` are imported and built only if tests are enabled. The reason is that I think it should be left to the user the decision whether to use and how to import them. And the test CMakeLists.txt shows how that could be done.
Please take a look and let me know what you think. Happy to provide extra feedback and explanation

indirectlylit:
Thank you @filippobrizzi - iterating on Github CI code can be frustrating and I appreciate the effort!

I've created internal tracking issue FG-6815 to make sure this stays on our radar.

defunctzombie:
@filippobrizzi Thanks for proposing this change. While we have been silent here we have had some internal discussions around this. Unfortunately no internal champion has risen to own this and we can't take on the maintenance overhead that this would bring without one. As your own commit history shows, there's a lot of gotchas and churn adding cmake can uncover and we are not in a good position to take that on right now.

I am going to close this since we won't be taking this on. If you want to host an example repo with "how to build mcap with cmake" we would be happy to link to it or even host a guest blog about it.

filippobrizzi:
Thanks for the update, what I do not understand is how do people use MCAP? 
Not every project uses Conan, and the only alternative I see is using FetchContent and manually adding the files as dependencies. 
Given that MCAP is a library, I would have expected you would like to support easy integration of the project for maximum adoption. 
To build MCAP with CMake the changes need to be inside this repo, so I cannot move them to a different repo.

defunctzombie:
> what I do not understand is how do people use MCAP?

It depends on your build system of choice. Here's one example with CMake:
https://github.com/ros2/rosbag2/blob/rolling/mcap_vendor/CMakeLists.txt#L34

Where the build system will fetch the sources and setup the library for use within that project.

Other build systems can follow similar patterns where they can reference the sources from the repo at some specific version. Since it is a header-only library the setup is "easy" as you only need to make the include folder available.

PR #1066, opened 2024-02-27
Add strict-message-order flag to `mcap doctor`

defunctzombie:
### Public-Facing Changes
New `--strict-message-order` flag for mcap doctor

The `strict-message-order` flag enables stricter message log_time order requirement checks for the `doctor` command. When this flag is specified, doctor will surface an error when the log time of a message is earlier than the latest known log time of messages while reading the message data in file order.

This flag helps surface situations where a writer wrote out-of-order (log time) message data to a chunk.


go/cli/mcap/cmd/doctor.go
```diff
@@ -182,13 +183,23 @@ func (doctor *mcapDoctor) examineChunk(chunk *mcap.Chunk) {
...
+				doctor.error("Message.log_time %d on %s is less than the oldest log time %d",
+					message.LogTime, channel.Topic, doctor.maxLogTime)
+			}
+
 			if message.LogTime < minLogTime {
 				minLogTime = message.LogTime
 			}
 
 			if message.LogTime > maxLogTime {
 				maxLogTime = message.LogTime
```
james-rms, line 196: do we still need this local variable `maxLogTime`?

defunctzombie: I think maybe not? Its not clear to me why these were not updating the `doctor` instance directly

achim-k (approved):

james-rms:
I would prefer for this check to always run and produce a warning by default, which can be upgraded to an error with `--strict-message-order` or some kind of `-Werror` equivalent.

defunctzombie:
@james-rms I've updated the PR to print a warning when the order is violated and print an error when the `--strict-message-order` flag is set.

go/cli/mcap/cmd/doctor.go
```diff
@@ -182,13 +183,29 @@ func (doctor *mcapDoctor) examineChunk(chunk *mcap.Chunk) {
 				doctor.error("Got a Message record for channel: %d before a channel info.", message.ChannelID)
 			}
 
+			if message.LogTime < doctor.maxLogTime {
+				errStr := fmt.Sprintf("Message.log_time %d on %s is less than the oldest log time %d",
```
james-rms, line 187:
```suggestion
				errStr := fmt.Sprintf("Message.log_time %d on %s is less than the latest log time %d",
```

james-rms (approved):

PR #1069, opened 2024-02-29
CLI instructions: determine tag from git history

james-rms:
### Public-Facing Changes

- Sets the download link for the linux MCAP CLI in docs at build time from the latest CLI release tag.
- Runs website deploy when a new MCAP CLI release tag is pushed, so this gets updated immediately.


jtbandes (approved):
magical

PR #1071, opened 2024-03-05
go: make time range options backwards-compatible

james-rms:
### Public-Facing Changes

Fixed: updated changes from #1068 to be backwards-compatible.

### Description

#1068 changed the type of some public API interfaces, which is a breaking change. To avoid needing to publish a v2 library, we deprecate those APIs and introduce new ones that take the new type.

### DIFF
--- a/go/cli/mcap/cmd/cat.go
+++ b/go/cli/mcap/cmd/cat.go
@@ -168,10 +168,10 @@ func getReadOpts(useIndex bool) []mcap.ReadOpt {
 	topics := strings.FieldsFunc(catTopics, func(c rune) bool { return c == ',' })
 	opts := []mcap.ReadOpt{mcap.UsingIndex(useIndex), mcap.WithTopics(topics)}
 	if catStart != 0 {
-		opts = append(opts, mcap.After(catStart*1e9))
+		opts = append(opts, mcap.AfterNanos(catStart*1e9))
 	}
 	if catEnd != math.MaxInt64 {
-		opts = append(opts, mcap.Before(catEnd*1e9))
+		opts = append(opts, mcap.BeforeNanos(catEnd*1e9))
 	}
 	return opts
 }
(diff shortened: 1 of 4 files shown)


james-rms:
cc. @wkalt

wkalt:
@james-rms no objections to the deprecation dance, but part of me thinks though,
* the expanded type definition could be argued as a bugfix
* keeping the deprecated stuff around forever is crufty
* not sure how much code is realistically getting broken here (maybe you have info on that)
* why not just roll a major?
* if we someday want a time.Time based API because we get sick of uint64 nanos, then it's going to be inconvenient to support three variants.

Anyway, up to you guys.

james-rms:
Yeah, i can justify my stance here.

> the expanded type definition could be argued as a bugfix

That'd be a reach. I can't really see the unexpected behavior with the original types that would be considered a bug. The only example I can think of is  if a user provided a negative Before value, this got rounded to 0 inside the library, then they would receive any message with timestamp 0 when they should have received no messages.
> keeping the deprecated stuff around forever is crufty

I agree.
> not sure how much code is realistically getting broken here (maybe you have info on that)

I don't have better info than you do - there are a few users of the golang library in the wild. Any code that uses either `Before`, `After` or `ReadOptions` would be broken by this change, because Go will not implicitly cast between int64 and uint64. 
> why not just roll a major?

The expectation in the go ecosystem is laid out [here](https://research.swtch.com/vgo-import):
```
If an old package and a new package have the same import path,
the new package must be backwards compatible with the old package.
```
The next major version needs to be available at a new module path. We could either continue to support both in `main`, and import one from the other, *or* move the whole `mcap` library tree to a different path in one commit and stop publishing bug fixes for v1. Either way, it's I feel this is too great a disruption to justify doing it for the sake of this change.

> if we someday want a time.Time based API because we get sick of uint64 nanos, then it's going to be inconvenient to support three variants.

This doesn't block us from doing that or make it harder - if we added `BeforeTime` and `AfterTime` I guess they would internally set the `*Nanos` fields on `ReadOptions`. That patch would not get larger or more complex because of the presence of some deprecated fields and functions.

defunctzombie (approved):

PR #1135, opened 2024-03-09
Go writer: support schemas/channels outside chunks

wkalt:
Adds a new option, ChannelsOutsideChunks, to the go writer.

Today in chunked MCAP files, schemas and channels are inside the chunks. I would like to be able to concatenate two nonoverlapping MCAP files without either decompressing chunks or having duplicate schemas/channels in the output. This option should make it possible to do that.

### DIFF
--- a/go/mcap/writer.go
+++ b/go/mcap/writer.go
@@ -50,6 +50,8 @@ type Writer struct {
 	opts *WriterOptions
 
 	closed bool
+
+	channelsOutsideChunks bool
 }
 
 // WriteHeader writes a header record to the output.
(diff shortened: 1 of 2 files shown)


wkalt:
I think this will require a new conformance test variant.. still studying how to do that.

go/mcap/writer_test.go
```diff
@@ -142,6 +143,80 @@ func TestOutputDeterminism(t *testing.T) {
...
+func TestChannelsOutsideChunks(t *testing.T) {
+	buf := &bytes.Buffer{}
+	w, err := NewWriter(buf, &WriterOptions{
+		Chunked:               true,
+		ChunkSize:             1024,
+		Compression:           CompressionNone,
+		IncludeCRC:            true,
+		ChannelsOutsideChunks: true,
+	})
+	require.NoError(t, err)
```
james-rms, line 155: what is this require business?

go/mcap/writer.go
```diff
@@ -827,6 +829,12 @@ type WriterOptions struct {
 	// Compressor is a custom compressor. If supplied it will take precedence
 	// over the built-in ones.
 	Compressor CustomCompressor
+
+	// ChannelsOutsideChunks causes channels and schemas to be written outside
```
james-rms, line 833: how do you feel about `ChunkMessagesOnly`?

go/mcap/writer_test.go
wkalt: I learned about this recently because my linter started complaining about usage of assert in dp3. Probably when you guys upgrade you'll get warnings too. I don't mind changing this to assert though - will do that.

As far as I understand it this causes the test to bail right away instead of plowing through, but I haven't actually verified the behavior, just that the linter was happy.

PR #1138, opened 2024-03-14
Support 'none' compression option in merge command

wkalt:
The merge command documents "none" is an option, but we were not converting that to the appropriate compression string. This led to an "unsupported compression" error when none was supplied. With this commit both "none" and the empty string are accepted.

### DIFF
--- a/go/cli/mcap/cmd/merge.go
+++ b/go/cli/mcap/cmd/merge.go
@@ -79,6 +79,8 @@ const (
 	AutoCoalescing  = "auto"
 	ForceCoalescing = "force"
 	NoCoalescing    = "none"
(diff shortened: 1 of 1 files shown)


jtbandes (approved):

go/cli/mcap/cmd/merge.go
```diff
@@ -79,6 +79,8 @@ const (
 	AutoCoalescing  = "auto"
 	ForceCoalescing = "force"
 	NoCoalescing    = "none"
+
+	compressionNoneAlias = "none"
```
wkalt, line 83: to satisfy the linter

PR #1144, opened 2024-03-20
go: use v1.22 in go.work, update golangci-lint, enable `testifylint`

james-rms:
### Public-Facing Changes

None.

### Description
- Updates go to `v1.22` in go.work. This doesn't affect the required go version for our packages. However, at least v1.21 is require for this new golangci-lint version. 
- Upgrades golangci-lint to 1.56.2 - this is newer, and matches the version we use for internal projects.
- enable `testifylint` - most significantly this forces all error assertions to use "require" instead of "assert". "require" halts the test at the point of failure, rather than assert which allows the test to keep going. IMO this is a significantly better experience, because usually after an error check fails, every other check will also fail and you'll also get a bunch of nil-dereference tracebacks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,7 +374,7 @@ jobs:
         with:
           go-version-file: go/go.work
       - name: install golangci-lint
-        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.54.2
+        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.56.2
       - run: make lint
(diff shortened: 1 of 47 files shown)


bryfox (commented):
The `require.NoError` change seems good to me.

I see we still have `assert.Nil(t, err)`-like statements in the conformance tests. Was this purposeful? Why doesn't the linter complain about those?

james-rms:
I think that's a coverage miss - i'll update the makefiles to lint those as well.

bryfox (approved):

PR #1145, opened 2024-03-22
go: add ordered lexer as a reader implementation

james-rms:

### DIFF
--- a/go/mcap/ordered_lexer.go
+++ b/go/mcap/ordered_lexer.go
@@ -1,0 +1,382 @@
+package mcap
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"slices"
+	"sort"
+
+	"github.com/klauspost/compress/zstd"
+	"github.com/pierrec/lz4/v4"
+)
+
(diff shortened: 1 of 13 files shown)


go/mcap/ordered_lexer.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+	}
+	return maxChunkRecordSize
+}
+
+// NewOrderedLexer constructs a reader. It inspects the summary section of the MCAP, checking for
+// overlapping chunks. For any range of chunks that overlap in time range, they are tagged as
+// needing to be merged together in order to be sorted. It then determines the size of chunk buffer
+// required, and returns an error if this is greater than `maxBufferSize`.
+func NewOrderedLexer(
```
wkalt, line 99: I kinda wonder if this is too smart to call a lexer? The current structuring of the lib is that the lexer is the dumb low-level in-order iterator, and the reader is the index-aware smart iterator. It kinda seems like having a lexer construct a reader subverts that.

james-rms: Yeah, i really don't have a good name for this thing. I don't really intend to merge this - the point was to demonstrate that it passes conformance tests, then port the good parts into the existing seeking reader.

Issue #1067, opened 2024-02-28
`mcap cat` start/end filtering is in seconds, but output is in nanoseconds


jhurliman:
**Description**

```
$ mcap cat input.mcap
1709146829659264519 /topic [schema] [10 12 8 205 133 254 174 6 16 128]...
1709146829760054571 /topic [schema] [10 12 8 205 133 254 174 6 16 128]...
1709146829858813612 /topic [schema] [10 12 8 205 133 254 174 6 16 128]...
```

I want to print only the first two messages, but I can't use `mcap cat --end-secs 1709146829760054571` because the output is in a different unit (nanoseconds) than the input arguments (seconds). Maybe additional `--start-nsecs` / `--end-nsecs` arguments?

- Version: 0.0.43
- Platform: Linux

PR #1151, opened 2024-04-04
CLI to allow filtering by nanoseconds

wimagguc:
### Changelog

The CLI now accepts `--start-nsecs` and `--end-secs` parameters to filter by nanosecond-precision timestamps.

### Description

Usage: 

```
mcap cat --start-nsecs 1616425750123456789 --end-nsecs 1616425750777888999 my.mcap
```

...or for the filter:

```
mcap filter --start-nsecs 1616425750123456789 --end-nsecs 1616425750777888999 my.mcap -o out.mcap
```

In response to this issue: https://github.com/foxglove/mcap/issues/1067

> ```
> $ mcap cat input.mcap
> 1709146829659264519 /topic [schema] [10 12 8 205 133 254 174 6 16 128]...
> 1709146829760054571 /topic [schema] [10 12 8 205 133 254 174 6 16 128]...
> ```
> I want to print only the first two messages, but I can't use mcap cat --end-secs 1709146829760054571 because the output is in a different unit (nanoseconds) than the input arguments (seconds). Maybe additional --start-nsecs / --end-nsecs arguments?

### DIFF
--- a/go/cli/mcap/cmd/cat.go
+++ b/go/cli/mcap/cmd/cat.go
@@ -27,8 +27,10 @@ import (
 
 var (
 	catTopics     string
-	catStart      uint64
-	catEnd        uint64
+	catStartSec   uint64
+	catEndSec     uint64
+	catStartNano  uint64
+	catEndNano    uint64
 	catFormatJSON bool
 )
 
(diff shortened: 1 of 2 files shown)


go/cli/mcap/cmd/cat.go
```diff
@@ -304,6 +318,13 @@ var catCmd = &cobra.Command{
 		readingStdin := (stat.Mode()&os.ModeCharDevice == 0 && len(args) == 0)
 		// stdin is a special case, since we can't seek
 
+		if catStartSec > 0 && catStartNano > 0 {
```
bryfox, line 321: nit: is there any mechanism to see which flags were actually passed? If `--start-nsec 0` is valid as "since epoch", then we might also want to prevent `--start-sec 123 --start-nsec 0` since I might otherwise expect the last flag to be used?

bryfox (commented):
I think you should seek others' input on the public-facing change (new flag names), but this seems good to me.

james-rms (approved):

go/cli/mcap/cmd/cat.go
wimagguc: It turns out that we can do even better: use Cobra's built-in `MarkFlagsMutuallyExclusive` for this very job. Thanks for the hint!

wimagguc:
> I think you should seek others' input on the public-facing change (new flag names), but this seems good to me.

Re standup this is good to go; @wkalt mentioned that it could be nice to introduce `--start --end` params that accept _any_ kind of timestamp — but that obviously would have been a _much_ bigger lift.

PR #1157, opened 2024-04-10
address SVG logo issue

indirectlylit:
### Changelog

none

### Description

As described in https://github.com/facebook/docusaurus/issues/8297, the default webpack config for docusaurus has svgo configured to replace `id` attributes with simple incrementing characters like `"a"`, `"b"`, etc. This leads to ID conflicts between logos and incorrect clipping behavior.

This PR modifies the underlying webpack config to disable changes to ID attributes.

Note that on mac, I was only able reproduce the visual clipping issue in Safari; FF and Chrome seemed to ignore the incorrectly duplicated IDs .

image

</td><td>

image

</td></tr></table>

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -9,58 +9,7 @@ const util = require("util");
 const webpack = require("webpack");
 const execAsync = util.promisify(require("child_process").exec);
 
-/**
- * Modify the svgo configuration (in place) to prevent it from minifying IDs in SVGs.
- * This is necessary because it doesn't account for the global ID namespace, and causes
- * ID collisions between the SVGs loaded into the same page.
- *
- * Refs:
- * - https://github.com/facebook/docusaurus/issues/8297
- * - https://github.com/svg/svgo/issues/1714
- * - https://linear.app/foxglove/issue/FG-7251/logos-are-cut-off-on-mcapdev
- *
- * @param {webpack.Configuration} config
- */
-function modifySvgoConfig(config) {
-  const NEW_SVGO_CONFIG = {
-    plugins: [
-      {
-        name: "preset-default",
-        params: {
-          overrides: {
-            removeTitle: false,
-            removeViewBox: false,
-            cleanupIDs: false, // do not change IDs
-          },
-        },
-      },
-    ],
-  };
-  // find the svgo config rule and replace it
-  if (config.module?.rules instanceof Array) {
-    for (const rule of config.module.rules) {
-      if (typeof rule === "object" && rule.test?.toString() === "/\\.svg$/i") {
-        if (rule.oneOf instanceof Array) {
-          for (const nestedRule of rule.oneOf) {
-            if (nestedRule.use instanceof Array) {
-              for (const loader of nestedRule.use) {
-                if (
-                  typeof loader === "object" &&
-                  /* cspell:disable */
-                  loader.loader === require.resolve("@svgr/webpack")
-                ) {
-                  if (typeof loader.options === "object") {
-                    loader.options.svgoConfig = NEW_SVGO_CONFIG;
-                  }
-                }
-              }
-            }
-          }
-        }
-      }
-    }
-  }
-}
+const modifySvgoConfigInPlace = require("./modifySvgoConfigInPlace");
 
 /** @type {import('@docusaurus/types').Config} */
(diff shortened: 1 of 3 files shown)


.github/workflows/website.yml
```diff
@@ -76,7 +76,8 @@ jobs:
           retention-days: 1
 
   docs-swift:
-    runs-on: ubuntu-latest
+    # ref: https://github.com/swift-actions/setup-swift/issues/506
+    runs-on: ubuntu-20.04
```
indirectlylit, line 80: The `docs-swift` check was failing with the error below until I made this change. It's not clear to me why this wouldn't have been an issue before, or if this change is correct.

```
Run swift-actions/setup-swift@v2
/usr/bin/gpg --import /home/runner/work/_temp/1feadda7-402d-4250-a671-d56e59bbe291
gpg: directory '/home/runner/.gnupg' created
gpg: keybox '/home/runner/.gnupg/pubring.kbx' created
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
Error: Unexpected error, unable to continue. Please report at https://github.com/swift-actions/setup-swift/issues
The process '/usr/bin/gpg' failed with exit code 2
Stacktrace:
Error: The process '/usr/bin/gpg' failed with exit code 2
    at ExecState._setResult (/home/runner/work/_actions/swift-actions/setup-swift/v2/dist/index.js:3190:25)
    at ExecState.CheckComplete (/home/runner/work/_actions/swift-actions/setup-swift/v2/dist/index.js:3173:18)
    at ChildProcess.<anonymous> (/home/runner/work/_actions/swift-actions/setup-swift/v2/dist/index.js:3067:27)
    at ChildProcess.emit (node:events:514:28)
    at maybeClose (node:internal/child_process:1105:16)
    at ChildProcess._handle.onexit (node:internal/child_process:305:5)
```

jtbandes: The job appears to be flaky (https://github.com/swift-actions/setup-swift/issues/591), I think you can remove this change.

website/docusaurus.config.js
```diff
@@ -9,6 +9,57 @@ const util = require("util");
...
+  };
+  // find the svgo config rule and replace it
+  if (config.module?.rules instanceof Array) {
+    for (const rule of config.module.rules) {
+      if (typeof rule === "object" && rule.test?.toString() === "/\\.svg$/i") {
+        if (rule.oneOf instanceof Array) {
+          for (const nestedRule of rule.oneOf) {
+            if (nestedRule.use instanceof Array) {
+              for (const loader of nestedRule.use) {
+                if (
```
jtbandes, line 45: This level of nesting is really intense... at minimum it would be nice to flip these conditions around and use early returns.

Alternatively this version looks a bit cleaner:
https://github.com/facebook/docusaurus/issues/8297#issuecomment-1323343137

I imagine you could use a `try/catch` around some of the modifications to reduce the number of things that need to be checked up front and just silently continue in the event of an error.

website/docusaurus.config.js
```diff
@@ -9,6 +9,57 @@ const util = require("util");
 const webpack = require("webpack");
 const execAsync = util.promisify(require("child_process").exec);
 
+/**
+ * Modify the svgo configuration (in place) to prevent it from minifying IDs in SVGs
```
jtbandes, line 13: This comment is missing a tiny bit more context on _why_ we don't want to minify IDs in SVGs

jtbandes (approved):
Wow...crazy that this does appear to be the best solution at the moment 😵‍💫

website/docusaurus.config.js
indirectlylit: The conditionals are more for the linter than actually being used because the webpack config is fairly static. This was really instead of using a lot of `ts-ignore` comments. (I tried something like the linked version and it required a lot of them.)

I'll play around with it a bit more, but I didn't want to spend too much time fighting the linter over webpack config type definitions :)

indirectlylit:
@jtbandes cleaned up a bit to isolate ugliness

.github/workflows/website.yml
indirectlylit: This appears to fail consistently once removing the change

jtbandes (approved):

PR #1158, opened 2024-04-12
fix python example raw reader topic to match writer

ChristopherAyling:
### Changelog

Fixed python raw example topics to match.

### Description

In the following example code: https://github.com/foxglove/mcap/tree/main/python/examples/raw

the topic specified in the writer (`"sample_topic"`) did not match the topic specified in the reader (`"/diagnostics"`). This led to the reader not being able to read the message that was written by the writer.

I have manually tested that this example now works.

The reader printed out nothing.

</td><td>

The reader prints out the message written by the writer.

</td></tr></table>


indirectlylit (approved):
thank you!

PR #1159, opened 2024-04-18
[go/cli] Fix uint64 bug in `mcap cat`

bradsquicciarini-coco:
### Changelog

`mcap cat` is not reading `uint64` ros1 msgs  properly

### Description

Super minor, but looks like the `JSONTranscoder` for `ros1msg` is only reading 4 bytes for the `uint64` type.

### DIFF
--- a/go/cli/mcap/utils/ros/json_transcoder.go
+++ b/go/cli/mcap/utils/ros/json_transcoder.go
@@ -277,7 +277,7 @@ func (t *JSONTranscoder) int64(w io.Writer, r io.Reader) error {
 }
 
 func (t *JSONTranscoder) uint64(w io.Writer, r io.Reader) error {
-	_, err := io.ReadFull(r, t.buf[:4])
+	_, err := io.ReadFull(r, t.buf[:8])
 	if err != nil {
 		return err
 	}
--- a/go/cli/mcap/utils/ros/json_transcoder.go
+++ b/go/cli/mcap/utils/ros/json_transcoder.go
@@ -282,7 +282,7 @@ func (t *JSONTranscoder) uint64(w io.Writer, r io.Reader) error {
 		return err
 	}
 	x := binary.LittleEndian.Uint64(t.buf[:8])
-	t.formattedNumber = strconv.AppendInt(t.formattedNumber, int64(x), 10)
+	t.formattedNumber = strconv.AppendUint(t.formattedNumber, x, 10)
 	_, err = w.Write(t.formattedNumber)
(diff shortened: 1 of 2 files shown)


jtbandes:
Thanks for the fix! Could you add some tests for this? Looks like there are existing tests in https://github.com/foxglove/mcap/blob/main/go/cli/mcap/utils/ros/json_transcoder_test.go

bradsquicciarini-coco:
Kinda of interesting there already was a test, but...

Running all the existing tests worked
```bash
$ go test ./utils/ros -count=1                                           
ok      github.com/foxglove/mcap/go/cli/mcap/utils/ros  0.115s
```

Running just the uint64 test didn't
```bash
$ go test ./utils/ros -run TestSingleRecordConversion/uint64  -count=1
--- FAIL: TestSingleRecordConversion (0.00s)
    --- FAIL: TestSingleRecordConversion/uint64 (0.00s)
        json_transcoder_test.go:532: 
                Error Trace:    /Users/bradsquicciarini/coco/mcap/go/cli/mcap/utils/ros/json_transcoder_test.go:532
                Error:          Not equal: 
                                expected: "{\"foo\":506381209866536711}"
                                actual  : "{\"foo\":117901063}"
                                
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -{"foo":506381209866536711}
                                +{"foo":117901063}
                Test:           TestSingleRecordConversion/uint64
FAIL
FAIL    github.com/foxglove/mcap/go/cli/mcap/utils/ros  0.117s
FAIL
```

Looks like the test was reading some well placed garbage from prior tests. So just zeroed out the buffer between test cases.

jtbandes:
lol…amazing 🙌

bryfox (commented):
Thanks for the fix!

It looks like there's also an incorrect conversion a few lines below; would you like to deal with that in this PR (and maybe use a larger value in the test to catch it)?

https://github.com/foxglove/mcap/blob/main/go/cli/mcap/utils/ros/json_transcoder.go#L285

https://pkg.go.dev/strconv#AppendUint

bryfox (approved):
Thank you 👍

wkalt:
@bryfox @jtbandes I ported this to the copy of this code at https://github.com/foxglove/go-rosbag/pull/5

IIRC the foxglove CLI tool consumes that version so you will probably want to update its deps once that goes in. The MCAP CLI tool should probably also be updated to refer to that code instead.

PR #1160, opened 2024-04-22
CLI: scan in log time order when using index

wkalt:
This changes the cat command to do a log-time order scan when it is able to use the index (e.g seekable reader).

This was the intent of the existing code (there is no other reason for it to try to use the index), but it was not supplying both necessary parameters.


wkalt:
cc @james-rms

james-rms (approved):

PR #1161, opened 2024-04-22
Remove the ros1msg JSON transcoder and reference go-rosbag

wkalt:
This removes the JSON transcoder from the CLI ros1msg utils, and references instead the copy of the same code in foxglove/go-rosbag.

That transcoder was previously split out from the CLI in order to better package it more explicitly for standalone consumption.


wkalt:
ok, this should now be good to go.

bryfox (approved):
👍 

Resolves FG-7417

PR #1162, opened 2024-04-23
mcap-protobuf-support: restore topological sort

james-rms:
### Changelog

- Fixed: `mcap-protobuf-support` previously failed to decode Protobuf messages where the message schema file descriptor set was not written in topological dependency order.

### Description

#1148 changed the logic around loading file descriptor sets to decode protobuf messages. The new logic is not able to handle protobuf schemas where the file descriptors were not in topological dependency order. Tests did not catch this because the `mcap_protobuf` python writer writes file descriptors in topological order. However, the reader needs to be able to handle different orders.

</td><td>

</td></tr></table>


james-rms:
Currently stuck on https://github.com/pypi/warehouse/issues/15824

jtbandes (approved):
Awesome!

jtbandes:
@james-rms looks like we forgot to tag this version so it got published to PyPI – I just tagged it now.

PR #1164, opened 2024-04-25
go: reduce memory footprint of writer

james-rms:
### Changelog

- Changed: reduced memory footprint of writer by removing an uneccessary chunk buffer.

### Description

We are unneccessarily copying the entire compressed chunk into an internal buffer before writing it out - this increases the memory footprint of an mcap writer by the size of a compressed chunk. This isn't too important usually, but when you're running an application that writes 200 MCAP files at once, those 4MB buffers can add up.

This PR removes that chunk buffer and instead uses the `writer.msg` scratch buffer to marshal only the "header" of the chunk. This is all fields of the chunk record besides the compressed data.

This PR also adds a benchmark to demonstrate the memory saving. This benchmark is similar to the existing WriterAllocs benchmark, but:
- we allocate 100 writers and write the nth message to the (n % 100)th writer
- instead of writing the output MCAP to a buffer, the 100 writers all write to io.Discard. This is because we don't care about the output result, and we also want to reduce the effect of the allocated output on the result as much as possible.

Benchmark results:
```
Before:

(python-sNIFi2pF) j@192-168-1-105 mcap %  go test ./... -bench=BenchmarkManyWriterAllocs -benchtime=10s -benchmem
goos: darwin
goarch: arm64
pkg: github.com/foxglove/mcap/go/mcap
BenchmarkManyWriterAllocs/big_chunks_many_messages-8                  22         510907131 ns/op           4038205 messages/sec 515778960 B/op     48796 allocs/op
BenchmarkManyWriterAllocs/small_chunks_many_messages-8                25         458119757 ns/op           4355813 messages/sec 11032190 B/op      79796 allocs/op
BenchmarkManyWriterAllocs/many_channels-8                              2        5889904000 ns/op            345263 messages/sec 4025878024 B/op 22587677 allocs/op
PASS

After:

(python-sNIFi2pF) j@192-168-1-105 mcap %  go test ./... -bench=BenchmarkManyWriterAllocs -benchtime=10s -benchmem
goos: darwin
goarch: arm64
pkg: github.com/foxglove/mcap/go/mcap
BenchmarkManyWriterAllocs/big_chunks_many_messages-8                  22         513380782 ns/op           3988771 messages/sec 341286355 B/op     48698 allocs/op
BenchmarkManyWriterAllocs/small_chunks_many_messages-8                25         464017923 ns/op           4417701 messages/sec  8853856 B/op      79700 allocs/op
BenchmarkManyWriterAllocs/many_channels-8                              2        5858519938 ns/op            347740 messages/sec 2881396708 B/op 22587001 allocs/op
PASS
ok      github.com/foxglove/mcap/go/mcap        43.179s
```

As expected, the write speed is roughly the same, but far fewer bytes are allocated in all benchmark cases.

MCAP writer copies the full chunk record into a buffer before writing that buffer out to the writer.

</td><td>

MCAP writer copies the chunk header into a buffer, writes it out, then writes out the compressed chunk data.

</td></tr></table>

### DIFF
--- a/go/mcap/writer.go
+++ b/go/mcap/writer.go
@@ -433,7 +433,8 @@ func (w *Writer) flushActiveChunk() error {
 	crc := w.compressedWriter.CRC()
 	compressedlen := w.compressed.Len()
 	uncompressedlen := w.compressedWriter.Size()
-	msglen := 8 + 8 + 8 + 4 + 4 + len(w.opts.Compression) + 8 + compressedlen
+	topFieldsLen := 8 + 8 + 8 + 4 + 4 + len(w.opts.Compression) + 8
+	msglen := topFieldsLen + compressedlen
 	chunkStartOffset := w.w.Size()
 	var start, end uint64
 	if w.currentChunkMessageCount != 0 {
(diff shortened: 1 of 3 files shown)


james-rms:
cc. @wkalt

wkalt:
@james-rms the measurements you show are nearly identical. I don't doubt the effect but those examples don't really show it.. does this make a big difference in the 200+ files case you mention?

go/mcap/writer.go
```diff
@@ -445,25 +444,45 @@ func (w *Writer) flushActiveChunk() error {
 	// when writing a chunk, we don't go through writerecord to avoid needing to
 	// materialize the compressed data again. Instead, write the leading bytes
 	// then copy from the compressed data buffer.
-	recordlen := 1 + 8 + msglen
-	if len(w.chunk) < recordlen {
-		w.chunk = make([]byte, recordlen*2)
-	}
-	offset, err := putByte(w.chunk, byte(OpChunk))
+	_, err = w.w.Write([]byte{byte(OpChunk)})
```
wkalt, line 447: I would expect this could change the dynamics depending on whether the user is passing a buffered or unbuffered writer. The previous implementation effectively buffered writes no matter what kind of writer got passed in, while this one seems like it will do more syscalls if the writer passed is unbuffered. I don't think that explains the lack of performance changes though since the benchmark is writing to ram.

go/mcap/writer.go
```diff
@@ -445,25 +444,45 @@ func (w *Writer) flushActiveChunk() error {
...
 	}
-
-	offset += putUint64(w.chunk[offset:], uint64(msglen))
-	offset += putUint64(w.chunk[offset:], start)
-	offset += putUint64(w.chunk[offset:], end)
-	offset += putUint64(w.chunk[offset:], uint64(uncompressedlen))
-	offset += putUint32(w.chunk[offset:], crc)
-	offset += putPrefixedString(w.chunk[offset:], string(w.opts.Compression))
-	offset += putUint64(w.chunk[offset:], uint64(w.compressed.Len()))
-	offset += copy(w.chunk[offset:recordlen], w.compressed.Bytes())
```
wkalt, line 464: if you keep the existing stuff up to line 463, then just do the write of compressed.Bytes() out to the original writer instead of the chunk buffer (with the chunk "header" flushed out first) - do you get all the same benefits with fewer line changes + two calls to write?

wkalt (approved):
@james-rms thanks for the updates, looks good to me.

go/mcap/writer.go
```diff
@@ -434,7 +433,8 @@ func (w *Writer) flushActiveChunk() error {
 	crc := w.compressedWriter.CRC()
 	compressedlen := w.compressed.Len()
 	uncompressedlen := w.compressedWriter.Size()
-	msglen := 8 + 8 + 8 + 4 + 4 + len(w.opts.Compression) + 8 + compressedlen
+	topFieldsLen := 8 + 8 + 8 + 4 + 4 + len(w.opts.Compression) + 8
```
defunctzombie, line 436: What is "topFields"? That's not an mcap spec term I am familiar with. Might a comment be helpful here - especially as it relates to all these magic numbers being added together.

go/mcap/writer.go
```diff
@@ -445,24 +445,25 @@ func (w *Writer) flushActiveChunk() error {
...
-	offset += copy(w.chunk[offset:recordlen], w.compressed.Bytes())
-	_, err = w.w.Write(w.chunk[:offset])
+	offset += putUint64(w.msg[offset:], uint64(msglen))
+	offset += putUint64(w.msg[offset:], start)
+	offset += putUint64(w.msg[offset:], end)
+	offset += putUint64(w.msg[offset:], uint64(uncompressedlen))
+	offset += putUint32(w.msg[offset:], crc)
+	offset += putPrefixedString(w.msg[offset:], string(w.opts.Compression))
+	offset += putUint64(w.msg[offset:], uint64(w.compressed.Len()))
+	_, err = w.w.Write(w.msg[:offset])
```
defunctzombie, line 462: nit: not for this PR but I personally find `w.w` rather questionable from a maintenance standpoint. What is `w` ?

defunctzombie (approved):
Does this change any semantics about the concurrency of the writer or its use in a concurrency environment?

james-rms:
No change to concurrency behavior -the writer remains non-thread-safe and should be locked externally if people want to use it across multiple goroutines.

PR #1165, opened 2024-05-01
go: add file offset to lexer API

james-rms:
### Changelog

- go: added GetLastTokenOffset() API to Lexer. This returns the location of the last yielded record in the file, which the caller can use to build up their own index of records in an MCAP.
### Description

An application reading an MCAP might find some records it wants to use later, but cannot afford to hold all of them in memory until that time. If the MCAP is indexed, the reader might use the index to find these records later, but reading the index requires seeking and reading different parts of the file, which is not free. Also, some MCAPs are not indexed.  Ideally the reader could store of the offsets of those records as it reads past them, then seek directly to them later. However, there's not a great way to do that now.

This PR adds a new method `Lexer.GetLastTokenOffset()`, which returns the offset of the last record yielded by the Lexer. This allows them to seek directly to it later. The returned value is a pair of int64 offsets (file and chunk). 

`int64` was chosen here because then we can use a negative sentinel value `RecordNotInChunk` to mark records that are not in chunks. It also fits well in the Go ecosystem, where:
- The `Seek()` API takes and returns an `int64` position
- `Read()` returns an `int` indicating bytes read, which is equivalent to int64 on 64-bit systems (int32 on 32-bit)
- `len()` returns an `int`

All of this to say, if an offset needs a larger int representation, the caller would not be able to seek to it anyway.

### DIFF
--- a/go/mcap/lexer.go
+++ b/go/mcap/lexer.go
@@ -14,12 +14,12 @@ import (
 
 type countingReader struct {
 	r   io.Reader
-	pos uint64
+	pos int64
 }
 
 func (cr *countingReader) Read(p []byte) (int, error) {
 	n, err := cr.r.Read(p)
-	cr.pos += uint64(n)
+	cr.pos += int64(n)
 	return n, err
 }
 
@@ -211,6 +211,7 @@ func (l *Lexer) Next(p []byte) (TokenType, []byte, error) {
 	for {
 		if !l.inChunk {
 			l.lastOffset.FileOffset = l.reader.pos
+			l.lastOffset.ChunkOffset = RecordNotInChunk
 		} else {
 			l.lastOffset.ChunkOffset = l.reader.pos
 		}
@@ -220,7 +221,6 @@ func (l *Lexer) Next(p []byte) (TokenType, []byte, error) {
 			eof := errors.Is(err, io.EOF)
 			if l.inChunk && (eof || unexpectedEOF) {
 				l.inChunk = false
-				l.lastOffset.ChunkOffset = 0
 				l.reader = l.basereader
 				continue
 			}
--- a/go/mcap/lexer.go
+++ b/go/mcap/lexer.go
@@ -198,8 +198,8 @@ type Lexer struct {
 	decompressors            map[CompressionFormat]ResettableReader
 }
 
-// GetLastTokenOffset returns the TokenOffset of the last token returned from Next().
-func (l *Lexer) GetLastTokenOffset() RecordOffset {
+// GetLastRecordOffset returns the offset of the last record returned from Next().
+func (l *Lexer) GetLastRecordOffset() RecordOffset {
 	return l.lastOffset
 }
 
--- a/go/mcap/lexer.go
+++ b/go/mcap/lexer.go
@@ -12,6 +12,17 @@ import (
 	"github.com/pierrec/lz4/v4"
 )
 
+type countingReader struct {
+	r   io.Reader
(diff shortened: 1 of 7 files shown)


james-rms:
cc. @wkalt

wkalt:
Can the caller pass in a counting reader and get the same benefits today (by counting lengths) without a code change?

james-rms:
There are two issues with that approach:

1. records in chunks - if I want to seek to a message in a compressed chunk, a counting reader will tell me some indeterminate offset into the compressed data which isn't useful to start a decoder at.
2. attachments - If i have some records [attachment, message] in my MCAP, I can't get the offset of the message using a counting reader. One call to Next() will first yield the attachment into the attachment callback, then the message. If i check the count before Next(), i get the offset at the start of the attachment, and if I check the count after next() I get the offset after the message.

wkalt:
yeah. For compressed chunks, I guess you would need to use the lexer in "emit chunks" mode and then compute offsets from that. I suspect (from memory) that you might not be able to make the lexer use a counting reader for the actual handling of chunk data today without doing that. It probably constructs its own chunk reader.

For number 2 I'm not sure I follow the part about attachments, I guess you're saying you want the indexes of those as well? If that's the case I need to take another look at how this would get handled.

I was thinking you would call Count() on the reader after you call Next(), and subtract 1 + 8 + length of returned token. That seems like it would give you the message offsets (provided you have gone through the hassle of using emit chunks) and allow you to build the indexes.

I was mainly trying to understand if this is possible with the existing SDK or not. If it is possible but this is way more convenient, then this could still be good though.

wkalt:
well, I guess actually if you're handling the chunks manually you don't even need a counting reader. You can just keep count directly by summing the lengths + 9 on each iteration

wkalt:
@james-rms this almost makes me wonder if a better API would be some kind of "message callback" similar to what we have for attachments, metadata. That could be used to to track offsets and maintain an index, but potentially also for other stuff.

james-rms:
The other idea I had here was to implement `NextIncludingOffset(p []byte) (TokenType, []byte, RecordOffset, error)` - might be a slightly cleaner API, but to my eye it feels like a more unwieldy API to deal with.

wkalt:
Another possible API would be to add a config option to the writer:
```
OnChunk func(n int64)
```

this gets called with the chunk offset every time a new chunk gets opened up. For maintaining the index, you can use a scheme like this:

```
var messageIndexes [][]int64{}
var currentChunkOffset, withinChunkOffset int64
OnChunk := func(n int) {
  currentChunkOffset = n
  withinChunkOffset = 0
}

for {
  tokenType, token, err := lexer.Next(p)
  if err != nil {
    // handle
  }

  messageIndexes = append(messageIndexes,  []int64{currentChunkOffset, withinChunkOffset})
  withinChunkOffset += 1 + 8 + len(token)
}

```

I'm not sure if any snags exist with this approach but it feels nice to me because it is a similar API to what we already have for metadata/attachments, and also would be symmetric with a hypothetical future API that executes a callback on chunks when they are written, which could be used for maintaining an index outside the file. Thoughts?

defunctzombie:
@james-rms I am gonna let you and @wkalt talk thought this and come to some agreement on what to ship here since ya'll have much more context on the interface needs. Happy to stamp this once there is agreement - give me a ping.

james-rms:
I've thought about this some more, I don't need this right now - closing until maybe a more compelling use case comes up that might drive the discussion.

PR #1168, opened 2024-05-07
go: reader: make ordered iteration faster

james-rms:
### Changelog
- Added `Message.PopulateFrom([]byte) error` method to Message, which allows readers to avoid reallocating Message structs between messages.
- Added `MessageIterator.Next2(*Message) (*Schema, *Channel, *Message, error)` method to the MessageIterator interface. This method allows the caller to re-use memory between message reads, and to avoid allocating a new Message on the heap for every message read.
- optimized indexed reading to read much faster and use much less memory.

### Docs

- [x] Update docstrings
- [x] Update examples

### Description

This PR makes message iteration much faster. The included benchmark shows a pretty significant speedup:

#### Before

Note: the comparison benchmark based on `main` uses MessageIterator.Next() to read messages.

```
$ go test . -run=^$ -bench=BenchmarkReader -benchmem -benchtime=10x 
goos: darwin
goarch: arm64
pkg: github.com/foxglove/mcap/go/mcap
BenchmarkReader/inorder/no_index-8                    10         532682221 ns/op         238.83 MB/s       7826155 msg/s        533306349 B/op   8054975 allocs/op
BenchmarkReader/inorder/index_file_order-8            10        1886590288 ns/op          67.29 MB/s       2204813 msg/s        909884028 B/op  12114339 allocs/op
BenchmarkReader/inorder/index_time_order-8            10        2248067917 ns/op          54.39 MB/s       1782145 msg/s        909889379 B/op  12114382 allocs/op
BenchmarkReader/inorder/index_rev_order-8             10        2324738488 ns/op          48.90 MB/s       1602346 msg/s        910261216 B/op  12114355 allocs/op
BenchmarkReader/inorder/bare_lexer-8                  10         196806788 ns/op         660.42 MB/s      21640757 msg/s        17005039 B/op       4672 allocs/op
BenchmarkReader/minor/no_index-8                      10         509497992 ns/op         241.92 MB/s       7927082 msg/s        531637254 B/op   8054932 allocs/op
BenchmarkReader/minor/index_file_order-8              10        1837735883 ns/op          66.97 MB/s       2194637 msg/s        909846889 B/op  12114373 allocs/op
BenchmarkReader/minor/index_time_order-8              10        2250390946 ns/op          54.82 MB/s       1796497 msg/s        909844632 B/op  12114332 allocs/op
BenchmarkReader/minor/index_rev_order-8               10        2360883250 ns/op          53.23 MB/s       1744292 msg/s        910212621 B/op  12114308 allocs/op
BenchmarkReader/minor/bare_lexer-8                    10         195830417 ns/op         638.20 MB/s      20912477 msg/s        15341232 B/op       4655 allocs/op
BenchmarkReader/major/no_index-8                      10         510946658 ns/op         241.74 MB/s       7921189 msg/s        532934768 B/op   8054945 allocs/op
BenchmarkReader/major/index_file_order-8              10        1841807000 ns/op          66.35 MB/s       2174050 msg/s        909833931 B/op  12114348 allocs/op
BenchmarkReader/major/index_time_order-8              10        2247866758 ns/op          54.53 MB/s       1786987 msg/s        909836941 B/op  12114379 allocs/op
BenchmarkReader/major/index_rev_order-8               10        2328824133 ns/op          51.12 MB/s       1675101 msg/s        910215724 B/op  12114370 allocs/op
BenchmarkReader/major/bare_lexer-8                    10         198086167 ns/op         632.44 MB/s      20724011 msg/s        16635893 B/op       4661 allocs/op
PASS
ok      github.com/foxglove/mcap/go/mcap        248.060s
```

#### After

```
% go test . -run=^$ -bench=BenchmarkReader -benchmem -benchtime=10x 
goos: darwin
goarch: arm64
pkg: github.com/foxglove/mcap/go/mcap
BenchmarkReader/inorder/no_index-8                    10         209814421 ns/op         596.62 MB/s      19550071 msg/s        17491784 B/op       6310 allocs/op
BenchmarkReader/inorder/index_file_order-8            10         340508775 ns/op         360.08 MB/s      11799088 msg/s        10446040 B/op      16981 allocs/op
BenchmarkReader/inorder/index_time_order-8            10         341343088 ns/op         359.00 MB/s      11763672 msg/s        10443932 B/op      16955 allocs/op
BenchmarkReader/inorder/index_rev_order-8             10         348526088 ns/op         356.74 MB/s      11689775 msg/s         9996309 B/op      16964 allocs/op
BenchmarkReader/inorder/bare_lexer-8                  10         187405846 ns/op         664.57 MB/s      21776806 msg/s        17439823 B/op       4674 allocs/op
BenchmarkReader/minor/no_index-8                      10         211110267 ns/op         587.06 MB/s      19236916 msg/s        16522652 B/op       6284 allocs/op
BenchmarkReader/minor/index_file_order-8              10         356903517 ns/op         336.77 MB/s      11035283 msg/s        10419253 B/op      17006 allocs/op
BenchmarkReader/minor/index_time_order-8              10         552369746 ns/op         218.44 MB/s       7157955 msg/s        10444996 B/op      17744 allocs/op
BenchmarkReader/minor/index_rev_order-8               10         555191658 ns/op         220.27 MB/s       7217936 msg/s         9971279 B/op      17665 allocs/op
BenchmarkReader/minor/bare_lexer-8                    10         194812112 ns/op         554.57 MB/s      18172347 msg/s        16473271 B/op       4670 allocs/op
BenchmarkReader/major/no_index-8                      10         211406192 ns/op         579.88 MB/s      19001450 msg/s        17365727 B/op       6291 allocs/op
BenchmarkReader/major/index_file_order-8              10         354124750 ns/op         347.12 MB/s      11374355 msg/s        10418725 B/op      16979 allocs/op
BenchmarkReader/major/index_time_order-8              10         566783688 ns/op         215.38 MB/s       7057431 msg/s        16452847 B/op      17690 allocs/op
BenchmarkReader/major/index_rev_order-8               10         563155871 ns/op         218.15 MB/s       7148236 msg/s        15986112 B/op      17699 allocs/op
BenchmarkReader/major/bare_lexer-8                    10         195610721 ns/op         633.25 MB/s      20750327 msg/s        17316992 B/op       4672 allocs/op
PASS
ok      github.com/foxglove/mcap/go/mcap        68.716s
```

For the unindexed message iterator, all of the difference comes from:
- being able to re-use a Message struct between calls to Next()
- switching from storing channels and schemas in maps to slices.
- using and re-using an internal buffer for lexing MCAP records.

For the indexed message iterator, we do all of the same things plus:
- We maintain a pool of decompressed chunk buffers, which are re-used after all of the messages from a given chunk are read out.
- We no longer read message indexes from the file, choosing instead to read the chunk content and build our own message index in memory. This allows us to read files written without message indexes in order, and also reduces I/O, which in some cases is *probably faster* (slow network connections with large message index overheads)
- we no longer use a heap to maintain order, instead we maintain a sorted array of unread message indexes. Every time we encounter a new chunk:
  - if this chunk does not overlap with the last, clear the message index array and write the new chunk's messages in. If they are already in order, do not sort. This makes ordered iteration over an in-order MCAP as fast as unordered iteration.
  - if the new chunk's messages are not in order, sort the new chunk's messages.
  - if the new chunk overlaps with the last, append the new chunk's messages, and sort all unread messages.

#### New API justification
The issues with `Next(p []byte) (*Schema, *Channel, *Message, error)` that caused me to explore alternatives are:
- The buffer `p` is used to store the message data, but if it isn't big enough, a new buffer is allocated for every message. A common idiom is to return the newly allocated buffer so that the caller can take ownership of it and re-use it, but the current API doesn't allow that.
- A new Message struct is allocated on the heap for every iteration. Even if the message goes out of scope on the next loop, this still causes significant work for the garbage collector to do.

This new function signature re-uses the message passed in, if one is passed in. If `nil` is used, it creates a new Message.


go/mcap/range_index_heap.go
```diff
@@ -93,15 +92,83 @@ func (h *rangeIndexHeap) Less(i, j int) bool {
...
 	return h.lastErr
 }
 
-func (h *rangeIndexHeap) HeapPop() (*rangeIndex, error) {
-	result := heap.Pop(h).(rangeIndex)
+func (h *rangeIndexHeap) Pop() (rangeIndex, error) {
+	n := h.len() - 1
+	h.swap(0, n)
+	heapDown(h, 0, n)
+	old := h.indices
```
james-rms, line 105: simplify this

go/mcap/reader.go
```diff
@@ -66,6 +66,8 @@ type Reader struct {
 
 type MessageIterator interface {
 	Next([]byte) (*Schema, *Channel, *Message, error)
+	ReadNextInto(buf []byte, msg *Message) ([]byte, error)
```
james-rms, line 69: docstrings here

go/mcap/reader.go
```diff
@@ -66,6 +66,8 @@ type Reader struct {
...
+	SchemaAndChannelForID(channelID uint16) (*Schema, *Channel, error)
```
james-rms, line 70: explain why we need this - alternatively, consider an API where you populate the message and also return the schema and channel pointers in one call?

james-rms:
cc. @wkalt

wkalt:
Do you have any thought on simplifying this to,

```
ReadNext(msg *message.Message) error
```

and making the reuse of the message's data buffer implicit?

Secondly, can you quantify the cost of the schema/channel lookups relative to the data allocations (and heap iterations IIRC) and share how much the relative contribution of each is hurting us? My recollection is that the main costs are in allocations and in the heap, rather than in the lookups. The lookups can also be converted to slice lookups pretty easily, either with a dynamically sized slice or (worst case) the full set of 2^16 which would only take 512kb. I would guess the cost of the lookups if that were done should go to approximately zero.

Third, could you post comparisons of the iteration performance of the reader vs the lexer? IIRC the lexer is much faster, like maybe 5x on adrian's mcnasty.mcap file. I think the goal of the reader improvements should roughly be to match the performance of the lexer, so it would be useful to have that as a comparison.

Fourth, have you investigated whether it is possible to bypass the heap completely in the case of a fully-ordered chunk (which can be known from inspection of the index)? My hunch previously was that some significant savings may come from here.

Great work overall, will take a closer look/test it out shortly.

go/mcap/indexed_message_iterator.go
```diff
@@ -191,49 +211,56 @@ func (it *indexedMessageIterator) loadChunk(chunkIndex *ChunkIndex) error {
...
-			timestamp := messageIndex.Records[i].Timestamp
-			if timestamp >= it.start && timestamp < it.end {
-				if err := it.indexHeap.HeapPush(rangeIndex{
-					chunkIndex:        chunkIndex,
-					messageIndexEntry: &messageIndex.Records[i],
-					buf:               chunkData,
-				}); err != nil {
-					return err
+		recordContent := chunkSlot.buf[recordStart:recordEnd]
+		msg := Message{}
```
wkalt, line 240: isn't this still allocating on every call to next?

go/mcap/indexed_message_iterator.go
```diff
@@ -254,60 +281,99 @@ func readRecord(r io.Reader) (OpCode, []byte, error) {
...
+	}
+	if channel.SchemaID == 0 {
+		return nil, channel, nil
+	}
+	schema, ok := it.schemas[channel.SchemaID]
+	if !ok {
+		return nil, nil, fmt.Errorf("no associated schema with id %d for channel %d", channel.SchemaID, channelID)
+	}
+	return schema, channel, nil
+}
```
wkalt, line 373: isn't this just pushing the same lookups we were previously doing, onto the user? i.e to actually use this code they need to call this on each iteration, thus obviating any benefits of removing it from the signature?

wkalt:
@james-rms I'm not sure if it's useful at all but this is the code I was working on when I was digging into this previously: https://github.com/wkalt/mcap/blob/task/golang-reader-perf-improvements/go/mcap/writer_test.go#L703

The test linked there will generate a sample file. This benchmark is the one I guess I was then using for comparisons:
https://github.com/wkalt/mcap/blob/task/golang-reader-perf-improvements/go/mcap/reader_test.go#L749

I don't remember unfortunately where I left things off, except that I broke the code and then didn't get it working again. It looks to me like the ideas that patch is implementing are, a) switch to a slice-based lookup for the channels/schemas, and b) allow the user to pre-allocate structures (I doubt preallocating schema/channel actually matters at all, since those are comparatively infrequent).

My overall reaction to this PR is that I suspect we can probably maintain something closer to a traditional "iterator" interface (modulo whatever is needed for message preallocation -- i.e still returning our schema/channel pointers) while still getting all the performance improvements we want, so that in the 99% case where files are ordered, the lexer and the reader are doing identical work and are equally fast. I'm not wild about the interface changes required in this specific patch. I'm not really a fan of SchemaAndChannelForID - I think it just pushes the work our code did previously onto the user and probably results in the same amount of aggregate lookups being done in actual use (possibly with the user maintaining their own, duplicate mappings). If feasible I would definitely prefer if we could keep returning those pointers and instead make this efficient by changing the backing structures to slices.

go/mcap/indexed_message_iterator.go
wkalt: ok, not every call

wkalt:
I guess my concrete thought is, can we make this the Next signature:

```
Next(*Message) (*Schema, *Channel, *Message, error)
```

and get performance identical to the lexer, with that being the only required new interface. I suspect the answer is yes -- and if so, that seems like a better outcome.

james-rms:
Regarding the signature - i've pushed an update which makes the API `ReadInto(*Message) (*Schema, *Channel, error)`. i can't see a point returning the message pointer if it's always going to be the same as the passed-in pointer.

Regarding always looking up the schema and channel - this makes a noticeable difference to the benchmark. I tested 2d2d767e4 against a patch that skips lookup and always returns `nil` for schemas and channels:

#### Without lookup
```
go test ./... -run=^$ -bench=BenchmarkReader -benchmem -benchtime=20s 
BenchmarkReader/no_index-8                   100         209160724 ns/op         149.31 MB/s       4892764 msg/s        18126126 B/op       2883 allocs/op
BenchmarkReader/with_index-8                  62         381960212 ns/op          74.79 MB/s       2450602 msg/s         7425817 B/op       5960 allocs/op
PASS
```

#### With lookup
```
go test ./... -run=^$ -bench=BenchmarkReader -benchmem -benchtime=20s 
BenchmarkReader/no_index-8                   100         230150386 ns/op         132.22 MB/s       4332659 msg/s        16929751 B/op       2871 allocs/op
BenchmarkReader/with_index-8                  55         429351720 ns/op          71.90 MB/s       2356059 msg/s         7425896 B/op       5960 allocs/op
PASS
```

My idea when separating the lookup to its own function is that for users who need to go fast, they rarely want the schema and channel information more than once. 
- If you're splitting an MCAP up,  for the first message you figure out which channel IDs need to go into which stream, then you build up a mapping and key off the channel ID after then.
- If you're trying to decode messages, you instantiate your deserializers with schema information once per channel, you store them in a mapping of channel ID -> deserializer, and you never look at the schema information again.

go/mcap/indexed_message_iterator.go
james-rms: it's on the stack, and it's once per chunk.

james-rms:
Also re: testing against the lexer: I have a rough estimate here:
```
BenchmarkReader/lexer-8                    10         175206304 ns/op         173.64 MB/s       5689735 msg/s        17305244 B/op       1260 allocs/op
```

wkalt:
@james-rms I think there is definitely value in returning the message pointer because some users may prefer to allocate on each call, i.e by passing in a nil as the input pointer.

This would be useful if they want to pass the messages through some kind of processing pipeline and aren't too concerned about maximal perf. I'm assuming that in a future version we'll want to deprecate & remove the old interface and push users toward this one.

wkalt:
Is that benchmark you posted for the lexer comparable to the other two posted earlier? That seems like a great basis for comparison

wkalt:
Here's my results fwiw:
```
[~/repos/mcap/go/mcap] (jrms/faster-message-iterator) $ go test ./... -run=^$ -bench=BenchmarkReader -benchmem -benchtime=20s 
goos: linux
goarch: amd64
pkg: github.com/foxglove/mcap/go/mcap
cpu: AMD Ryzen 9 7950X 16-Core Processor            
BenchmarkReader/no_index-32                  132         182788097 ns/op         163.32 MB/s       5351822 msg/s        17561395 B/op       2879 allocs/op
BenchmarkReader/with_index-32                 69         340445082 ns/op          89.79 MB/s       2942353 msg/s         7415722 B/op       5960 allocs/op
BenchmarkReader/lexer-32                     156         151886422 ns/op         338.66 MB/s       6576108 msg/s        17148071 B/op       1238 allocs/op
PASS
ok      github.com/foxglove/mcap/go/mcap        111.347s
```

Here's the code I added to the benchmark for the lexer. I believe it's comparable to what you must have done, though I'm getting 2x your throughput for the lexer case only so it's possible I'm doing something different. Curious what that would be:

```
	// lexer
	b.Run("lexer", func(b *testing.B) {
		b.StopTimer()
		buf := &bytes.Buffer{}
		writer, err := NewWriter(buf, &WriterOptions{
			Chunked:     true,
			Compression: CompressionZSTD,
		})
		require.NoError(b, err)
		channelCount := 200
		messageCount := uint64(1000000)
		messageSize := 32
		require.NoError(b, writer.WriteHeader(&Header{}))
		require.NoError(b, writer.WriteSchema(&Schema{ID: 1, Name: "empty", Encoding: "none"}))
		for i := 0; i < channelCount; i++ {
			require.NoError(b, writer.WriteChannel(&Channel{
				ID:              uint16(i),
				SchemaID:        1,
				Topic:           "/chat",
				MessageEncoding: "none",
			}))
		}
		contentBuf := make([]byte, messageSize)
		for i := uint64(0); i < messageCount; i++ {
			channelID := uint16(i % uint64(channelCount))
			_, err := rand.Read(contentBuf)
			require.NoError(b, err)
			require.NoError(b, writer.WriteMessage(&Message{
				ChannelID:   channelID,
				Sequence:    uint32(i),
				LogTime:     i,
				PublishTime: i,
				Data:        contentBuf,
			}))
		}
		require.NoError(b, writer.Close())
		b.StartTimer()
		for i := 0; i < b.N; i++ {
			s := time.Now()
			lexer, err := NewLexer(bytes.NewReader(buf.Bytes()))
			require.NoError(b, err)

			buf := make([]byte, 1024)
			var msgBytes uint64
			var readMessages uint64
			for {
				tokenType, token, err := lexer.Next(buf)
				if errors.Is(err, io.EOF) {
					break
				}
				require.NoError(b, err)

				if len(token) > len(buf) {
					buf = token
				}
				if tokenType == TokenMessage {
					readMessages++
					msgBytes += uint64(len(token))
				}
			}
			b.ReportMetric(float64(messageCount)/time.Since(s).Seconds(), "msg/s")
			b.ReportMetric(float64(msgBytes)/(time.Since(s).Seconds()*1024*1024), "MB/s")
			require.Equal(b, messageCount, readMessages)
		}
	})

```

wkalt:
pprof001

this visualization of CPU consumption in the indexed case puts >33% of the time in heap operations. For the specific file we are looking at, messages are ordered so heap is technically not needed.

We also have ~8.6% of time in map operations. I suspect that's mostly from the channel/schema lookups in NextInto:

```
(pprof) list NextInto
Total: 26.27s
ROUTINE ======================== github.com/foxglove/mcap/go/mcap.(*indexedMessageIterator).NextInto in /home/wyatt/repos/mcap/go/mcap/indexed_message_iterator.go
     890ms     15.30s (flat, cum) 58.24% of Total
      20ms       20ms    284:func (it *indexedMessageIterator) NextInto(msg *Message) (*Schema, *Channel, error) {
      40ms       40ms    285:   if !it.hasReadSummarySection {
         .       60ms    286:           err := it.parseSummarySection()
         .          .    287:           if err != nil {
         .          .    288:                   return nil, nil, err
         .          .    289:           }
         .          .    290:           // take care of the metadata here
         .          .    291:           if it.metadataCallback != nil {
         .          .    292:                   for _, idx := range it.metadataIndexes {
         .          .    293:                           _, err = it.rs.Seek(int64(idx.Offset), io.SeekStart)
         .          .    294:                           if err != nil {
         .          .    295:                                   return nil, nil, fmt.Errorf("failed to seek to metadata: %w", err)
         .          .    296:                           }
         .          .    297:                           opcode, data, err := readRecord(it.rs)
         .          .    298:                           if err != nil {
         .          .    299:                                   return nil, nil, fmt.Errorf("failed to read metadata record: %w", err)
         .          .    300:                           }
         .          .    301:                           if opcode != OpMetadata {
         .          .    302:                                   return nil, nil, fmt.Errorf("expected metadata record, found %v", data)
         .          .    303:                           }
         .          .    304:                           metadata, err := ParseMetadata(data)
         .          .    305:                           if err != nil {
         .          .    306:                                   return nil, nil, fmt.Errorf("failed to parse metadata record: %w", err)
         .          .    307:                           }
         .          .    308:                           err = it.metadataCallback(metadata)
         .          .    309:                           if err != nil {
         .          .    310:                                   return nil, nil, fmt.Errorf("metadata callback failed: %w", err)
         .          .    311:                           }
         .          .    312:                   }
         .          .    313:           }
         .          .    314:   }
         .          .    315:
      60ms       60ms    316:   for it.indexHeap.len() > 0 {
      10ms      8.70s    317:           ri, err := it.indexHeap.Pop()
      10ms       10ms    318:           if err != nil {
         .          .    319:                   return nil, nil, err
         .          .    320:           }
         .          .    321:           if ri.ChunkSlotIndex == -1 {
         .      3.91s    322:                   err := it.loadChunk(ri.chunkIndex)
         .          .    323:                   if err != nil {
         .          .    324:                           return nil, nil, err
         .          .    325:                   }
         .          .    326:                   continue
         .          .    327:           }
         .          .    328:           chunkOffset := ri.MessageOffsetInChunk
      40ms       40ms    329:           decompressedChunk := it.decompressedChunks[ri.ChunkSlotIndex]
      60ms       60ms    330:           length := binary.LittleEndian.Uint64(decompressedChunk.buf[chunkOffset+1:])
     120ms      120ms    331:           messageData := decompressedChunk.buf[chunkOffset+1+8 : chunkOffset+1+8+length]
      10ms       10ms    332:           existingbuf := msg.Data
      30ms      220ms    333:           if err := msg.PopulateFrom(messageData); err != nil {
         .          .    334:                   return nil, nil, err
         .          .    335:           }
     200ms      310ms    336:           msg.Data = append(existingbuf[:0], msg.Data...)
      30ms       30ms    337:           it.decompressedChunks[ri.ChunkSlotIndex].unreadMessages--
     150ms      1.11s    338:           channel, ok := it.channels[msg.ChannelID]
         .          .    339:           if !ok {
         .          .    340:                   return nil, nil, fmt.Errorf("message with unrecognized channel ID %d", msg.ChannelID)
         .          .    341:           }
     100ms      590ms    342:           schema, ok := it.schemas[channel.SchemaID]
         .          .    343:           if !ok && channel.SchemaID != 0 {
         .          .    344:                   return nil, nil, fmt.Errorf("channel %d with unrecognized schema ID %d", msg.ChannelID, channel.SchemaID)
         .          .    345:           }
      10ms       10ms    346:           return schema, channel, nil
         .          .    347:   }
         .          .    348:   return nil, nil, io.EOF
         .          .    349:}
```

if we can switch to slices for that I bet it would get cheaper.


On the memory side:
pprof001

we're doing a fair amount of allocation in heapPush, which is just appending to a slice. I'm not sure how to make it convenient for the code, but we should have everything we need to preallocate the indices array.

Another thought is that if this can be converted to a merge of _sorted_ lists, then only the top element of each list needs to be in the heap at a time, if a heap is used. That would mean we don't need to aggregate big indices lists (responsible for 18% of allocations here). I think this would be applicable in the case of no overlapping channels.

Annoyingly this picture also includes allocations from the test setup, accounting for > 40%. If I figured out how to drop that stuff out then the contribution of the indices appends would grow.

wkalt:
Here's the updated results after decrementing 22 bytes from my message bytes increment, noted by James.
```
[~/repos/mcap/go/mcap] (jrms/faster-message-iterator) $ go test ./... -run=^$ -bench=BenchmarkReader -benchmem -benchtime=20s -memprofile mem.prof -cpuprofile cpu.prof
goos: linux
goarch: amd64
pkg: github.com/foxglove/mcap/go/mcap
cpu: AMD Ryzen 9 7950X 16-Core Processor            
BenchmarkReader/no_index-32                  128         184627158 ns/op         165.97 MB/s       5438499 msg/s        17651637 B/op       2879 allocs/op
BenchmarkReader/with_index-32                 68         343562631 ns/op          88.49 MB/s       2899692 msg/s         7428013 B/op       5956 allocs/op
BenchmarkReader/lexer-32                     153         156128651 ns/op         197.77 MB/s       6480498 msg/s        17214019 B/op       1239 allocs/op
PASS
ok      github.com/foxglove/mcap/go/mcap        111.862s
```

wkalt:
my results on latest code:
```
[~/repos/mcap/go/mcap] (jrms/faster-message-iterator) $ go test ./... -run=^$ -bench=BenchmarkReader -benchmem -benchtime=20s -cpuprofile cpu.prof -memprofile mem.prof
goos: linux
goarch: amd64
pkg: github.com/foxglove/mcap/go/mcap
cpu: AMD Ryzen 9 7950X 16-Core Processor            
BenchmarkReader/no_index-32         	     151	 157370167 ns/op	 192.06 MB/s	   6293340 msg/s	17173000 B/op	    2866 allocs/op
BenchmarkReader/index_file_order-32 	     148	 161401921 ns/op	 195.04 MB/s	   6391049 msg/s	 6091088 B/op	    5971 allocs/op
BenchmarkReader/index_time_order-32 	     148	 161096652 ns/op	 192.76 MB/s	   6316549 msg/s	 6090255 B/op	    5960 allocs/op
BenchmarkReader/index_rev_order-32  	     150	 159409812 ns/op	 191.95 MB/s	   6289894 msg/s	 5306228 B/op	    5965 allocs/op
BenchmarkReader/bare_lexer-32       	     154	 151403426 ns/op	 213.52 MB/s	   6996691 msg/s	17123125 B/op	    1243 allocs/op
PASS
ok  	github.com/foxglove/mcap/go/mcap	199.761s
```

go/mcap/indexed_message_iterator.go
```diff
@@ -84,6 +99,7 @@ func (it *indexedMessageIterator) parseSummarySection() error {
 	}
 	defer lexer.Close()
 
+lexerloop:
```
wkalt, line 102: this is a small nitpick, but I'd consider calling this something like "Top". I recall using that convention elsewhere in Foxglove code but I don't have anything to point at - it's probably all in the DP. IMO "break Top" is a clearer expression to the reader than "lexerloop", since "Top" immediately conveys "topmost loop level over this one" where "lexerloop" would make me search for that label.

go/mcap/indexed_message_iterator.go
```diff
@@ -144,9 +165,48 @@ func (it *indexedMessageIterator) parseSummarySection() error {
...
-			return nil
+			break lexerloop
+		}
+	}
+	it.chunksOverlap = false
+	switch it.indexHeap.order {
+	case FileOrder:
+		sort.Slice(it.chunkIndexesToLoad, func(i, j int) bool {
+			return it.chunkIndexesToLoad[i].ChunkStartOffset < it.chunkIndexesToLoad[j].ChunkStartOffset
+		})
```
wkalt, line 176: is this logic correct? It seems like it will not always result in "file order" records. Imagine for instance that I have,

[chunk [100 500]]
[chunk [50 300]]

And my query is for [50-500]. What should happen? It seems like this would reorder the chunks, whereas I would expect them both in file order.

go/mcap/indexed_message_iterator.go
```diff
@@ -144,9 +165,48 @@ func (it *indexedMessageIterator) parseSummarySection() error {
...
+	case LogTimeOrder:
+		sort.Slice(it.chunkIndexesToLoad, func(i, j int) bool {
+			return it.chunkIndexesToLoad[i].MessageStartTime < it.chunkIndexesToLoad[j].MessageStartTime
+		})
+		for i := 1; i < len(it.chunkIndexesToLoad); i++ {
```
wkalt, line 181: iterating from the 1 seems a little unnatural IMO. Iterating from 0 to len(it.chunkIndexesToLoad)-1 seems a bit more natural, if there isn't something I'm missing.

go/mcap/indexed_message_iterator.go
```diff
@@ -144,9 +165,48 @@ func (it *indexedMessageIterator) parseSummarySection() error {
...
+			prev := it.chunkIndexesToLoad[i-1]
+			cur := it.chunkIndexesToLoad[i]
+			if prev.MessageStartTime < cur.MessageEndTime {
+				it.chunksOverlap = true
+				break
+			}
 		}
 	}
+	if it.chunksOverlap {
```
wkalt, line 202: I think we probably need to expand the commenting of this function, both in the docstring and inline, to describe the approach it is implementing.

go/mcap/indexed_message_iterator.go
james-rms: ChunkStartOffset is the byte offset of the chunk, not a message timestamp.

go/mcap/indexed_message_iterator.go
james-rms: yeah, i'll likely break this out to its own function instead of continuing with labelled breaks

go/mcap/indexed_message_iterator.go
wkalt: ah, thanks. Seems fine.

wkalt:
Would it be sensible at all to call this Next2 instead of NextInto? I understand "Into" conveys something but maybe that is also conveyed by the arg type.

The benefit to me seems like Next2 -> Next is a more natural change when major V2 is published, than NextInto -> Next.

wkalt:
a possible drawback is once we call something Next2, we better release V2 before we need a Next3. But I think that's not really different from NextInto. We'd look equally silly if we had Next, NextInto, NextIntoForReal, etc.

go/cli/mcap/cmd/sort.go
```diff
@@ -124,8 +124,9 @@ func sortFile(w io.Writer, r io.ReadSeeker) error {
 	}
 	schemas := make(map[uint16]*mcap.Schema)
 	channels := make(map[uint16]*mcap.Schema)
+	message := mcap.Message{}
 	for {
-		schema, channel, message, err := it.Next(nil)
+		schema, channel, _, err := it.Next2(&message)
```
wkalt, line 129: this is personal preference obviously, but I wonder if this is the calling style we should encourage vs

```
schema, channel, message, err := it.Next2(&message)
```

the effect is the same, the main difference is you can switch from the allocating to the reusing form just by changing the arg, rather than both the arg and the return value.

wkalt: I guess you'd need to change the example to declare message as a pointer value.

One other nice thing about this seems like it allows the user to handle channel, schema, and message all the same in their code, versus dealing with a pointer for channel/schema and a struct for message.

go/mcap/slicemap/slicemap.go
```diff
@@ -0,0 +1,38 @@
+// slicemap implements a generic array-map for uint16 keys and arbitrary structs.
+package slicemap
+
+import "math"
+
+// gets the item at idx, returning nil if not found.
+func GetAt[T any](items []*T, idx uint16) *T {
```
wkalt, line 7: nitpick - but if we are exposing a map-style interface, shouldn't we call these Get(k) and Put(k, v) or something? Seems like the caller should get a map abstraction and not need to understand what "At" means.

Related to that, would it be more convenient to package this as a type rather than passing the arrays in? i.e

```
 type smap[T any] []*T

func (m smap[T]) Get(key uint16) *T {
}

func (m smap[T]) Set(key uint16, value *T) {
}

func (m smap[T]) Slice() []*T {
	return m
}

func NewSmap[T any]() smap[T] {
	return make([]*T, 0)
}
```

go/mcap/indexed_message_iterator.go
```diff
@@ -254,16 +393,23 @@ func readRecord(r io.Reader) (OpCode, []byte, error) {
 	return opcode, record, nil
 }
 
-func (it *indexedMessageIterator) Next(_ []byte) (*Schema, *Channel, *Message, error) {
-	if !it.hasReadSummarySection {
-		err := it.parseSummarySection()
-		if err != nil {
+// Next2 yields the next message from the iterator, re-using the allocations from `msg` if
+// provided.
```
wkalt, line 397: re-use only occurs if the message data buffer has sufficient room. Maybe we should note that for completeness here.

go/mcap/indexed_message_iterator.go
```diff
@@ -191,50 +281,99 @@ func (it *indexedMessageIterator) loadChunk(chunkIndex *ChunkIndex) error {
...
+			return fmt.Errorf("record length extends past uint64 range: start: %d, len: %d", recordStart, recordLen)
 		}
-		offset += int(recordLen)
-		// skip message indexes for channels we don't need
-		if _, ok := it.channels[messageIndex.ChannelID]; !ok {
-			continue
+		if bufSize < recordEnd {
+			return fmt.Errorf(
+				"%s record in chunk has length %d bytes but only %d remaining in chunk",
+				op.String(), recordLen, bufSize-recordStart)
```
wkalt, line 310: String() is redundant here. Anything that implements String() can be directly formatted with %s.

go/mcap/indexed_message_iterator.go
```diff
@@ -191,50 +281,99 @@ func (it *indexedMessageIterator) loadChunk(chunkIndex *ChunkIndex) error {
...
+						maxLogTime = msg.LogTime
+					}
+					chunkSlot.unreadMessages++
 				}
 			}
+		case OpChannel, OpSchema:
+			// no-op
+		default:
+			return fmt.Errorf(
+				"expected only schema, channel, message opcodes in chunk, found %s at offset %d",
```
wkalt, line 334: Since the spec tells the user they are free to implement their own record types, should we be returning error here or skipping over unrecognized stuff?

go/mcap/indexed_message_iterator.go
```diff
@@ -286,28 +432,85 @@ func (it *indexedMessageIterator) Next(_ []byte) (*Schema, *Channel, *Message, e
...
+		if err := loadMessageAtOffset(decompressedChunk.buf, messageIndex.Offset, msg); err != nil {
 			return nil, nil, nil, err
 		}
-		channel := it.channels[message.ChannelID]
-		schema := it.schemas[channel.SchemaID]
-		return schema, channel, message, nil
+		decompressedChunk.unreadMessages--
+		it.curMessageIndex++
+		schema, channel, err := it.getSchemaAndChannel(msg.ChannelID)
+		return schema, channel, msg, err
```
wkalt, line 456: this would be a little clearer for me to read with an explicit error check:

```
schema, channel, err := it.getSchemaAndChannel(msg.ChannelID)
if err != nil {
  return nil, nil, nil, err
}
return schema, channel, msg, nil
```

that tells the reader that for sure this library won't return real values along with a non-nil error.

The other reason IMO to always do this and avoid the shortcut, is you will sometimes find libraries that do stuff like this:

```
package main

import "fmt"

type MyError struct{}

func (e MyError) Error() string {
	return fmt.Sprintf("uh oh")
}

func foo() *MyError {
	return nil
}

func bar() error {
	err := foo()
	return err
}

func main() {
	err := bar()
	if err != nil {
		fmt.Printf("error! %+v", err)
		return
	}
}
```

i.e they will type the return value with their custom error type, rather than Error. In this example bar() implements the shortcut approach. When you run this program you will get a surprise:

```
[~] $ go run main.go 
error! <nil>
```

As a library consumer you usually won't pay attention to how the library deals with errors until you have a need to. So better to always check the errors and be explicit. This program doesn't have the problem:

```
package main

import "fmt"

type MyError struct{}

func (e MyError) Error() string {
	return fmt.Sprintf("uh oh")
}

func foo() *MyError {
	return nil
}

func bar() error {
	err := foo()
	if err != nil {
		return err
	}
	return nil
}

func main() {
	err := bar()
	if err != nil {
		fmt.Printf("error! %+v", err)
		return
	}
}
```

go/mcap/parse.go
```diff
@@ -99,32 +99,40 @@ func ParseChannel(buf []byte) (*Channel, error) {
...
-		Sequence:    sequence,
-		LogTime:     logTime,
-		PublishTime: publishTime,
-		Data:        data,
-	}, nil
+	m.ChannelID = channelID
+	m.Sequence = sequence
+	m.LogTime = logTime
+	m.PublishTime = publishTime
+	m.Data = data
```
wkalt, line 125: if msg.Data was oversized, we drop the excess capacity here, which will cause more allocations as big and small messages alternate. Ideally I think we want the capacity of the msg.Data array to be monotonically increasing across calls. I'm not sure if our benchmarks surface this situation.

go/mcap/indexed_message_iterator.go
```diff
@@ -286,28 +432,85 @@ func (it *indexedMessageIterator) Next(_ []byte) (*Schema, *Channel, *Message, e
...
+func loadMessageAtOffset(decompressedChunk []byte, offset uint64, msg *Message) error {
+	length := binary.LittleEndian.Uint64(decompressedChunk[offset+1:])
+	messageData := decompressedChunk[offset+1+8 : offset+1+8+length]
+	previousData := msg.Data
+	if err := msg.PopulateFrom(messageData); err != nil {
+		return err
+	}
+	// after PopulateFrom, msg.Data is a slice of `decompressedChunk`. Copy it out so
+	// that the `decompressedChunk` buffer can be reused later.
+	msg.Data = append(previousData[:0], msg.Data...)
```
wkalt, line 503: note that if PopulateFrom performs the copy into the existing msg.Data, we can drop the copy here

go/mcap/indexed_message_iterator.go
james-rms: The spec is less friendly towards arbitrary records in chunks: https://mcap.dev/spec#chunk-op0x06

> A Chunk contains a batch of Schema, Channel, and Message records.

go/mcap/indexed_message_iterator.go
```diff
@@ -15,32 +19,54 @@ const (
 	chunkBufferGrowthMultiple = 1.2
 )
 
-// indexedMessageIterator is an iterator over an indexed mcap read seeker (as
-// seeking is required). It makes reads in alternation from the index data
-// section, the message index at the end of a chunk, and the chunk's contents.
+type decompressedChunk struct {
+	buf            []byte
+	unreadMessages uint64
+}
```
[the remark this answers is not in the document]
james-rms: I don't think wrapping these single-line field manipulations in methods reduces any complexity - I am open to suggestions about how to actually reduce it though. It feels like there's some room to do that here.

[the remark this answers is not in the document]
wkalt: /shrug that's my suggestion I guess. Encapsulate the operations.

defunctzombie:
Re-posting some thoughts I shared with James in Slack:

I don't have many strong feelings on golang things but I think ya'll can do better than `Next2` as an API name to live alongside the existing api. If it isn't possible to re-use the existing name with overloading, then something like `NextInto(...)` as James proposed sounded reasonable to me. I cannot recall exactly but I have used c/c++ like apis where you could either pass an existing object or _null_ if you wanted one allocated for you. Seems like a sane pattern.

wkalt:
The object or null approach is the one we've taken. I obviously disagree on the naming but won't block you guys. I think NextInto is kind of a dead end. What do you call the thing after that? What do you do when we ship a V2 and there is no more Next, only NextInto?

Just seems kinda strange but again, not a dealbreaker to me.

wkalt:
another option to consider, would be to simply roll a major. I wonder if there's a good reason to avoid that?

wkalt:
Final minor suggestion - I think we should change "major" and "minor" in these benchmark names to "major_overlap" and "minor_overlap" for clarity.

Anyway, final results from me:
```
BenchmarkReader/inorder/no_index-32              100     210357017 ns/op         609.76 MB/s      19980710 msg/s        18784082 B/op       6300 allocs/op
BenchmarkReader/inorder/index_file_order-32          129         180875585 ns/op         677.58 MB/s      22202914 msg/s        10451106 B/op      16981 allocs/op
BenchmarkReader/inorder/index_time_order-32      130     180912838 ns/op         675.79 MB/s      22144266 msg/s        10451253 B/op      16989 allocs/op
BenchmarkReader/inorder/index_rev_order-32           128         185554867 ns/op         647.86 MB/s      21229217 msg/s         9993184 B/op      16985 allocs/op
BenchmarkReader/inorder/bare_lexer-32                122         192062239 ns/op         600.44 MB/s      19675251 msg/s        18732559 B/op       4671 allocs/op
BenchmarkReader/minor/no_index-32                    100         215605448 ns/op         499.60 MB/s      16371129 msg/s        15390843 B/op       6275 allocs/op
BenchmarkReader/minor/index_file_order-32            132         178575332 ns/op         673.93 MB/s      22083281 msg/s        10418086 B/op      16978 allocs/op
BenchmarkReader/minor/index_time_order-32             68         362007549 ns/op         347.44 MB/s      11384992 msg/s        10447251 B/op      17721 allocs/op
BenchmarkReader/minor/index_rev_order-32              69         355197195 ns/op         353.53 MB/s      11584413 msg/s         9977055 B/op      17704 allocs/op
BenchmarkReader/minor/bare_lexer-32                  124         197898411 ns/op         586.82 MB/s      19229078 msg/s        15341271 B/op       4651 allocs/op
BenchmarkReader/major/no_index-32                    100         228166159 ns/op         550.78 MB/s      18048103 msg/s        16619558 B/op       6283 allocs/op
BenchmarkReader/major/index_file_order-32            132         176790090 ns/op         671.87 MB/s      22016043 msg/s        10418646 B/op      16992 allocs/op
BenchmarkReader/major/index_time_order-32             63         371470199 ns/op         342.28 MB/s      11215845 msg/s        16457851 B/op      17714 allocs/op
BenchmarkReader/major/index_rev_order-32              66         356158503 ns/op         356.68 MB/s      11687706 msg/s        15990102 B/op      17716 allocs/op
BenchmarkReader/major/bare_lexer-32                  120         198373003 ns/op         633.73 MB/s      20766082 msg/s        16569771 B/op       4656 allocs/op
PASS
ok      github.com/foxglove/mcap/go/mcap        513.643s
```

Behavior is close on files with small overlap, and we still have a gap with heavy overlap as expected. Looks great!

defunctzombie:
> I wonder if there's a good reason to avoid that

In golang all dependencies of a project have to use the same version? So if you make API incompatible changes then some dependency that also needs the mcap library will break if not updated to the latest? This is why golang uses the folder major versioning approach iirc? I have not done enough golang to say if its better to roll the major or adopt the `2` strategy. Can you point to examples of popular libraries that have added `2` to their APIs rather than rolling a major? Or some other approach?

wkalt:
@defunctzombie I'm not sure I'm quite following - hopefully this will be helpful -
* All dependencies of a project in go do not need to use the same version. They can use whatever tagged version they desire.
* Dependency updates are opt-in, so shipping a new version does not force any upgrade on users.
* There is no distinction between releasing a major version of the lib and adopting the /v2 suffix on import paths. The way to release a major would be to bump the major version in the tag, and do the /v2 suffix. So it's not a "rather than" - you'd do both.
* There are instructions here: https://go.dev/wiki/Modules#releasing-modules-v2-or-higher . Of the two approaches mentioned, you want to use the first one. The second approach was only relevant during a transitory period during adoption of modules, for projects that needed to support older versions of go.
* An example of a go project on a version greater than 1 is https://github.com/pierrec/lz4
* Because MCAP has released a 1.y.z already, and because you guys like semver, you don't have a choice about the v2 suffixing. If you bump the tag without suffixing the import path, the new tag won't be consumable.

While it isn't an option for MCAP, numerous projects choose to ignore SIV (semantic import versioning, this suffixing stuff) for various reasons. The way to do this is to adopt a convention that isn't real semver, such as keeping a permanent 0 or 1 major version and communicating breaking changes in other ways. Foxglove has dependencies that take this approach as well. I don't recall which offhand (protobuf maybe?) but I expect you might find a few if you executed `go get -u` in the data platform project and run the tests. I generally have no expectation that my dependencies adhere to semver, whether or not they claim to, so I consider this fine as well.

The point of SIV is to allow consumers to use multiple major versions of a lib at the same time. That's not a requirement I've ever encountered so to me personally, it's an inconvenience to update paths, docs, etc just for a version bump.

wkalt:
I think a good argument against rolling a major update into this PR, is that in order to avoid update churn it would be best if the major were specifically planned so that any breaking changes we want to put in can be bundled together. Due in part to SIV, multiple major upgrades in a row is annoying for consumers.

defunctzombie (approved):

PR #1169, opened 2024-05-07
simplify release binary download instructions, link to latest release page

sofuture:
### Changelog

Remove cli download script, and link to the github latest release page

### Docs

None

### Description

The helper script would have trouble matching architecture and release binary depending on OS/platform. This should simplify the process, and not require tracking the version.


website/docs/guides/cli.mdx
```diff
@@ -2,8 +2,6 @@
 sidebar_position: 3
 ---
 
-import CLIDownloadScript from "../../src/components/CLIDownloadScript";
```
jtbandes, line 5: Might be able to rename this file to `.md` — I think it was only made `.mdx` because of this CLIDownloadScript thing

website/src/components/CLIDownloadScript.tsx
```diff
@@ -1,15 +0,0 @@
-import { usePluginData } from "@docusaurus/useGlobalData";
-import CodeBlock from "@theme/CodeBlock";
-import React from "react";
-
-export default function CLIDownloadScript(): JSX.Element {
-  const latestVersion = usePluginData("latestCLIReleaseTag") as { tag: string };
```
jtbandes, line 6: I think there's more code around this `latestCLIReleaseTag` that may be unused now?

jtbandes: Basically it seems like you want to revert #1069 and then modify cli.md.

sofuture: Oh, I see. Makes sense 👍

website/docs/guides/cli.md
```diff
@@ -82,23 +78,23 @@ Use "mcap [command] --help" for more information about a command.
 
 Convert a ROS 1 bag file to mcap:
 
-{/* cspell: disable */}
+{/_ cspell: disable _/}
```
jtbandes, line 81: These are broken now; probably need to be removed or replaced with html comments. Since it's no longer mdx `{ }` will not work

jtbandes (approved):

PR #1170, opened 2024-05-10
spec: Clarify legal contents of chunk records

wkalt:
Prior to this commit, the wording of the spec implied that only channel, schema, and message records are allowed in chunks. This could lead tooling implementers to validate that assertion.

This is inconsistent with our support for extensible record types. Users should be able to put custom records in chunks as well, and readers that parse chunks should ignore unrecognized records rather than error on them.

### DIFF
--- a/website/docs/spec/index.md
+++ b/website/docs/spec/index.md
@@ -181,7 +181,7 @@ The message encoding and schema must match that of the Channel record correspond
 
 ### Chunk (op=0x06)
 
-A Chunk contains a batch of Schema, Channel, and Message records. The batch of records contained in a chunk may be compressed or uncompressed.
+A Chunk contains a batch of records. Readers should expect Schema, Channel, and Message records to be present in chunks, but future spec changes or user extensions may include others. The batch of records contained in a chunk may be compressed or uncompressed.
 
 All messages in the chunk must reference channels recorded earlier in the file (in a previous chunk, earlier in the current chunk, or earlier in the data section).
 


wkalt:
cc @james-rms

james-rms (commented):
Other ideas that I think are clearer ways to state this:


> A Chunk contains a batch of Schema, Channel, and Message records, in addition to any application extension records in the 0x80-0xFF opcode range.

> A Chunk contains a batch of records. The only MCAP-specified records allowed in Chunks are Schema, Channel, and Message records, but custom writers may also include records in the 0x80-0xFF opcode range.


> A Chunk contains a batch of records. Readers should expect Schema, Channel, and Message records to be present in chunks, but future spec changes or user extensions may include others.

wkalt:
thanks! adopted your second suggestion.

wkalt:
... and just updated to the third suggestion. I like that better since it gives us wiggle room for our own future extensions.

james-rms:
I tend to like my third suggestion more (sorry, i edited it in much later) - this gives us room to extend the existing spec with a small-message record one day without that being a breaking change.

james-rms:
>  The batch of records contained in a chunk may be compressed or uncompressed.

we lost this sentence - i don't think it added much, but doesn't seem related to this PR.

wkalt:
added back in. I agree it's not a good sentence, since we should ideally have users thinking of "none" as a compression algorithm.

But in fact we use "" instead of "none", and I found that way too confusing to spell out. Maybe another good change for v2.

james-rms (approved):

PR #1171, opened 2024-05-20
cli: sort does not duplicate channels and schemas on every message

james-rms:
### Changelog

- fixed: `mcap sort` would rewrite the schema and channel for every message in the output file. This is fixed.

### Description

</td><td>

</td></tr></table>


jtbandes (approved):
Can `sort_test.go` be updated to verify this?

PR #1175, opened 2024-05-30
Stop example protobuf code from leaking

kyle-basis:
### Changelog
protobufs allocated with `New()` must be `delete`d later.

### Docs
None

### Description

Leaks

</td><td>

Doesn't leak (leaks less?)

</td></tr></table>

### DIFF
--- a/website/docs/guides/cpp/protobuf.md
+++ b/website/docs/guides/cpp/protobuf.md
@@ -162,7 +162,7 @@ descriptor = protoPool.FindMessageTypeByName(it->schema->name);
 We can use this descriptor to parse our message:
 
 ```cpp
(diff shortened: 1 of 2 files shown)


website/docs/guides/cpp/protobuf.md
```diff
@@ -31,6 +31,11 @@ We also include the MCAP reader implementation:
 #include "mcap/reader.hpp"
 ```
 
+And standard library dependencies:
+```cpp
+#include <memory>
```
achim-k, line 36:
```suggestion
And standard library dependencies:

```cpp
#include <memory>
```

jtbandes (approved):
Ownership semantics seem correct according to https://protobuf.dev/reference/cpp/api-docs/google.protobuf.message/#Message.New.details

PR #1186, opened 2024-06-24
validate: check both message encoding and schema encoding

jtbandes:
The validate example was incorrect for ros2msg/cdr channels, and it was also only checking the message encoding and not the schema encoding.

### DIFF
--- a/typescript/examples/validate/scripts/validate.ts
+++ b/typescript/examples/validate/scripts/validate.ts
@@ -139,7 +139,7 @@ async function validate(
           throw new Error(`Missing schema ${record.schemaId} for channel ${record.id}`);
         }
         let messageDeserializer: (data: ArrayBufferView) => unknown;
-        if (record.messageEncoding === "ros1") {
+        if (record.messageEncoding === "ros1" && schema.encoding === "ros1msg") {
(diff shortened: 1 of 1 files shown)


typescript/examples/validate/scripts/validate.ts
```diff
@@ -139,7 +139,7 @@ async function validate(
           throw new Error(`Missing schema ${record.schemaId} for channel ${record.id}`);
         }
         let messageDeserializer: (data: ArrayBufferView) => unknown;
-        if (record.messageEncoding === "ros1") {
+        if (record.messageEncoding === "ros1" && schema.encoding === "ros1msg") {
```
snosenzo, line 142: nitty nit: feels a bit more natural to me to check schema first, then the message encoding, going from less specific to more specific.

snosenzo (approved):

typescript/examples/validate/scripts/validate.ts
jtbandes: https://github.com/foxglove/mcap/assets/14237/968f69bf-3fb4-4695-b796-6456c5a73472

PR #1191, opened 2024-07-05
Fix table output with wide columns

### LINKED ISSUE #1189: [CLI] mcap list metadata doesn't show the metadata table with huge amount of data in a single entry
**Description**

- Version: v0.0.46 (CLI)
- Platform: macOS

**Steps To Reproduce**

1. Have a mcap without metadata
2. Run `mcap list metadata your_file.mcap`
3. See a empty table of metadata.
4. Create and give execution permission to this script (or do your own):
```
#!/bin/bash

# Function to generate the key-value pairs
generate_key_value_pairs() {
  local n=$1
  local pairs=""
  for ((i=1; i<=n; i++)); do
    if [[ $i -eq $n ]]; then
      pairs+="key$i=value$i"
    else
      pairs+="key$i=value$i,"
    fi
  done
  echo $pairs
}

# Check if the correct number of arguments is provided
if [ "$#" -ne 2 ]; then
  echo "Usage: $0 N file.mcap"
  exit 1
fi

# Number of key-value pairs and the file
N=$1
METADATA_NAME=$2
FILE=$3

# Generate the key-value pairs
KEY_VALUE_PAIRS=$(generate_key_value_pairs $N)

# Construct and run the command
COMMAND="mcap add metadata -k \"$KEY_VALUE_PAIRS\" -n $METADATA_NAME $FILE"
echo "Running command: $COMMAND"
eval $COMMAND
```

5. Run it like: `./script.sh 10000 name_of_metadata your_file.mcap`

6. Run `mcap list metadata your_file.mcap`

It doesn't show nothing:
image

**Expected Behavior**

It was expected it shows the table with the metadata, at least it could trim it and show only a part of it.

Running `mcap get metadata -n name_of_metadata your_file.mcap`or `mcap get metadata -n name_of_metadata your_file.mcap | wc -l` shows that it really added, but the `mcap list metadata` shows nothing.

image

bryfox:
### Changelog
Fix: table output from list commands supports wide columns such as large amounts of metadata

### Docs
None

### Description

Previously, listing the metadata for an mcap file with many (>64kb) key/value pairs failed silently.

The Scanner used in the table formatter has a buffer limit of 64kb, and the error result was being ignored. This removes the scanner entirely from most table output. It was introduced to trim leading whitespace from lines, but the formatter accepts other options to make this happen.

I've given the info command its own formatter, which I think is appropriate because it's formatting data differently, even including tabs in its row data. The scanner's error result is now checked, though we shouldn't see this in practice for `mcap info`.

`mcap list metadata` would silently fail if a record had >64kb

</td><td>

all metadata is output as expected

</td></tr></table>

As a side effect, this also fixes the column alignment for something like `list schemas` that contains linebreaks:

```
id  name                encoding  data
24  tf2_msgs/TFMessage  ros1msg   geometry_msgs/TransformStamped[] transforms

================================================================================
MSG: geometry_msgs/TransformStamped
# This expresses a transform from coordinate frame header.frame_id
# to the coordinate frame child_frame_id
#
# This message is mostly used by the
# <a href="http://wiki.ros.org/tf">tf</a> package.
# See its documentation for more information.

Header header
string child_frame_id # the frame id of the child frame
Transform transform

```

</td><td>

```
id	name              	encoding	data
24	tf2_msgs/TFMessage	ros1msg 	geometry_msgs/TransformStamped[] transforms

  	                  	        	================================================================================
  	                  	        	MSG: geometry_msgs/TransformStamped
  	                  	        	# This expresses a transform from coordinate frame header.frame_id
  	                  	        	# to the coordinate frame child_frame_id
  	                  	        	#
  	                  	        	# This message is mostly used by the
  	                  	        	# <a href="http://wiki.ros.org/tf">tf</a> package.
  	                  	        	# See its documentation for more information.

  	                  	        	Header header
  	                  	        	string child_frame_id # the frame id of the child frame
  	                  	        	Transform transform

```

</td></tr></table>

This also makes a couple of copy fixes to the inline help for the metadata commands.

Fixes #1189.

### DIFF
--- a/go/cli/mcap/cmd/info.go
+++ b/go/cli/mcap/cmd/info.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
@@ -13,6 +14,7 @@ import (
 
 	"github.com/foxglove/mcap/go/cli/mcap/utils"
 	"github.com/foxglove/mcap/go/mcap"
+	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
 )
 
@@ -89,7 +91,9 @@ func printInfo(w io.Writer, info *mcap.Info) error {
 			header = addRow(header, "end:", "%s", decimalTime(endtime))
 		}
 	}
-	utils.FormatTable(buf, header)
+	if err := printSummaryRows(buf, header); err != nil {
+		return err
+	}
 	if len(info.ChunkIndexes) > 0 {
 		compressionFormatStats := make(map[mcap.CompressionFormat]struct {
 			count            int
@@ -166,7 +170,9 @@ func printInfo(w io.Writer, info *mcap.Info) error {
 		}
 		rows = append(rows, row)
 	}
-	utils.FormatTable(buf, rows)
+	if err := printSummaryRows(buf, rows); err != nil {
+		return err
+	}
 	if info.Statistics != nil {
 		fmt.Fprintf(buf, "attachments: %d\n", info.Statistics.AttachmentCount)
 		fmt.Fprintf(buf, "metadata: %d\n", info.Statistics.MetadataCount)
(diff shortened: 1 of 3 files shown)


james-rms (approved):
This change is good. It's not clear to me why other commands need to use the original formatter rather than this one - is there a regression, or did you not want to change too many things in this PR?

bryfox:
> It's not clear to me why other commands need to use the original formatter rather than this one

My goal was to leave the output of `mcap info` exactly as-is. I was not able to get correct formatting for both 'info' and other commands using the same formatter. They're really different formats. Most are column-aligned tables; info has left-aligned labels with tabs in the row content itself to indent some rows.

PR #1197, opened 2024-07-16
Add Rust ordered message stream implementation

### LINKED ISSUE #659: rust: support time-ordered message reading

alxhill:
### Changelog
Add OrderedMessageStream implementation for reading timestamp ordered messages in Rust.

### Docs

Happy to add a docs change!

### Description

The Python API allows reading in message-timestamp order:

```python
todo
```

This PR brings roughly the same functionality to Rust:

```rust
todo
```

</td><td>

</td></tr></table>

Fixes: https://github.com/foxglove/mcap/issues/659

### DIFF
--- a/rust/src/read.rs
+++ b/rust/src/read.rs
@@ -777,21 +777,17 @@ impl Ord for OrderedMessage<'_> {
     fn cmp(&self, other: &Self) -> Ordering {
         let mut cmp = self.message.log_time.cmp(&other.message.log_time);
 
-        if cmp.is_eq() {
-            // todo: does this match up to the python position behaviour? does it need to?
-            cmp = self.message.sequence.cmp(&other.message.sequence);
-        }
-
+        // default order is largest to smallest
         match self.reversed {
-            true => cmp.reverse(),
-            false => cmp,
+            true => cmp,
+            false => cmp.reverse(),
         }
     }
 }
 
 pub struct OrderedMessageStream<'a> {
     remaining_iter: MessageStream<'a>,
-    ordered_iter: IntoIter<OrderedMessage<'static>>,
+    ordered_msgs: BinaryHeap<OrderedMessage<'static>>,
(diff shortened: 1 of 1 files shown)


jtbandes:
Thanks for spiking out this PR. At a quick glance, it looks like this is reading all messages into memory and storing them in a heap – is that right? If so, this would not be usable with larger MCAP files, so while it is a good proof of concept, I don't think it would make sense to include in the upstream repo. Did you have additional work planned on this PR?

alxhill:
@jtbandes yeah good callout, I was looking at the Python implementation which I think roughly does the same thing iiuc.

I was looking at the go implementation which requires a seekable stream - I think the challenge with not reading the whole stream in Rust is that LinearReader moves the `buf` pointer forwards, so there isn't a natural way to seek across chunks without a refactor. Does that seem correct or do you have thoughts on an alternative?

jtbandes:
Looking at the Python now, it seems to only do the in-memory sort if using a NonSeekingReader; for a SeekingReader the ordering is handled by MessageQueue here: https://github.com/foxglove/mcap/blame/main/python/mcap/mcap/reader.py#L288

I'm not very familiar with the Rust implementation yet but it may be true that a refactor is needed - @james-rms may know more.

defunctzombie:
ping @james-rms on making a decision here on whether we want to push forward with the approach, do something different, or close

alxhill:
fwiw, happy to close this - we're not using it at the moment, and timestamp ordered reading would be better suited to a version of this library built on top of `Seek` instead of mmap

passchaos:
@james-rms hi, does Rust have a plan to support ordered message reading? In the matrix description of the supported functions, only the Rust version does not support this feature.
image

jtbandes:
The docs might be out of date, I believe you should be able to achieve this using IndexedReader: https://docs.rs/mcap/latest/mcap/sans_io/indexed_reader/struct.IndexedReader.html

PR #1199, opened 2024-07-22
Remove dead members in McapReader

kyle-basis:
### Changelog
Remove unused `startTime_` and `endTime_` members of `McapReader` that were initialized and never used.

### Docs

None

### Description

This is fairly self evident - if the code compiles it should be the correct change.


indirectlylit (approved):
thank you!

PR #1201, opened 2024-07-24
rust: Support writing without chunks

HALtheWise:
On very memory-constrained systems like microcontrollers, the overhead of storing all the chunk offset data in memory until the end of the file can be significant. To avoid that, we instead write all Channel, Schema, and Message data directly as records rather than using chunks, and rely on the reader (or an intermediate ingestion process) to generate the index information.

### Changelog
The rust library now (optionally) supports writing unchunked MCAP files for decreased memory footprint.

### Docs

None

### Description

In our internal testing, storing chunk data required >40KB of memory for several-minute logs, which is a lot for an embedded system with only 300KB of RAM total. Notably, this grows without bound during the test, since the offsets and data about all chunks needs to be stored until the end of the log when it will be written to the index.

For our usecase, it's acceptable to disable chunking entirely (along with compression and indexing), since we can add those things later during the log upload process.

We're using this internally on an ESP32-S3 and it is working well.

</td><td>

</td></tr></table>

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -108,6 +108,7 @@ pub struct WriteOptions {
     compression: Option<Compression>,
     profile: String,
     chunk_size: Option<u64>,
+    use_chunks: bool,
 }
 
 impl Default for WriteOptions {
@@ -119,6 +120,7 @@ impl Default for WriteOptions {
             compression: None,
             profile: String::new(),
             chunk_size: Some(1024 * 768),
+            use_chunks: true,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


rust/src/write.rs
```diff
@@ -389,6 +440,10 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
         // (That would leave it in an unspecified state if we bailed here!)
         // Instead briefly swap it out for a null writer while we set up the chunker
         // The writer will only be None if finish() was called.
+        if !self.options.use_chunks {
+            unreachable!("Trying to write to a chunk when chunking is disabled")
```
eloff, line 444: I would use panic!, not unreachable!. In English the word communicates your meaning more clearly, but to the compiler it means something else.

In fact this could be an `assert!(self.options.use_chunks, "Trying to write to a chunk when chunking is disabled");`

eloff (dismissed):
This looks good, and the rationale makes a lot of sense. Just one small change needed.

rust/src/write.rs
HALtheWise: Done. [unreachable!](https://doc.rust-lang.org/std/macro.unreachable.html#panics) is documented to be syntactic sugar around `panic!`, but I'm happy to reserve it's use for cases where it's locally obvious why a property is always true.

eloff: You're right, I was thinking of https://doc.rust-lang.org/std/hint/fn.unreachable_unchecked.html#

I typically favor assert for these kinds of checks, but that's just my preference. Thanks for making the change.

james-rms (commented):
seems good with a test that demonstrates that this mode produces valid MCAP.

HALtheWise:
Done

eloff (approved):
Thanks for adding the test, this looks good to me

PR #1192, opened 2024-07-05
Specify patch version and sync

bryfox:
This updates the go workspace to specify a version of go, and syncs the workspace.

When I used the existing workspace locally, go commands failed, because it could not install a toolchain for version "1.22". These (omitted patch) versions are development versions, which are removed once out of development.

The [docs](https://go.dev/doc/tutorial/workspaces) still imply that `work init` produces a file with an omitted patch version, but it seems to now init with a specific version (1.22.5), and this comment indicates it's required: https://github.com/golang/go/issues/62278#issuecomment-1693538776. Presumably I could manually install a version of 1.22, but using a specific version makes the tooling work nicely.

PR #1204, opened 2024-07-26
Update go linter

bryfox:
### Changelog
None
### Docs
None
### Description

This upgrades the version of `golangci-lint` to the latest version, and removes three deprecated linters, which now error when included because they are completely unused: https://github.com/foxglove/mcap/pull/1192#issuecomment-2252886768.


eloff (approved):
This passes cleanly now for me

PR #1206, opened 2024-07-31
go: return a bad magic error in check when parsing summary

james-rms:
### Changelog
- Changed: when attempting to read a corrupt MCAP using using the index, the reader will return an ErrBadMagic instance rather than a bare Error.

### Docs

None.

### Description

it's useful to be able to tell this error apart from errors that might result from i/o, but right now there is no simple way to determine this using `errors.Is`. This PR gives the error a publicly known type, so that callers can check for it.

</td><td>

</td></tr></table>

### DIFF
--- a/go/mcap/indexed_message_iterator.go
+++ b/go/mcap/indexed_message_iterator.go
@@ -83,7 +83,7 @@ func (it *indexedMessageIterator) parseSummarySection() error {
 	}
 	magic := buf[20:]
 	if !bytes.Equal(magic, Magic) {
-		return fmt.Errorf("not an MCAP file")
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

go/mcap/indexed_message_iterator.go
```diff
@@ -83,7 +83,7 @@ func (it *indexedMessageIterator) parseSummarySection() error {
 	}
 	magic := buf[20:]
 	if !bytes.Equal(magic, Magic) {
-		return fmt.Errorf("not an MCAP file")
+		return &ErrBadMagic{location: magicLocationEnd, actual: magic}
```
eloff, line 86: LGTM

PR #1210, opened 2024-08-12
perf(ts): use Numbers instead of BigInts

AaronO:
**⚠️  WARNING:** this is a light breaking change for consumers libs & apps, I've prototyped the change across the entire codebase but if we want to avoid any upstream breakage we could provide this as an alt call path.

### Changelog

Offsets/sizes/timestamps now use `Number` instead of `BigInt` in typescript, improving perf & heap usage

### Before

```
McapStreamReader
        3.51±0.02 op/s  Heap Used: 48.88±12.97 MB/op    Heap Total: 40.45±11.90 MB/op   ArrayBuffers: 110.11±6.36 MB/op
McapIndexedReader
        2.18±0.01 op/s  Heap Used: 69.65±2.74 MB/op     Heap Total: 59.15±3.03 MB/op    ArrayBuffers: 17.86±0.76 MB/op
McapIndexedReader_reverse
        2.20±0.00 op/s  Heap Used: 61.75±2.81 MB/op     Heap Total: 40.92±0.67 MB/op    ArrayBuffers: 16.00±0.00 MB/op
```

### After

```
McapStreamReader
        3.93±0.02 op/s  Heap Used: 31.33±1.44 MB/op     Heap Total: 23.83±1.34 MB/op    ArrayBuffers: 120.07±10.98 MB/op
McapIndexedReader
        2.49±0.03 op/s  Heap Used: 24.15±2.40 MB/op     Heap Total: 12.36±1.12 MB/op    ArrayBuffers: 3.28±3.82 MB/op
McapIndexedReader_reverse
        2.58±0.03 op/s  Heap Used: 26.39±0.64 MB/op     Heap Total: 12.84±0.34 MB/op    ArrayBuffers: 7.20±0.80 MB/op
```

### Description

MCAP currently decodes `u64` values into `BigInt`s, which comes at a cost in terms of heap usage and performance.
`u64` values are single-words on 64-bit systems, BigInts are at least 3 (2 + ceil(log2(x)/64)).

BigInts were previously used because `f64`s can only represents integers up to `2^53 - 1` with full resolution.
For integers `> 2^53` you have a max error of `2^(log2(x)-53)`.

Note that `2^53-1 ~= 9e15` or 

In MCAP those `u64`s are used for offsets, timestamps and (set-)sizes, that'd be impacted as following:
1. **offsets:** limited to 9e15 or ~9PB (peta-bytes)
2. **sizes:** limited to 9e15 or 9 quadrillion
3. **timestamps:** would be rounded to the nearest `128ns` (2^7), that error would grow up to `2048ns` (2^11) over the next 500y.
  Note: `Math.log2(Date.now()*1e6)` ~= 60.5
  Also `u64 -> f64` will preserve loose inequalities, but strict inequalities might be loosened. So (loose) event order shouldn't be impacted

IMO, the above tradeoffs are all acceptable, the main potential gotcha is that decoding and re-encoding in TS would no longer be the identity function, timestamps would be rounded. However consuming applications such as readers/viewers probably don't care about sub µs resolution. The current `128ns` loss in precision is on a similar order of magnitude of getting the clock value on some systems (that could a few hundred clock cycles)

Note: v8 is exploring optimizing this: https://issues.chromium.org/issues/42212588

### DIFF
--- a/typescript/benchmarks/index.ts
+++ b/typescript/benchmarks/index.ts
@@ -20,7 +20,7 @@ class FakeMemoryWritable {
     this.#size = 0;
   }
   position() {
-    return BigInt(this.#size);
+    return this.#size;
   }
   async write(data: Uint8Array) {
     if (data.byteLength > this.#lastWrittenData.byteLength) {
@@ -51,8 +51,8 @@ async function benchmarkReaders() {
     await writer.addMessage({
       channelId,
       sequence: i,
-      logTime: BigInt(i),
-      publishTime: BigInt(i),
+      logTime: i,
+      publishTime: i,
       data: messageData,
     });
   }
@@ -133,8 +133,8 @@ async function runWriteBenchmark({
         await writer.addMessage({
           channelId,
           sequence: i,
-          logTime: BigInt(i),
-          publishTime: BigInt(i),
+          logTime: i,
+          publishTime: i,
           data: messageData,
         });
       }
--- a/typescript/browser/src/BlobReadable.ts
+++ b/typescript/browser/src/BlobReadable.ts
@@ -10,11 +10,11 @@ export class BlobReadable implements McapTypes.IReadable {
     this.#blob = blob;
   }
 
-  public async size(): Promise<bigint> {
-    return BigInt(this.#blob.size);
+  public async size(): Promise<number> {
+    return this.#blob.size;
   }
 
-  public async read(offset: bigint, size: bigint): Promise<Uint8Array> {
+  public async read(offset: number, size: number): Promise<Uint8Array> {
     if (offset + size > this.#blob.size) {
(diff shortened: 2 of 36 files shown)


jtbandes:
The fact that v8/chrome is exploring optimizing it is interesting. The issue seems to have had quite a lot of activity; have you learned whether any bigint64 optimizations exist already today and what we need to do to take advantage of them?

> consuming applications such as readers/viewers probably don't care about sub µs resolution

I kinda understand where this is coming from but I'm not sure it applies to Foxglove's use cases (and ROS for that matter). There may be some users who don't care, but as long as there are some that do, I think we need to support original non-rounded values, unless there is some way of dynamically enabling more precision as a user setting?

PR #1212, opened 2024-08-12
perf(ts): reuse Reader whilst parsing records

AaronO:
Reduces `McapStreamReader` heap usage by ~25% and boosts throughput by ~30%.

This is both a refactor and perf boost, with more room for improvement, key changes:
- Removes StreamBuffer, hoisted into McapStreamReader
- Reuses Reader and DataView class across parse calls, only resetting them when necessary (e.g: append in McapStreamReader)
- Splits `parseRecord` into small scoped parsing functions, this itself is perf neutral (slightly positive) but facilitates future monomorphic fast paths
- Moves offsets tracking into Reader which is cleaner and faster

### Before

```
McapStreamReader
        3.48±0.03 op/s  Heap Used: 49.56±12.75 MB/op    Heap Total: 41.47±11.83 MB/op   ArrayBuffers: 112.95±6.87 MB/op
McapIndexedReader
        2.15±0.02 op/s  Heap Used: 70.02±2.84 MB/op     Heap Total: 58.34±3.36 MB/op    ArrayBuffers: 17.86±0.76 MB/op
McapIndexedReader_reverse
        2.18±0.01 op/s  Heap Used: 59.92±2.86 MB/op     Heap Total: 39.81±1.00 MB/op    ArrayBuffers: 14.58±1.42 MB/op
```

### After

```
McapStreamReader
        4.47±0.08 op/s  Heap Used: 42.35±2.23 MB/op     Heap Total: 32.93±3.76 MB/op    ArrayBuffers: 105.93±12.19 MB/op
McapIndexedReader
        2.38±0.02 op/s  Heap Used: 72.00±1.70 MB/op     Heap Total: 55.12±2.51 MB/op    ArrayBuffers: 17.86±1.85 MB/op
McapIndexedReader_reverse
        2.38±0.02 op/s  Heap Used: 63.41±1.55 MB/op     Heap Total: 39.33±0.53 MB/op    ArrayBuffers: 18.40±1.60 MB/op
```

### Followups

- Simplify parsing code further after exploring monomorphic callpaths
- Can further improve DataView/Reader churn in indexed readers


typescript/core/src/McapStreamReader.ts
```diff
@@ -89,7 +91,41 @@ export default class McapStreamReader {
     if (this.#doneReading) {
       throw new Error("Already done reading");
     }
-    this.#buffer.append(data);
+    this.#appendOrShift(data);
+    this.#reader.reset(this.#view);
+  }
+  #appendOrShift(data: Uint8Array): void {
```
jtbandes, line 97: Curious what the motivation was for merging this logic into the McapStreamReader – it reduces encapsulation a bit, but maybe there is a perf win? (if so, do we know what the specific source of the win is?)

jtbandes, line 97: nit: append() is so small I could see merging appendOrShift into it as a single function. Initially I was confused why `#reader` was not updated in appendOrShift, then I noticed it was here – that update could just be moved to the end of the function if they are combined

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
-  validateCrcs,
-}: {
-  view: DataView;
-  startOffset: number;
-  validateCrcs: boolean;
-}): { record: TypedMcapRecord; usedBytes: number } | { record?: undefined; usedBytes: 0 } {
-  if (startOffset + /*opcode*/ 1 + /*record content length*/ 8 >= view.byteLength) {
-    return { usedBytes: 0 };
+// eslint-disable-next-line @foxglove/no-boolean-parameters
+export function parseRecord(reader: Reader, validateCrcs = false): TypedMcapRecord | undefined {
```
jtbandes, line 31: there's been a consensus (or ruling?) among the team that naked boolean parameters harm readability at call sites, so if this _needs_ to be this way for performance reasons, we'll need at least a comment to that effect. (I'm not sure how big of an effect the removal of the params object had?)

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
     throw new Error(`Record content length ${recordLength} is too large`);
   }
+
   const recordLengthNum = Number(recordLength);
-  const recordEndOffset = headerReader.offset + recordLengthNum;
-  if (recordEndOffset > view.byteLength) {
-    return { usedBytes: 0 };
+
+  if (reader.bytesRemaining() < recordLengthNum) {
+    reader.offset = start; // Rewind to the start of the record
```
jtbandes, line 47: this behavior of resetting the reader is interesting and should probably be called out in the doc comment on this function, and can we add unit tests for it?

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
+      break;
   }
 
-  const recordView = new DataView(
-    view.buffer,
-    view.byteOffset + headerReader.offset,
-    recordLengthNum,
+  // NOTE: a bit redundant, but ensures we've advanced by the full record length
+  // TODO: simplify this when we explore monomorphic paths
+  reader.offset = start + RECORD_HEADER_SIZE + recordLengthNum;
```
jtbandes, line 105: should this be more like an assert()? or is there some legitimate case today where this is not already at the right offset?

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
-      };
-
-      return { record, usedBytes: recordEndOffset - startOffset };
-    }
+function parseMessage(reader: Reader, recordLength: number): TypedMcapRecord {
+  const channelId = reader.uint16();
+  const sequence = reader.uint32();
+  const logTime = reader.uint64();
+  const publishTime = reader.uint64();
+  const data = reader.u8ArrayCopy(recordLength - 22 /*channelId, sequence, logTime, publishTime*/);
```
jtbandes, line 190: nit: could keep this as 2+4+8+8 for readability (I imagine/hope that would not make a difference in performance since it would be constant-folded?)

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
-      return { record, usedBytes: recordEndOffset - startOffset };
-    }
+function parseChunk(reader: Reader, recordLength: number): TypedMcapRecord {
+  const start = reader.offset;
+  const startTime = reader.uint64();
+  const endTime = reader.uint64();
+  const uncompressedSize = reader.uint64();
+  const uncompressedCrc = reader.uint32();
+  const compression = reader.string();
+  const recordByteLength = Number(reader.uint64());
```
jtbandes, line 208: drive-by nit: I think this would be better named `recordsByteLength` since it's the length of the `records` field. As is, I was confused by how similar it is to the new parameter `recordLength`

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
+  const end = reader.offset;
+  if (recordLength - (end - start) < recordByteLength) {
```
jtbandes, line 210: nit: I find this condition harder to understand than the old version. Perhaps it can be flipped around more like `... > recordLength` as before?

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
+    throw new Error("Chunk records length exceeds remaining record size");
+  }
+  const records = reader.u8ArrayCopy(recordByteLength);
```
jtbandes, line 213: I wonder if we should actually defer to the u8ArrayCopy method for bounds-checking? Now that the Reader class is being used for this, it feels redundant for the bounds check to be done by the caller

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
+  const logTime = reader.uint64();
+  const createTime = reader.uint64();
+  const name = reader.string();
+  const mediaType = reader.string();
+  const dataLen = reader.uint64();
+  // NOTE: probably not necessary, but just in case
+  if (BigInt(reader.offset) + dataLen > Number.MAX_SAFE_INTEGER) {
+    throw new Error(`Attachment too large: ${dataLen}`);
+  }
+  if (reader.offset + Number(dataLen) + 4 /*crc*/ > startOffset + recordLength) {
```
jtbandes, line 286: similar to above comment, it feels like we could now rely on u8ArrayCopy() for the bounds check

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
+    throw new Error(`Attachment data length ${dataLen} exceeds bounds of record`);
+  }
+  const data = reader.u8ArrayCopy(Number(dataLen));
+  const crcLength = reader.offset - startOffset;
+  const expectedCrc = reader.uint32();
+  if (validateCrcs && expectedCrc !== 0) {
+    reader.offset = startOffset;
+    const fullData = reader.u8ArrayBorrow(crcLength);
+    const actualCrc = crc32(fullData);
+    reader.offset = startOffset + crcLength + 4;
```
jtbandes, line 296: nit: might be a little simpler to capture `reader.offset` after reading the `expectedCrc` and then reset it to that value here

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,415 @@
...
-      const record: TypedMcapRecord = {
-        type: "DataEnd",
-        dataSectionCrc,
-      };
-      return { record, usedBytes: recordEndOffset - startOffset };
+function parseAttachment(
+  reader: Reader,
+  recordLength: number,
+  // eslint-disable-next-line @foxglove/no-boolean-parameters
+  validateCrcs: boolean,
```
jtbandes, line 274: same comment about boolean parameters applies here

jtbandes (commented):
Seems like a nice refactor overall! Most of it makes sense to me, I just left a few questions and cleanup suggestions. I do find the duplication of `reader.offset = startOffset + recordLength;` a bit repetitive and I wonder if there is a way that could be cleaned up...

typescript/core/src/Reader.ts
```diff
@@ -103,4 +113,16 @@ export default class Reader {
     }
     return result;
   }
+
+  u8ArrayBorrow(length: number): Uint8Array {
```
defunctzombie, line 117: These methods could use a comment on what they do and how to use them.

typescript/core/src/Reader.ts
```diff
@@ -7,13 +7,25 @@ const textDecoder = new TextDecoder();
...
+  #viewU8: Uint8Array;
   offset: number;
 
   constructor(view: DataView, offset = 0) {
     this.#view = view;
+    this.#viewU8 = new Uint8Array(view.buffer, view.byteOffset, view.byteLength);
     this.offset = offset;
   }
 
+  reset(view: DataView, offset = 0): void {
```
defunctzombie, line 19: This could use a comment on what you are resetting and how it works, etc.

typescript/core/src/StreamBuffer.test.ts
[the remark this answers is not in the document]
defunctzombie: Would it be covered by all the conformance testing? These tests seem really specific to this now deleted class so seems ok?

typescript/core/src/parse.ts
AaronO: Because throwable functions are harder for v8 to optimize / not-cost-free, I plan to shift error handling up not down the call tree.

In the optimistic/happy/common paths we should never throw here, so IMO it can be handled ad-hoc, e.g: return `null` then run ad-hoc code to determine precisely what the error was.

The indirection might be a little counterintuitive, but will help keep hot path call tree fast / simple

typescript/core/src/parse.ts
AaronO: As explained above, best to move error throwing up not down the tree

typescript/core/src/parse.ts
AaronO: Will do a cleanup pass on all parse logic once perf passes are mostly settled

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,417 @@
...
-    recordLengthNum,
+  // NOTE: a bit redundant, but ensures we've advanced by the full record length
+  // TODO: simplify this when we explore monomorphic paths
+  reader.offset = start + RECORD_HEADER_SIZE + recordLengthNum;
+
+  return result;
+}
+
+function parseUnknown(reader: Reader, recordLength: number, opcode: number): TypedMcapRecord {
+  const data = reader.u8ArrayBorrow(recordLength);
```
achim-k, line 111: Shouldn't this be a `u8ArrayCopy` here? Or is the data kept internally?

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,417 @@
...
+  return {
+    type: "Footer",
+    summaryStart,
+    summaryOffsetStart,
+    summaryCrc,
+  };
+}
+
+function parseSchema(reader: Reader, recordLength: number): TypedMcapRecord {
+  const start = reader.offset;
```
achim-k, line 142: nit: Rename to `startOffset` to stay consistent with other `parse*` functions

typescript/core/src/parse.ts
```diff
@@ -1,374 +1,417 @@
...
+
+function parseChannel(reader: Reader, recordLength: number): TypedMcapRecord {
+  const startOffset = reader.offset;
+  const channelId = reader.uint16();
+  const schemaId = reader.uint16();
+  const topicName = reader.string();
+  const messageEncoding = reader.string();
+  const metadata = reader.map(
+    (r) => r.string(),
+    (r) => r.string(),
```
achim-k, line 171: Would it performance-wise be better to avoid the closures? E.g. something like that:

```ts
function readString(reader: Reader) {
  return reader.string();
}
...

const metadata = reader.map(readString, readString)
```

AaronO: Yes, it'd likely be a small net gain, but it's worth optimizing only/mainly if metadata are in the hot path (like messages).

jtbandes (approved):
Looks good, but before merging please add a comment to explain the disabling of the no-boolean-parameters lint rule (see https://github.com/foxglove/mcap/pull/1212#discussion_r1715989185 and also consider this suggestion: https://github.com/foxglove/mcap/pull/1212#discussion_r1716715764)

PR #1213, opened 2024-08-13
perf(rust): add MessageStream benches

AaronO:
Benches MessageStream on 1M messages uncompressed/zstd/lz4.

### Baseline (m1 pro)

```
mcap_read/MessageStream_1M_uncompressed
                        time:   [67.954 ms 68.277 ms 68.946 ms]
                        thrpt:  [14.504 Melem/s 14.646 Melem/s 14.716 Melem/s]
mcap_read/MessageStream_1M_lz4
                        time:   [337.77 ms 344.16 ms 350.76 ms]
                        thrpt:  [2.8510 Melem/s 2.9056 Melem/s 2.9606 Melem/s]
mcap_read/MessageStream_1M_zstd
                        time:   [381.62 ms 384.54 ms 388.35 ms]
                        thrpt:  [2.5750 Melem/s 2.6005 Melem/s 2.6204 Melem/s]
```

### DIFF
--- a/rust/benches/reader.rs
+++ b/rust/benches/reader.rs
@@ -13,7 +13,8 @@ fn create_test_mcap(n: usize, compression: Option<mcap::Compression>) -> Vec<u8>
             .profile("fooey")
             .create(Cursor::new(&mut buffer))
             .unwrap();
-        const MESSAGE_DATA: &[u8] = &[0; 42];
+        // Mock message data to align with reader benchmarks in ts
+        const MESSAGE_DATA: &[u8] = &[42; 10];
 
         let schema = Arc::new(Schema {
             name: "TestSchema".to_string(),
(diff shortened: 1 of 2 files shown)


james-rms (approved):

rust/benches/reader.rs
```diff
@@ -0,0 +1,90 @@
...
+
+fn create_test_mcap(n: usize, compression: Option<mcap::Compression>) -> Vec<u8> {
+    let mut buffer = Vec::new();
+    {
+        let mut writer = mcap::WriteOptions::new()
+            .compression(compression)
+            .profile("fooey")
+            .create(Cursor::new(&mut buffer))
+            .unwrap();
+        const MESSAGE_DATA: &[u8] = &[0; 42];
```
james-rms, line 16: any reason for 42 here?

AaronO: It was derived from:
https://github.com/foxglove/mcap/blob/c95275f41fc931f3f444c3d3b9edccb4d677c17d/typescript/benchmarks/index.ts#L40

To give the bench some message-data so cross language benches can somewhat be compared apples-to-apples, but it should instead be `[42; 10]` (10 times `42` instead of 42 zero bytes).

I'll fix that, good catch

PR #1215, opened 2024-08-14
Expose underlying writer of ROS1/2/Protobuf writers

hktonylee:
### Changelog

Expose underlying writer of ROS1/2/Protobuf writers

### Docs

None

### Description

This is needed because the underlying writer exposes more API that are needed by end-users, for example, metadata read/write.

</td><td>

</td></tr></table>


james-rms:
I don't love this solution, and would prefer to move towards a future where the serialization-specific support code can be composed with an MCAP writer rather than wrap it entirely.

hktonylee:
Absolutely agree. Composition would be much better solution. Let me close this PR first.

PR #1217, opened 2024-08-15
perf(ts): simple monomorphic message parse

AaronO:
Not fully monomorphic in the parse path, but we still see a ~10% throughput boost for this simple scoped change

### Before

```
McapStreamReader
        4.47±0.09 op/s  Heap Used: 38.73±6.90 MB/op     Heap Total: 30.21±6.23 MB/op    ArrayBuffers: 105.31±12.03 MB/op
```

### After

```
McapStreamReader
        4.88±0.10 op/s  Heap Used: 37.84±6.64 MB/op     Heap Total: 29.51±6.05 MB/op    ArrayBuffers: 106.73±12.33 MB/op
```

### DIFF
--- a/typescript/core/src/parse.ts
+++ b/typescript/core/src/parse.ts
@@ -417,3 +417,28 @@ function parseDataEnd(reader: Reader, recordLength: number): TypedMcapRecord {
     dataSectionCrc,
   };
 }
+
+// mono(morphic) parseMessage, used in the hot path
+export function monoParseMessage(reader: Reader): TypedMcapRecord | undefined {
+  const RECORD_HEADER_SIZE = 1 /*opcode*/ + 8; /*record content length*/
+  if (reader.bytesRemaining() < RECORD_HEADER_SIZE) {
+    return undefined;
+  }
+  const start = reader.offset;
+  const opcode = reader.uint8() as Opcode;
+  const recordLength = reader.uint64();
+
+  if (opcode !== Opcode.MESSAGE) {
+    reader.offset = start; // Rewind to the start of the record
+    return undefined;
+  }
+
+  const recordLengthNum = Number(recordLength);
+
+  if (reader.bytesRemaining() < recordLengthNum) {
+    reader.offset = start; // Rewind to the start of the record
+    return undefined;
+  }
+
+  return parseMessage(reader, recordLengthNum);
+}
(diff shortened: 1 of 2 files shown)


typescript/core/src/parse.ts
```diff
@@ -417,3 +417,27 @@ function parseDataEnd(reader: Reader, recordLength: number): TypedMcapRecord {
     dataSectionCrc,
   };
 }
+
+export function monoParseMessage(reader: Reader): TypedMcapRecord | undefined | null {
```
james-rms, line 421: I think returning `TypedMcapRecord | undefined | null` is really odd. I would prefer to return `undefined` in all cases and let the slow path (parseRecord) handle length checks and error throwing.

james-rms, line 421: I also prefer `maybeParseMessage` as a name, and a comment explaining why it exists.

defunctzombie:
What makes this monomorphic? What prevents "parse record" from doing similar checks to achieve the same results?

The name of the new function does not line up with the return variable names. Does the function return message records or chunk records? If chunk records then I think it should be named accordingly (if we can't make `parseRecord` dispatch better). A better comment for the function describing that behavior and when one might prefer it to `parseRecord` would help capture some of the learnings as well.

AaronO:
> What makes this monomorphic?
That its inputs & outputs are a single shape between calls.

> What prevents "parse record" from doing similar checks to achieve the same results?
`parseRecord` by construction checks opcode and then dispatches to sub-parsers that parse different kinds of messages with varying shapes, it's "megamorphic"

defunctzombie:
> `parseRecord` by construction checks opcode and then dispatches to sub-parsers that parse different kinds of messages with varying shapes, it's "megamorphic"

And in our case the input is the same (a Reader) but it is the difference in output (aka return value) that causes deoptimization?

AaronO:
> > `parseRecord` by construction checks opcode and then dispatches to sub-parsers that parse different kinds of messages with varying shapes, it's "megamorphic"
> 
> And in our case the input is the same (a Reader) but it is the difference in output (aka return value) that causes deoptimization?

Yes return values matter, outputs of one function are inputs to the other. You broadly want as much of your hot path call tree to be as "monomorphic" / specialized as possible.

typescript/core/src/parse.ts
AaronO: You're correct, differentiating between `undefined` & `null` isn't yet required or leveraged, maybe even not necessary.

I was thinking I might need to broadly monomorphise all parsing functions and lift error handling out of the hot call tree, whilst that might still be perf positive, it might not be necessary. This current monomorphized parseMessage is sufficient.

AaronO: `maybeParseMessage` isn't the best name here IMO, the fact that it might return undefined isn't the important part, the key defining quality of this function is to be optimizable / monomorphic.

PR #1222, opened 2024-08-21
fix(fseek&ftell): fix a bug of fseek/ftell on win, which fail to work when file size > 2GB(max bound of long type)

no5-aaron-wu:
when file size > 2GB, fseek will work wrong, ftell return -1 on windows

### DIFF
--- a/cpp/mcap/include/mcap/reader.inl
+++ b/cpp/mcap/include/mcap/reader.inl
@@ -51,9 +51,15 @@ FileReader::FileReader(std::FILE* file)
   assert(file_);
 
   // Determine the size of the file
+#ifdef WIN32
+  _fseeki64(file_, 0, SEEK_END);
+  size_ = _ftelli64(file_);
+  _fseeki64(file_, 0, SEEK_SET);
+#else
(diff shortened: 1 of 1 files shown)


jtbandes:
Thanks for the contribution! It looks like we already have a CI check that runs tests on Windows. Would you be able to add a test case to ensure this works as expected?

no5-aaron-wu:
my test env is VS2019 with msvc1929, read a mcap file  >  2GB will reproduce the problem, or you can just test fseek&ftell API derictly on win, i'm sorry for my pr not pass your CI due to my lack of consideration, or it should just be a issue.

no5-aaron-wu:
in LogTimeOrder will act more bad, nothing will return by readMessages API, in FileOrder still can get data although file size get wrong -1

defunctzombie:
@no5-aaron-wu Are you able to update this PR to add the unit test requested? It also looks like some other aspects of the PR are failing CI. Consider merging latest main and the swift CI issue might be resolved.

defunctzombie:
Closing this out since we've not heard back. We are happy to re-visit if you can address the issue around unit tests.

PR #1229, opened 2024-09-02
cpp: support linking statically on Windows

wrightsg:
### Changelog
Allows building and linking the `mcap` library statically on Windows.

### Docs

None

### Description

I'm using `mcap_builder` to use the `mcap` in my project.

Even though `mcap` is a header-only library, `mcap_builder` creates a library for the client application to link against.
`mcap_builder` currently always creates a shared library. I have a PR open to support creating a static library: https://github.com/olympus-robotics/mcap_builder/pull/4

Creating and linking against a static library on Windows causes linking issues because `visibility.hpp` implicitly assumes dynamic linking.

This PR adds a define that needs to be set when building the library statically. The PR is backwards compatible, i.e. the behavior remains the same as before if the define is not set. The define is named based on CMake convention, see https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html

### DIFF
--- a/cpp/mcap/include/mcap/visibility.hpp
+++ b/cpp/mcap/include/mcap/visibility.hpp
@@ -1,17 +1,21 @@
 #pragma once
 
 #if defined _WIN32 || defined __CYGWIN__
-#  ifdef __GNUC__
-#    define MCAP_EXPORT __attribute__((dllexport))
-#    define MCAP_IMPORT __attribute__((dllimport))
+#  ifndef MCAP_STATIC_DEFINE
+#    ifdef __GNUC__
+#      define MCAP_EXPORT __attribute__((dllexport))
+#      define MCAP_IMPORT __attribute__((dllimport))
+#    else
+#      define MCAP_EXPORT __declspec(dllexport)
+#      define MCAP_IMPORT __declspec(dllimport)
+#    endif
+#    ifdef MCAP_IMPLEMENTATION
+#      define MCAP_PUBLIC MCAP_EXPORT
+#    else
+#      define MCAP_PUBLIC MCAP_IMPORT
+#    endif
 #  else
-#    define MCAP_EXPORT __declspec(dllexport)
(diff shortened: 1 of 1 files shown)


defunctzombie:
A few thoughts after a quick review:

1. I don't like the undocumented and un-commented `MCAP_STATIC_DEFINE`
2. This seems like an issue created by `mcap_builder` and not specific to the mcap library here.

What is an alternative approach that does not mean introducing CMAKE specific un-documented defines to this library?

wrightsg:
I guess this could be solved in `mcap_builder` by not compiling the `mcap` library into a static or dynamic library but to consume it as header-only library (through CMake interface library).

However, `mcap` already provides the `MCAP_PUBLIC` `#define` which is only required when compiling the `mcap` library into a static or dynamic library. What is missing, is the support for the specific combination of compiling it as a static library on Windows. The other combinations (dynamic/Windows, static/Linux, dynamic/Linux) are already supported. And not supporting one specific combination is inconsistent.

Regarding documentation, I'm happy to add this to the PR if you point me to the location where it should go (comment in the `visibility.hpp` header or in some `README`?).

james-rms:
I think this is fine so long as this additional flag is explained in `cpp/README.md` under the "Including in your Project" heading.

james-rms (approved):
Will merge after readme is updated.

defunctzombie:
@wrightsg ping on updating the readme and then we can merge

wrightsg:
> @wrightsg ping on updating the readme and then we can merge

The `README` is now updated, apologies for the delay.

jtbandes:
I think this is obsolete now that https://github.com/foxglove/mcap/pull/1431 is merged — does that seem right?

defunctzombie:
Closing due to inactivity. If this is still relevant even after https://github.com/foxglove/mcap/pull/1431 as mentioned above please let us know.

PR #1234, opened 2024-09-12
go: indexed message iterator: guard against bad offsets in file

james-rms:
### Changelog

- if a file contains offsets that point outside of file bounds, the Go indexed reader implementation will return `mcap.ErrBadOffset` when attempting to use them.

### Docs

None.
### Description

There are files being uploaded to Foxglove that have invalid offsets inside them - these cause errors in `io.Seek`, which can't be told apart from true IO errors at the moment. This means if a read fails, we can't tell if it's because of a corrupt file or flaky I/O. This PR lets us disambiguate by returning a specific error from the library.

</td><td>

</td></tr></table>

### DIFF
--- a/go/mcap/indexed_message_iterator.go
+++ b/go/mcap/indexed_message_iterator.go
@@ -17,7 +17,7 @@ import (
 	"github.com/pierrec/lz4/v4"
 )
 
-var ErrBadOffset = errors.New("cannot seek to offset")
+var ErrBadOffset = errors.New("invalid offset")
 
 const (
 	chunkBufferGrowthMultiple = 1.2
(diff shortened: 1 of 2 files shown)


bryfox (commented):
It looks like this is effectively guarding that seeks happen within the file. Would it make sense for the public interface to return `UnexpectedEOF` in these cases, rather than expose this new error type?

james-rms:
I think UnexpectedEOF would be confusing here. We only return this error after we've already seeked to the end of the file and checked for end magic, so we know where the file ends and that it isn't truncated. This error indicates corruption, by the writer or through bit flips, rather than the file being truncated.

bryfox (approved):

PR #1235, opened 2024-09-12
docs: specify units for `log_time` and `publish_time`

kirillmorozov:
### Changelog

None

### Docs

None

### Description

It wasn't obvious what format `Writer.add_message` expects for `log_time` and `publish_time` argument but I've figure out that it's nanoseconds since the epoch and decided to update the docstrings so those that come after me don't have to guess.

### DIFF
--- a/python/mcap/mcap/writer.py
+++ b/python/mcap/mcap/writer.py
@@ -169,10 +169,10 @@ class Writer:
 
         :param channel_id: The id of the channel to which the message should be added.
         :param sequence: Optional message counter assigned by publisher.
-        :param log_time: Time at which the message was recorded as an integer number of nanoseconds
-            since the epoch.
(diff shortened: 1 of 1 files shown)


defunctzombie:
It does not need to be nanoseconds since epoch. The spec describes it as:

> uint64 nanoseconds since a user-understood epoch (i.e unix epoch, robot boot time, etc.)

https://mcap.dev/spec#timestamp

In our Foxglove app and data management we treat this as nanoseconds from epoch to simplify some things - but that is not a strict semantic for mcap files generally.

kirillmorozov:
Ah, alright. I used the nanoseconds since epoch to get the correct time in Foxglove Studio so naturally I thought that this should be the epoch.

I guess I should correct the docstring to reflect the spec.

defunctzombie (approved):

defunctzombie:
Thank you!

PR #1236, opened 2024-09-12
TypeScript: Reuse Reader for stream reader

achim-k:
### Changelog
Typescript: Performance improvements for readers

### Docs
None

### Description
Reintroduces the changes made in #1212, which have been reverted in #1227, with an additional fix & test (separate commit) for the buffer append logic. The original code of #1212 removed the `streamBuffer` object and implemented the buffer append ing logic directly in the `McapStreamReader` class. The append logic had a bug which caused existing data to be partially overridden by new data.

From #1212:

> Reduces `McapStreamReader` heap usage by ~25% and boosts throughput by ~30%.
>
> This is both a refactor and perf boost, with more room for improvement, key changes:
> - Removes StreamBuffer, hoisted into McapStreamReader
> - Reuses Reader and DataView class across parse calls, only resetting them when necessary (e.g: append in McapStreamReader)
> - Splits `parseRecord` into small scoped parsing functions, this itself is perf neutral (slightly positive) but facilitates future monomorphic fast paths
> - Moves offsets tracking into Reader which is cleaner and faster
>
> ### Before
>
> ```
> McapStreamReader
>         3.48±0.03 op/s  Heap Used: 49.56±12.75 MB/op    Heap Total: 41.47±11.83 MB/op   ArrayBuffers: 112.95±6.87 MB/op
> McapIndexedReader
>         2.15±0.02 op/s  Heap Used: 70.02±2.84 MB/op     Heap Total: 58.34±3.36 MB/op    ArrayBuffers: 17.86±0.76 MB/op
> McapIndexedReader_reverse
>         2.18±0.01 op/s  Heap Used: 59.92±2.86 MB/op     Heap Total: 39.81±1.00 MB/op    ArrayBuffers: 14.58±1.42 MB/op
> ```
> 
> ### After
> 
> ```
> McapStreamReader
>         4.47±0.08 op/s  Heap Used: 42.35±2.23 MB/op     Heap Total: 32.93±3.76 MB/op    ArrayBuffers: 105.93±12.19 MB/op
> McapIndexedReader
>         2.38±0.02 op/s  Heap Used: 72.00±1.70 MB/op     Heap Total: 55.12±2.51 MB/op    ArrayBuffers: 17.86±1.85 MB/op
> McapIndexedReader_reverse
>         2.38±0.02 op/s  Heap Used: 63.41±1.55 MB/op     Heap Total: 39.33±0.53 MB/op    ArrayBuffers: 18.40±1.60 MB/op
> ```


typescript/core/src/McapStreamReader.ts
```diff
@@ -92,27 +92,42 @@ export default class McapStreamReader {
...
-    const remainingBytes = this.#reader.bytesRemaining();
+    const consumedBytes = this.#reader.offset;
+    const remainingBytes = this.#view.byteLength - consumedBytes;
     const totalNeededBytes = remainingBytes + data.byteLength;
 
     if (totalNeededBytes <= this.#buffer.byteLength) {
       // Data fits in the current buffer
-      if (this.#view.byteOffset + totalNeededBytes <= this.#buffer.byteLength) {
+      if (
+        this.#view.byteOffset + this.#view.byteLength + data.byteLength <=
```
achim-k, line 105: To check if the new data can be appended (without shifting existing data), we do not care about how many bytes the reader has consumed. This was the case before, where `totalNeededBytes` was used.

achim-k:
```
this.#view.byteLength + data.byteLength != totalNeededBytes
this.#view.byteLength + data.byteLength != this.#view.byteLength - consumedBytes + data.byteLength;
this.#view.byteLength != this.#view.byteLength - consumedBytes;
```

typescript/core/src/McapStreamReader.ts
```diff
@@ -89,7 +91,61 @@ export default class McapStreamReader {
...
       throw new Error("Already done reading");
     }
-    this.#buffer.append(data);
+    this.#appendOrShift(data);
+  }
+  #appendOrShift(data: Uint8Array): void {
+    /** Add data to the buffer, shifting existing data or reallocating if necessary. */
+    const consumedBytes = this.#reader.offset;
+    const remainingBytes = this.#view.byteLength - consumedBytes;
+    const totalNeededBytes = remainingBytes + data.byteLength;
```
defunctzombie, line 100: How is `totalNeededBytes` correct here? `remainingBytes` is what has not been used in the buffer and then `data.byteLength` is the size of the incoming data. Does `totalNeededBytes` represent the amount of bytes we _need_ to fit the data? I don't understand what it is meant to represent.

This also makes it hard to answer whether the _if_ statement on the next line is correct. It compares `totalNeededBytes` to the length of the buffer which assumes that totalNeededBytes is how much is needed to store existing + new data but it doesn't seem obvious to me that `totalNeededBytes` includes existing data?

typescript/core/src/McapStreamReader.ts
```diff
@@ -50,7 +50,9 @@ type McapReaderOptions = {
  * ```
  */
 export default class McapStreamReader {
-  #buffer = new StreamBuffer(MCAP_MAGIC.length * 2);
+  #buffer = new ArrayBuffer(MCAP_MAGIC.length * 2);
+  #view = new DataView(this.#buffer);
+  #reader = new Reader(this.#view, MCAP_MAGIC.length * 2); // Cursor starts at end of initial buffer
```
jtbandes, line 55: I guess I didn't catch this in the original PR review, but why do we start with an offset that is not 0? Perhaps we should start with `#view = new DataView(this.#buffer, 0, 0)` instead and then start the reader with an offset of 0? (Not sure if that will work but it makes more sense to me)

typescript/core/src/McapStreamReader.ts
jtbandes: After initially writing a comment in support of roman's statement, I was about to press the submit button but then I double-checked and now I think I actually understand what is happening here — but I think we should find a better name for `remainingBytes`. It's not the amount of _capacity remaining_ in the buffer, but it's actually the number of _stored bytes that have not been consumed yet_. I think a better name would be `unconsumedBytes` (to contrast with `consumedBytes`) or `unreadBytes`.

The `totalNeededBytes` are the number of bytes we will need to store – this includes the ones which haven't been consumed yet + the new ones in `data`.  Maybe that should be named `neededCapacity`?

jtbandes (commented):
I wonder if this would've been caught by the [tests that were deleted](https://github.com/foxglove/mcap/pull/1212#discussion_r1715949551)...? I guess we'll never know 😄

typescript/core/src/McapStreamReader.ts
achim-k: Agreed, it looks odd. I will change it to `#view = new DataView(this.#buffer, 0, 0)`, even though I think that starting with an empty buffer would be also OK

defunctzombie (approved):
nit: The "buffer growing" bits of logic could be in a separate function to make it easier to follow the higher level code but I don't feel strongly about that.

PR #1051, opened 2024-01-11
go/cli: mcap info show channels count

pezy:
### Public-Facing Changes

show channels count like `channels(45):` in `mcap info` print.

### Description

The channels list printed by `mcap info` might have discontinuous sequence numbers due to certain reasons (for example, if one of the topics in the original recorded data has zero messages, then the mcap generated after using `mcap filter` will exhibit this). Therefore, it's important to visually see the number of channels.

PR #1243, opened 2024-09-25
`mcap info`: add channel count

james-rms:
Merging #1051 with a small tweak to use Statistics's channel count if available.


go/cli/mcap/cmd/info.go
```diff
@@ -174,9 +174,11 @@ func printInfo(w io.Writer, info *mcap.Info) error {
 		return err
 	}
 	if info.Statistics != nil {
+		fmt.Fprintf(buf, "channels: %d\n", info.Statistics.ChannelCount)
 		fmt.Fprintf(buf, "attachments: %d\n", info.Statistics.AttachmentCount)
 		fmt.Fprintf(buf, "metadata: %d\n", info.Statistics.MetadataCount)
 	} else {
+		fmt.Fprintf(buf, "channels: %d\n", len(chanIDs))
```
jtbandes, line 181: 🤔 this strikes me as odd — it looks like this list of channels is built during `parseSummarySection()`. What if the mcap was written without repeating channels in the summary section? Would this show a 0? What if there is no summary (maybe we fail earlier and don't get here in this case?)

I'm looking to be convinced that we shouldn't just say `"channels: unknown\n"` like we do for the other items here. For example, we could have chosen to get the attachment count from the AttachmentIndex records in the summary section, but we didn't.

jtbandes (approved):

PR #1244, opened 2024-09-26
rust: add sans-io linear reader

james-rms:
### Changelog
- Rust: Added a `sans_io` module to allow users to read MCAP data while remaining agnostic to I/O. Replace the internal implementation details of both the memory-mapped I/O and tokio-async read modules with this sans_io reader.

### Docs

See generated rust docs.

### Description

As we need more functionality from the async MCAP reader, I feel less comfortable adding those and not also ensuring they're available for the memory-mapped reader. 

This PR strips out all of that code and moves it into a [sans-io](https://sans-io.readthedocs.io/) reader module. The sans-io reader is a state machine with no dependency on any particular I/O framework.

</td><td>

</td></tr></table>


rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,894 @@
...
+    }
+}
+
+impl Default for LinearReader {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+pub enum ReadAction<'a> {
```
james-rms, line 624: I feel like this could have a better name.

rust/src/read.rs
```diff
@@ -23,8 +19,10 @@ use enumset::{enum_set, EnumSet, EnumSetType};
 use log::*;
 
 use crate::{
-    io_utils::CountingCrcReader,
     records::{self, op, Record},
+    sans_io::read::{
```
eloff, line 23: I don't love the name, sans_io::read seems like a oxymoron. I guess mcap is taken. Maybe mcap_reader or buffered_reader or something like that? Not sure, it's not a hill I will die on.

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,894 @@
...
+    fn get_decompressor(&mut self, name: &str) -> McapResult<Box<dyn Decompressor>> {
+        if let Some(decompressor) = self.decompressors.remove(name) {
+            return Ok(decompressor);
+        }
+        match name {
+            #[cfg(feature = "zstd")]
+            "zstd" => Ok(Box::new(zstd::ZstdDecoder::new())),
+            #[cfg(feature = "lz4")]
+            "lz4" => Ok(Box::new(lz4::Lz4Decoder::new()?)),
+            "" => Ok(Box::new(NoneDecompressor {})),
```
eloff, line 246: Does it make sense to have a NoneDecompressor instead of using Optional<None> here?

The latter seems like it would be lower overhead, no need to allocate memory or go through the dynamic method dispatch.

*edit*: it looks like NoneDecompressor copies from one buffer to the other, and you'd have to otherwise inline that code after checking for None. I guess this is fine.

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,894 @@
...
+                        Bounds(input) => input,
+                        Remainder(want) => return self.request(want),
+                    };
+                    let input = &self.uncompressed_data[start..end];
+                    let opcode = input[0];
+                    let record_length: u64 = u64::from_le_bytes(input[1..9].try_into().unwrap());
+                    // This is the start of the data end record, so stop loading new data into the
+                    // data section hasher. At this point, the data section hasher will already contain
+                    // the opcode and length bytes for the data end record, which is sad because
+                    // the data section ends at the end of the previous record. However,
```
eloff, line 329: However? Don't leave me hanging!

rust/src/sans_io/read.rs
eloff: ReadOperation? McapReadStep? ReaderState?

I guess ReadAction is fine.

eloff (commented):
I'll have to look at this some more when I get a chance.

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+            }
+
+            let tail = match &self.from {
+                File => self.uncompressed.tail(),
+                Chunk(state) => match &state.decompressor {
+                    Some(_) => self.compressed.tail(),
+                    None => self.uncompressed.tail(),
+                },
+            };
+            let written_region = &tail[..written];
```
jtbandes, line 360: The comments I can find for tail() say "returns a mutable view of the un-written part of the buffer" – when did this become written? Does the comment need to be updated?

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+                        }
+                        self.uncompressed.mark_written(written);
+                        state.compressed_remaining -= written as u64;
+                    }
+                },
+            }
+        }
+        // keep processing through the data we have until we need more data or can yield a record.
+        loop {
+            // check if we have consume all uncompressed data in the last iteration - if so,
```
jtbandes, line 384:
```suggestion
            // check if we have consumed all uncompressed data in the last iteration - if so,
```

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+                        clamp_to_usize(chunk_state.compressed_remaining),
+                        res.need,
+                    ))),
+                }
+            }
+        }
+    }
+
+    // Consume `amount` bytes of the uncompressed input buffer if enough is available. On failure,
+    // return the extra amount required as an error value.
```
jtbandes, line 679: why does consume not :just: produce a ReadAction when it needs more input?

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+                                            return Err(McapError::BadDataCrc {
+                                                saved: end.data_section_crc,
+                                                calculated,
+                                            });
+                                        }
+                                    }
+                                    _ => unreachable!("should not recieve any other record type"),
+                                }
+                            }
+                            self.data_section_hasher = None;
```
jtbandes, line 475: why bother clearing this?

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+                        Remainder(remainder) => return self.request(remainder),
+                    };
+                    let input = self.uncompressed.view(span);
+                    let compression_string_length =
+                        u32::from_le_bytes(input[28..32].try_into().unwrap());
+                    let true_chunk_header_len =
+                        min_chunk_header_len + compression_string_length as usize;
+                    if len < true_chunk_header_len as u64 {
+                        self.failed = true;
+                        return Err(McapError::RecordTooShort {
```
jtbandes, line 550: not quite following the flow here - why doesn't this `len < true_chunk_header_len` turn into a `request()`? (if the bytes read so far were enough to determine the `compression_string_length` but not enough to actually read the full chunk header?)

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+///     },
+///     ReadAction::GetRecord{ opcode, data } => {
+///       let record = mcap::parse_record(opcode, data)?;
+///       // do something with the record...
+///     }
+///   }
+///   Ok(())
+/// }
+/// ```
+pub enum ReadAction<'a> {
```
jtbandes, line 741: trying to make these verbs feels a bit weird – maybe something like `ReadResult`? and the case names would be something like `Record` & `Incomplete` or `NeedsMore`?

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+    /// Fill the input buffer with raw bytes from the MCAP file. call `set_filled()` with the number
+    /// of new bytes after the read operation.
+    Fill(InputBuf<'a>),
```
jtbandes, line 744: what's behind the decision to return an InputBuf with each "action"? for contrast, the typescript stream reader's `nextRecord` method (roughly equivalent to `next_action`) returns an optional record, and there's a separate `append` method for the user to provide more data. I see that it can be useful for the user to know a lower bound for how much data is needed, but what if it's more efficient for the user to provide more than that at once (e.g. a full page)?

jtbandes (approved):

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+pub struct InputBuf<'a> {
+    pub buf: &'a mut [u8],
+    last_write: &'a mut Option<usize>,
+}
+
+impl<'a> InputBuf<'a> {
+    /// Notify the reader that `written` new bytes are available. Only call this method after
+    /// copying data into [`self.buf`].
+    pub fn set_filled(&'a mut self, written: usize) {
+        *self.last_write = Some(written);
```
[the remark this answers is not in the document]
james-rms: What makes you think that? I originally had that behavior (where calling set_filled twice would add the two written values together) but felt it was unintuitive for a function called `set_*`. I tried making this clear in the docstring as well.

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+        match name {
+            #[cfg(feature = "zstd")]
+            "zstd" => Ok(Some(Box::new(zstd::ZstdDecoder::new()))),
+            #[cfg(feature = "lz4")]
+            "lz4" => Ok(Some(Box::new(lz4::Lz4Decoder::new()?))),
+            "" => Ok(None),
+            _ => Err(McapError::UnsupportedCompression(name.into())),
+        }
+    }
+    fn return_decompressor(&mut self, mut decompressor: Box<dyn Decompressor>) -> McapResult<()> {
```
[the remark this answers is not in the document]
james-rms: release seems fine - this is an internal method

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+        decompressor.reset()?;
+        self.decompressors
+            .insert(decompressor.name().into(), decompressor);
+        Ok(())
+    }
+
+    /// Yields the next action the caller should take to progress through the file.
+    pub fn next_action(&mut self) -> Option<McapResult<ReadAction>> {
+        self.next_action_inner().transpose()
```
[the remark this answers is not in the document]
james-rms: The transpose here is because when a function returns `Option<Result<...` , you can't use `?` syntax to return errors.  Having an inner method that returns `Result<` lets me use intuitive error return syntax while writing the big state machine logic, then i can transpose it to provide the more idiomatic `Option<Result<` iterator return value.

rust/src/sans_io/read.rs
james-rms: `len` is the record length that we got from the file right after the chunk opcode. If that length is shorter than`true_chunk_header_len`, that means the declared record length is too short to contain all the fields for a valid chunk record. If we request more bytes at this point, we'll start reading into the next record.

rust/src/sans_io/read.rs
james-rms: I'm open to that - let me try renaming and see how it feels.

rust/src/sans_io/read.rs
james-rms: By clearing this we signal to ourselves that we're not in the data section anymore, and we do not need to keep updating a CRC value every time more file data is read in.

rust/src/sans_io/read.rs
james-rms: Some notes here:

### 1. This PR
- Caller doesn't need any buffers of their own. Reader has internal buffers U and C, which contain uncompressed and compressed data respectively.
- LinearReader alternates between yielding immutable record views into U or requesting more data into either U or C. Caller fulfils these requests directly with `read()` calls.
- if the caller wants to read pages at a time, they can allocate their own buffer (or use std::io::BufReader) to buffer reads before passing any data to the LinearReader.

### 2. Typescript-like `append()` API
- Caller feeds bulk data into the linear reader in a single call. This can mean one of:
  - The allocate their own large buffer A, fill it with MCAP data with a read() call (or it's a memory-mapped file), then calls `append(&[u8])` to copy it into the LinearReader's internal buffer B.
  - The caller calls `reserve(n) -> &mut [u8]` on the LinearReader, which extends the reader-owned buffer B and returns something for the caller to read() into.
- Reader processes through this data, yielding views into the reader-owned buffer B OR returning a signal that it needs more data.
- Since B contains a mix of compressed and uncompressed data, the reader would also need a second internal buffer C to decompress into.

My judgement is that (2) ends up with the same number of copied bytes, but requires fewer copy operations and may be more cache efficient.

rust/src/sans_io/read.rs
james-rms: The answer is the borrow checker fights real hard when you do that. I need to come up with the mutable return buffer right at the moment of returning, otherwise too many mutable references are in scope at the same time.

rust/src/sans_io/read.rs
james-rms: This got written by the caller when they updated the InputBuf, but the caller did not update the start or end cursors then.

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1014 @@
...
+                self.at_eof = true;
+            }
+
+            let tail = match &self.from {
+                File => self.uncompressed.tail(),
+                Chunk(state) => match &state.decompressor {
+                    Some(_) => self.compressed.tail(),
+                    None => self.uncompressed.tail(),
+                },
+            };
```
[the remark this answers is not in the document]
james-rms: For something that only comes up twice (? haven't checked) in this file, I feel adding a layer of indirection wouldn't help things much.

rust/src/sans_io/read.rs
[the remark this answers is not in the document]
jtbandes: I think it was partly because of the wording of the comment:

> Notify the reader that `written` new bytes are available.

"new bytes" sounds like "new bytes since the last time this function was called".

rust/src/sans_io/read.rs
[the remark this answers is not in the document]
jtbandes: 👍 might be nice to have a brief comment about this

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1017 @@
...
+    pub validate_summary_section_crc: bool,
+}
+
+impl LinearReaderOptions {
+    pub fn with_skip_start_magic(self, skip_start_magic: bool) -> Self {
+        Self {
+            skip_start_magic,
+            ..self
+        }
+    }
```
bennetthardwick, line 130: nit: you could save two lines if you wanted

```suggestion
    pub fn with_skip_start_magic(mut self, skip_start_magic: bool) -> Self {
        self.skip_start_magic = skip_start_magic;
        self
    }
```

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1017 @@
...
+    GetRecord { data: &'a [u8], opcode: u8 },
+}
+
+fn len_as_usize(len: u64) -> McapResult<usize> {
+    len.try_into().map_err(|_| McapError::TooLong(len))
+}
+
+fn clamp_to_usize(len: u64) -> usize {
+    len.try_into().unwrap_or(usize::MAX)
+}
```
bennetthardwick, line 683: You should just be able to do `as usize` here right?

It probably doesn't matter here but `try_into` + `MAX` generates more assembly: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=e1784ebf677852e59782a0755b879292

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1017 @@
...
+    decompressor: Option<Box<dyn Decompressor>>,
+    uncompressed_data_hasher: Option<crc32fast::Hasher>,
+    compressed_remaining: u64,
+    uncompressed_len: u64,
+    padding_after_compressed_data: usize,
+    crc: u32,
+}
+
+#[derive(Default)]
+struct RWBuf {
```
bennetthardwick, line 50: nit: the naming conventions is normally to do acronyms as one word, i.e.

```rust
struct RwBuf {
  ..
}
```

rust/src/sans_io/read.rs
```diff
@@ -0,0 +1,1017 @@
...
+            }
+        }
+    }
+}
+
+// impl Default for LinearReader {
+//     fn default() -> Self {
+//         Self::new()
+//     }
+// }
```
bennetthardwick, line 665:
```suggestion
```

rust/src/sans_io/read.rs
james-rms: `as usize` will truncate the MSB rather than saturate, which is what I want here.

bennetthardwick: Ahh of course 🤦🏻

bennetthardwick (approved):

PR #1245, opened 2024-09-30
feat: add ids to channel and schema types

alxhill:
### Changelog
Add `id` fields to the `Schema` and `Channel` structs.

### Docs
None

### Description

The Channel ID is needed to be able to distinguish two channels with the same topic name, which are not guaranteed to be unique. I've also added the ID to the Schema type for consistency but this is not strictly necessary for us/our use case.

One challenge here is that the Channel and Schema types are used for both read and write workflows, but for write workflows it would be ignored (e.g `add_channel` generates a new channel id). In this PR I ignored the additional field, but would be happy to refactor another way. Some options I came up with:
1. make the field `Option<u16>` and set it on read. Simple enough, but makes the contract of when it will/won't be set (or how it'll behave if the field is set on write) unclear.
1. create a new `MessageStream` that returns message header with the message (or some other new combined type)
1. Split the `Channel` struct into readable/writeable versions. Seems like a fairly invasive change but may be helpful in the long-run?


james-rms:
I like the idea, but I don't think we should accept these structs in APIs that return channel IDs for the writer. I can see if i can integrate this change into a larger one that updates the writer API.

alxhill:
agree, this doesn't feel quite right. let me know if I can assist with any of the broader API changes, we're keen to use it more moving forwards

james-rms:
@alxhill https://github.com/foxglove/mcap/pull/1297 is my response to this issue, please chime in and review.

alxhill:
@james-rms didn't have a chance to deep-dive but the new approach looks great in that PR, thanks!

PR #1246, opened 2024-10-09
python: specify packages to install for each module

james-rms:
### Changelog
- Fixed: python packages `mcap`, `mcap-protobuf-support`, `mcap-ros1-support`, `mcap-ros2-support` now only install their main packages rather than all packages including tests.
- Fixed: Add type annotations to `__init__` methods with no arguments implementing this guide: https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods

### Description

Rely on setuptools' auto-discovery mechanism to find the python packages we're trying to expose while auto-excluding `tests`, which were previously being included into users' install directories.

### Testing
- Inspected wheel contents using `python -m zipfile --list mcap_protobuf_support-0.5.2-py3-none-any.whl` to ensure that the whole library content is present and no test directories are present.


indirectlylit (approved):
According to [the docs](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#automatic-discovery):

> Automatic discovery will only be enabled if you don’t provide any configuration for `packages` and `py_modules`. If at least one of them is explicitly set, automatic discovery will not take place.

And, during automatic discovery, `tests` is part of `FlatLayoutPackageFinder.DEFAULT_EXCLUDE` and `FlatLayoutModuleFinder.DEFAULT_EXCLUDE`.

So if I'm understanding all of that correctly, it might suffice to simply delete these config lines and let the [flat-layout/adhoc](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#flat-layout) discovery take place, and tests will be automatically excluded.

It sounds like this is a different codepath from `find:`, which is documented under [custom discovery](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#custom-discovery).

All that said, I think the change you made here is also fine: being explicit is safe and straightforward.

python/mcap-ros1-support/setup.cfg
```diff
@@ -12,10 +12,10 @@ classifiers =
 
 [options]
 install_package_data = True
-install_requires = 
+install_requires =
     mcap>=0.0.14
     pyyaml
-packages = find:
+packages = mcap_ros1
```
indirectlylit, line 18: It might work to just delete this?

```suggestion
```

james-rms: You're right! thanks, nice find.

PR #1247, opened 2024-10-10
go: annotate unexpected token errors

bennetthardwick:
### Changelog

Adds in a `UnexpectedTokenError` error class

### Docs

None

### Description

Adds in an `UnexpectedTokenError` error class for when the reader expects a
certain record but something else was retrieved. It could be useful to
know when this error happens to differentiate it from io-related errors.

invalid header / metadata come through as generic errors

</td><td>

invalid header / metadata errors can be differentiated

</td></tr></table>


go/mcap/error.go
```diff
@@ -0,0 +1,24 @@
+package mcap
```
james-rms, line 1: maybe `errors.go`, and move the other errors into this file as well? There are a few error sentinel values lying around: https://pkg.go.dev/github.com/foxglove/mcap/go/mcap#pkg-variables

go/go.work.sum
```diff
@@ -2,6 +2,7 @@ cloud.google.com/go/firestore v1.6.1/go.mod h1:asNXNOzBdyVQmEU+ggO8UPodTkEVFW5Qx
 github.com/armon/go-metrics v0.3.10/go.mod h1:4O98XIr/9W0sxpJ8UaYkvjk10Iff7SnFrb4QAOwNTFc=
 github.com/coreos/go-semver v0.3.0/go.mod h1:nnelYz7RCh+5ahJtPPxZlU+153eP4D4r3EedlOD2RNk=
 github.com/coreos/go-systemd/v22 v22.3.2/go.mod h1:Y58oyj3AT4RCenI/lSvhwexgC+NSVTIJ3seZv2GcEnc=
+github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
```
james-rms, line 5: I also get this occasionally locally - possibly something from a dependabot PR. i'm ok with merging it here.

james-rms (commented):
Idea seems right - need to bump the version in `version.go`. It looks like I missed this the last time, it needs to be bumped to `v1.7.0`.

go/mcap/error.go
bennetthardwick: Moved everything into `errors.go`. Fixed up the `UnexpectedTokenError` to be `ErrUnexpectedToken` to match the convention and implemented `error` for the pointer instead.

PR #1252, opened 2024-10-25
[C++] Add currentChunkStart() / currentChunkEnd() to McapWriter

jhurliman:
### Changelog
Adds two getter methods to the C++ `McapWriter` class to expose the in-progress chunk's start/end timestamps.

### Docs

None

### Description

The C++ McapWriter class currently supports one built-in rule for automatic chunking which is based on the sum of uncompressed message size. In addition to (or in lieu of) this rule, it is often useful to split chunks based on a fixed timing cadence such as every N seconds. This PR exposes `McapWriter::currentChunkStart()` and `McapWriter::currentChunkEnd()` which can be used for that calculation.

### DIFF
--- a/cpp/mcap/include/mcap/writer.hpp
+++ b/cpp/mcap/include/mcap/writer.hpp
@@ -395,6 +395,18 @@ public:
   Status write(const Metadata& metadata);
 
   /**
+   * @brief Returns the timestamp of the earliest message in the current chunk,
+   * or MaxTime if no chunk is in progress.
+   */
+  Timestamp currentChunkStart() const;
+
+  /**
+   * @brief Returns the timestamp of the latest message in the current chunk, or
+   * zero if no chunk is in progress.
+   */
+  Timestamp currentChunkEnd() const;
+
+  /**
    * @brief Current MCAP file-level statistics. This is written as a Statistics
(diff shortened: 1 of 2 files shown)


james-rms:
I think i would prefer to take the approach that the rust writer takes, where either you can use a built-in chunking rule or have a single method to manually split whenever you want. See: https://docs.rs/mcap/latest/mcap/write/struct.WriteOptions.html#method.chunk_size
https://docs.rs/mcap/latest/mcap/write/struct.Writer.html#method.flush

jhurliman:
@james-rms is there an argument for keeping these member variables private with no public getters?

jhurliman:
> I think i would prefer to take the approach that the rust writer takes, where either you can use a built-in chunking rule or have a single method to manually split whenever you want. See: https://docs.rs/mcap/latest/mcap/write/struct.WriteOptions.html#method.chunk_size https://docs.rs/mcap/latest/mcap/write/struct.Writer.html#method.flush

The C++ writer has a manual chunk closing method already.

jhurliman:
This PR has been merged into a fork of MCAP that will be maintained at <https://github.com/Verdant-Robotics/mcap>

PR #1263, opened 2024-10-30
CLI: accept RFC3339-formatted dates in relevant arguments

james-rms:
#### Added
CLI: `mcap add attachment` accepts RFC3339-formatted dates to specify the log and creation times for the new attachment.
CLI: `mcap filter` accepts RFC3339-formatted dates for start and end times.
### Docs

None.

### Description

Adds the ability for users to specify a date when filtering MCAP files, or adding attachments.

I did a quick pass to see if there are any other arguments where we accept a date or date-like value, I can't see any more.

### DIFF
--- a/go/cli/mcap/cmd/attachment.go
+++ b/go/cli/mcap/cmd/attachment.go
@@ -14,8 +14,10 @@ import (
 
 var (
 	addAttachmentLogTime      uint64
+	addAttachmentLogDate      string
 	addAttachmentName         string
 	addAttachmentCreationTime uint64
+	addAttachmentCreationDate string
 	addAttachmentFilename     string
 	addAttachmentMediaType    string
 )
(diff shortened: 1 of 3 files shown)


go/cli/mcap/cmd/filter.go
```diff
@@ -416,6 +441,18 @@ usage:
 			0,
 			"messages with log times before nanosecond-precision timestamp will be included.",
 		)
+		startDate := filterCmd.PersistentFlags().String(
+			"start-date",
```
defunctzombie, line 445: nit `start-datetime`? Since it is both a date and time? Probably I'd even say that `start` is sufficient and can accept the start string as RFC3339 or the nanosecond variant. Not "real" reason that the program can't figure it out between those two formatss

james-rms: should `start` and `start-secs` accept an RFC3339 formatted timestamp in that world?

defunctzombie: `start-secs` I'd have a harder time making an argument for given it says "secs". `start` seems to be asking for flexible support input formats

go/cli/mcap/cmd/attachment.go
```diff
@@ -187,11 +195,19 @@ func init() {
 	addAttachmentCmd.PersistentFlags().StringVarP(
 		&addAttachmentMediaType, "content-type", "", "application/octet-stream", "content type of attachment",
 	)
-	addAttachmentCmd.PersistentFlags().Uint64VarP(
-		&addAttachmentLogTime, "log-time", "", 0, "attachment log time in nanoseconds (defaults to current timestamp)",
+	addAttachmentCmd.PersistentFlags().StringVarP(
```
sofuture, line 198: Nit: I think these are fine on a single line and easier to read

sofuture (approved):

defunctzombie (commented):
👍 from me on the user-facing aspects of this. Will let some `golang` wizards do the code review part.

go/cli/mcap/cmd/filter.go
```diff
@@ -392,30 +427,43 @@ usage:
...
+			"",
+			"only include messages logged at or after this time. Accepts integer nanoseconds or RFC3339-formatted date.",
+		)
 		startSec := filterCmd.PersistentFlags().Uint64P(
 			"start-secs",
 			"s",
 			0,
-			"messages with log times after or equal to this timestamp will be included.",
+			"only include messages logged at or after this time. Accepts integer seconds."+
+				"Ignored if `--startk` or `--start-nsecs` are used.",
```
james-rms, line 441:
```suggestion
				"Ignored if `--start` or `--start-nsecs` are used.",
```

Issue #1254, opened 2024-10-26
c++: Chunk records are not immediately flushed to disk on write


jhurliman:
**Description**

It's often repeated that chunk size should be dialed in for MCAP writing based on how much data you are willing to lose, traded off vs the performance overhead of writing data to disk too frequently, the decreased compression ratio, and the container+index overhead of additional Chunk records.

However, the current C++ implementation uses `std::fwrite()` which introduces an additional in-memory buffer (inside the standard library) before data is actually scheduled for committing to disk. This is fairly critical for unchunked mode where every 4 and 8 byte field of MCAP records turn into a `std::fwrite()` call, but it introduces a second unnecessary layer of buffering in chunked mode where the chunk is the in-memory buffer and `std::fwrite()` will also buffer and not necessarily flush to disk right away.

One option would be for the `IWritable` interface to expose a `flush()` method that is explicitly called after writing a chunk, which would in turn call `std::fflush(file_)` in the `FileWriter` implementation.

Note that some users are operating under the impression that unchunked mode will cause each message to immediately flush to disk, which is not the case and I don't see an easy fix to make this the case (if it's even desirable in the general case).

PR #1266, opened 2024-11-07
C++: flush underlying writer after every chunk write

james-rms:
#### C++
- Added: added a `flush()` method to `IWritable` to flush any buffered bytes to the underlying writer. This is optional for any external implementations to implement, and defaults to a no-op.
- Changed: After writing a chunk, the McapWriter will call `flush()` on the underlying writer. This ensures that any completed chunk does not get held up in an internal buffer while the next chunk is being constructed.

### Docs

See new docstring on `flush()`.
### Description

Adds a virtual function to the `IWritable` interface to flush it, with a default no-op implementation. I don't really know the right policy for un-chunked files, but my current opinion is that we can leave that to the caller - they might call flush every few seconds, or every few messages.

I don't consider this a breaking change. This is almost certainly ABI-breaking, but we don't support that as a public interface, and I can't see a way that this change would cause existing code to fail to compile.

C++ writers using `FileWriter` might not see a chunk flushed to disk even after it has been fully finalized

</td><td>

C++ writers using `FileWriter` will see a  chunk flushed to disk as soon as it is finished.

</td></tr></table>

Fixes: #1254

### DIFF
--- a/cpp/mcap/include/mcap/writer.hpp
+++ b/cpp/mcap/include/mcap/writer.hpp
@@ -148,7 +148,7 @@ public:
    * completed chunk. Callers may also retain a reference to the writer and call flush() at their
    * own cadence. Defaults to a no-op.
    */
-  virtual void flush(){};
+  virtual void flush() {}
 
 protected:
   virtual void handleWrite(const std::byte* data, uint64_t size) = 0;
--- a/cpp/mcap/include/mcap/writer.hpp
+++ b/cpp/mcap/include/mcap/writer.hpp
@@ -144,9 +144,11 @@ public:
   void resetCrc();
 
   /**
-   * @brief flushes any buffered data to the output. Defaults to a no-op.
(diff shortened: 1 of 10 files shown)


james-rms:
cc. @jhurliman

cpp/mcap/include/mcap/writer.hpp
```diff
@@ -143,6 +143,13 @@ class MCAP_PUBLIC IWritable {
    */
   void resetCrc();
 
+  /**
+   * @brief flushes any buffered data to the output. This is called by McapWriter after every
+   * completed chunk. Callers may also retain a reference to the writer and call flush() at their
+   * own cadence. Defaults to a no-op.
+   */
+  virtual void flush(){};
```
jtbandes, line 151: nit
```suggestion
  virtual void flush(){}
```

jtbandes (approved):

PR #1273, opened 2024-11-17
Remove unnecessary usage of LFS in CI

amacneil:
GitHub charges for LFS bandwidth between GitHub and GitHub Actions. Therefore we should remove unnecessary usage of LFS in CI.

Blocked by https://github.com/foxglove/mcap/pull/1272 because yarn plugins were stored in LFS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          lfs: true
+          lfs: false
       - run: corepack enable
       - uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          lfs: true
+          lfs: false
```
amacneil, line 16: `false` is the default but I'm leaving it explicitly disabled to avoid someone unintentionally re-enabling

james-rms (approved):

PR #1274, opened 2024-11-18
rust: buffer chunks to allow writing without Seek

bennetthardwick:
### Changelog

- rust: add `use_buffered_chunks` option which, when set, will allow the mcap writer to write to a writer that doesn't support seeking

### Docs

None

### Description

This change adds in the concept of a `ChunkSink` used to write chunks to. It can either be `Direct`, where chunks are written directly to the output, or `Buffered`, where chunks are written to a temporary buffer before being written to the output.
Since chunks require seeking back to the start of the chunk to write the length, buffering chunks allows the mcap writer to write to streams that don't support seeking (like stdout, network responses, etc.), when used with something like `NoSeek` from binrw.

I spent a while trying to get this to work using generics so that using `NoSeek` wouldn't be a requirement (and the risk of a runtime error). Unfortunately the changes started to get quite complicated and would have resulted in signficant API changes due to how viral the `Write + Seek` bounds are currently. Ultimately I figured it would be simpler and more ergonomic to offer a runtime option.

~There is now a "linear" conformance writer to make sure that this doesn't regress~ edit: I realise now the conformance tests aren't working for writing so I've removed. There are "round trip" unit tests that cover this.

chunks would not be buffered, requiring an output writer that could `Seek`

</td><td>

using `use_buffered_chunks` allows you to use `NoSeek` with a writer that does not seek

</td></tr></table>


rust/src/chunk_sink.rs
```diff
@@ -0,0 +1,55 @@
...
+        /// The reusable buffer used by the [`ChunkSink`] when writing to [`ChunkSink::Buffered`]
+        buffer: Vec<u8>,
+    },
+}
+
+/// The writer used for writing chunks.
+///
+/// If chunks are buffered they will be written to an internal buffer, which can be flushed to the
+/// provided writer once the chunk is completed.
+pub enum ChunkSink<W> {
```
james-rms, line 20: would `pub(crate)` work?

rust/src/write.rs
```diff
@@ -121,6 +126,7 @@ impl Default for WriteOptions {
             profile: String::new(),
             chunk_size: Some(1024 * 768),
             use_chunks: true,
+            use_buffered_chunks: false,
```
james-rms, line 129: i might name this something more related to seeking like `disable_seeking`, `append_only` maybe?

bennetthardwick: Yeah I'm thinking that makes sense. That way it's more obvious that we're guaranteeing runtime no seeking.

rust/src/chunk_sink.rs
```diff
@@ -0,0 +1,55 @@
...
+    Direct(W),
+    Buffered(W, Cursor<Vec<u8>>),
+}
+
+impl<W: Seek> Seek for ChunkSink<W> {
+    fn seek(&mut self, pos: std::io::SeekFrom) -> std::io::Result<u64> {
+        match self {
+            Self::Direct(w) => w.seek(pos),
+            Self::Buffered(_, w) => w.seek(pos),
```
Muon, line 29: Factor out the matching to avoid repetition? Both `W` and `Cursor` are `Seek`.

bennetthardwick: How so? With `&dyn Seek`?

rust/src/write.rs
```diff
@@ -168,6 +174,17 @@ impl WriteOptions {
         Self { use_chunks, ..self }
     }
 
+    /// specifies whether to buffer chunks before writing them out
```
Muon, line 177: Missing capital/full stop.

rust/src/write.rs
```diff
@@ -207,8 +224,22 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
             }),
         )?;
 
+        // If both the `use_chunks` and `use_buffered_chunks` options are enabled set the chunk
+        // mode and pre-allocate the buffer. Checking both avoids allocating the temporary buffer
+        // if buffered chunks was enabled but chunks were not.
+        let chunk_mode = if opts.use_chunks && opts.use_buffered_chunks {
+            let buffer = opts
+                .chunk_size
+                .map(|size| Vec::with_capacity(size.try_into().unwrap_or(usize::MAX)))
```
Muon, line 233: Why `usize::MAX`? Even with optimistic allocation, that's just going to fail due to insufficient virtual memory address space.

bennetthardwick: Yeah wasn't super sure which way to go with this one. I figured semantically though if the consumer of the crate has chosen a chunk size larger than a `usize` then trying to allocate a buffer of that size would be somewhat expected behavior when buffering chunks. I'm not tied to this though so I'll happily return an error if you think that makes more sense.

rust/src/chunk_sink.rs
Muon: Yeah. If you're worried about the cost of indirect calls, you can see here they get compiled down to the same thing anyway due to inlining: https://godbolt.org/z/zo885v73d

rust/src/write.rs
james-rms: I think it makes sense to return a failed result here.

rust/src/chunk_sink.rs
bennetthardwick: Nice, might as well update!

rust/src/write.rs
```diff
@@ -168,6 +174,15 @@ impl WriteOptions {
         Self { use_chunks, ..self }
     }
 
+    /// Specifies whether the writer should seek or not.
+    ///
+    /// Enabling this option will allow you to use [`NoSeek`] on the destination writer to support
```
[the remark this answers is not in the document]
james-rms: Maybe also add a line mentioning why the writer seeks by default - so that it can write out chunk contents to disk without buffering in RAM.

james-rms (approved):

PR #1278, opened 2024-11-22
Add AV1 encoding to web demo if supported

jtbandes:
Like https://github.com/foxglove/mcap/pull/1275 but for AV1.

image


defunctzombie (approved):
The list grows

PR #1283, opened 2024-12-02
Use smaller testdata files

sofuture:
### Changelog
None

### Docs
None

### Description
Use smaller and simpler `demo.bag` and `demo.mcap` files.

Fixes: FG-9552

### DIFF
--- a/go/mcap/reader_test.go
+++ b/go/mcap/reader_test.go
@@ -445,7 +445,7 @@ func TestReaderCounting(t *testing.T) {
 				require.NoError(t, err)
 				c++
 			}
-			assert.Equal(t, 1606, c)
+			assert.Equal(t, 3, c)
 		})
 	}
 }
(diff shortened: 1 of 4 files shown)


jtbandes (approved):
Confirmed that `rosbag info` accepts this bag file:

```
version:     2.0
duration:    0.0s
start:       Dec 31 1969 16:00:00.00 (0.00)
end:         Dec 31 1969 16:00:00.00 (0.00)
size:        5.2 KB
messages:    3
compression: none [1/1 chunks]
types:       std_msgs/String [992ce8a1687cec8c8bd883ec73ca41d1]
topics:      /chatter       2 msgs    : std_msgs/String
             /diagnostics   1 msg     : std_msgs/String
```

Do we think there may be any code paths that were being tested by demo.bag which are no longer tested?

sofuture:
It's possible that we're losing coverage, but in that case we should solve with explicit test cases instead of an opaque binary artifact IMO.

PR #1284, opened 2024-12-03
rust: make interfaces Send

bennetthardwick:
### Changelog

rust: make sure structs are Send

### Docs

None

### Description

This PR makes sure that all MCAP readers and writers can be sent between threads. This is mainly just adding `Send` to the `Decompressor` trait, but it also adds compile time assertions to make sure we don't regress.

readers are not `Send`

</td><td>

readers are now `Send`

</td></tr></table>


james-rms (approved):
feel free to version bump in this PR if you want to release directly afterwards.

PR #1285, opened 2024-12-05
rust: write attachments in parts

bennetthardwick:
### Changelog

rust: support writing attachments in multiple chunks

### Docs

None

### Description

This change supports writing large attachments to MCAP files without the need to load them into memory.

Currently, in order to write an attachment, you need to load the entire file into memory, create an `Attachment` record, then use that with the `attach` method on the writer.
This means that it is not possible to attach files when the file exceeds the available memory on the computer.

This change adds The methods `start_attachment`, `put_attachment_bytes` and `finish_attachment` for interactively writing an attachment without needing to buffer the file in memory. It looks like this:

```rust
let writer = Writer::new(..);
let attachment_length = ..;
let header = AttachmentHeader { .. };

writer.start_attachment(attachment_length, header)?;
writer.put_attachment_bytes(..)?;
writer.put_attachment_bytes(..)?;
writer.put_attachment_bytes(..)?;
writer.finish_attachment()?;
```

I've updated the current `attach` method to use this approach and the roundtrip tests pass. I've also added some tests for using this API directly.

attachments must be loaded into memory to be written

</td><td>

attachments can be written in parts

</td></tr></table>

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -378,7 +378,7 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
 
     /// Start writing an attachment.
     ///
-    /// This is a low level API. It is recommended to use [`Self::attach`] instead.
+    /// This is a low level API. For small attachments, use [`Self::attach`].
     ///
     /// To start writing an attachment call this method with the [`AttachmentHeader`] as well as
     /// the length of the attachment in bytes. It is important this length is exact otherwise the
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -445,7 +445,7 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
 
     /// Write bytes to the current attachment.
     ///
-    /// This is a low level API. It is recommended to use [`Self::attach`] instead.
+    /// This is a low level API. For small attachments, use [`Self::attach`].
     ///
     /// Before calling this method call [`Self::start_attachment`].
     pub fn put_attachment_bytes(&mut self, bytes: &[u8]) -> McapResult<()> {
@@ -460,7 +460,7 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
 
     /// Finish the current attachment.
     ///
-    /// This is a low level API. It is recommended to use [`Self::attach`] instead.
+    /// This is a low level API. For small attachments, use [`Self::attach`].
     ///
     /// Before calling this method call [`Self::start_attachment`] and write bytes to the
     /// attachment using [`Self::put_attachment_bytes`].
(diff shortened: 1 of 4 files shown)


rust/src/write.rs
```diff
@@ -388,37 +376,93 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
+    ///
+    /// To start writing an attachment call this method with the [`AttachmentHeader`] as well as
+    /// the length of the attachment in bytes. It is important this length is exact otherwise the
+    /// writer will be left in an error state.
+    ///
+    /// This call should be followed by one or more calls to [`Self::put_attachment_bytes`].
+    ///
+    /// Once all attachment bytes have been written the attachment must be completed with a call to
+    /// [`Self::finish_attachment`]. Failing to finish the attachment will leave the write in an
+    /// error state.
```
james-rms, line 391: would be good to include an example which shows how to copy a file to an attachment correctly.

rust/src/write.rs
```diff
@@ -388,37 +376,93 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
 
-    pub fn attach(&mut self, attachment: &Attachment) -> McapResult<()> {
-        let header = records::AttachmentHeader {
-            log_time: attachment.log_time,
-            create_time: attachment.create_time,
-            name: attachment.name.clone(),
-            media_type: attachment.media_type.clone(),
+    /// Start writing an attachment.
+    ///
+    /// This is a low level API. It is recommended to use [`Self::attach`] instead.
```
james-rms, line 381:
```suggestion
    /// This is a low level API. For small attachments, use [`Self::attach`].
```

james-rms (approved):
LGTM.

rust/src/write.rs
```diff
@@ -388,37 +376,93 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
-        write_record(
-            w,
-            &Record::Attachment {
-                header,
-                data: Cow::Borrowed(&attachment.data),
-            },
-        )?;
+    /// Write bytes to the current attachment.
+    ///
+    /// This is a low level API. It is recommended to use [`Self::attach`] instead.
```
james-rms, line 417:
```suggestion
    /// This is a low level API. For small attachments, use [`Self::attach`].
```

rust/src/write.rs
```diff
@@ -388,37 +376,93 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
-            offset,
-            length,
+        writer.put_bytes(bytes)?;
+
+        Ok(())
+    }
+
+    /// Finish the current attachment.
+    ///
+    /// This is a low level API. It is recommended to use [`Self::attach`] instead.
```
james-rms, line 432:
```suggestion
    /// This is a low level API. For small attachments, use [`Self::attach`].
```

PR #1289, opened 2024-12-12
mcap recover: print status message to stderr instead of stdout

jtbandes:
### Changelog
Fixed a bug in `mcap recover` which would produce an invalid file when redirecting stdout to a file.

### Docs

None

### Description

Status message now goes to stderr instead of stdout.

Also fixed another nearby line which looked wrong, it was using Fprintln with `%w` but Fprintln doesn't accept format strings. https://pkg.go.dev/fmt#Fprintln


jtbandes:
Fixing CI in #1290

james-rms (approved):

PR #1294, opened 2024-12-15
rust: disable lz4 block checksum

bennetthardwick:
### Changelog

rust: disable lz4 block checksum

### Docs

None

### Description

Disabled the lz4 block checksum to mimic the 

</td><td>

</td></tr></table>


james-rms (approved):
Could use a comment regarding incompatibilities with other lz4 readers here.

PR #1297, opened 2024-12-19
rust: Channel and Schema structs have IDs

james-rms:
#### Breaking changes
- adds `id` members to the crate::Channel and crate::Schema structs.
- refactored and separated `Writer::add_channel` into two methods, `Writer::add_schema` and `Writer::add_channel`.

### Docs
In docstrings, please review if not clear.
### Description

When a message reader recieves schema and channel information, the schema and channel IDs in the MCAP are hidden from them.  This PR adds those so that readers can observe ID information.

Additionally, this PR updates the Writer interface in these ways:

1. The `add_channel(Channel)` method is replaced with `add_channel(schema_id, topic, message_encoding, metadata)`. This was done because the Channel struct now contains a channel ID, and to continue to use it would force the user to come up with their own channel ID before calling this function. 
2. An `add_schema(...)` method is added to register schemas ahead of channels. This is needed to allow the caller to register a new schema without coming up with an ID in advance.
3. The existing `write(Message)` method is kept, but instead of automatically assigning new IDs for the message channel and schema, it uses the IDs passed in. This means that when writing functions that transform MCAP data, the IDs from the input can be kept in the output.

</td><td>

</td></tr></table>

### DIFF
--- a/rust/examples/common/conformance_writer_spec.rs
+++ b/rust/examples/common/conformance_writer_spec.rs
@@ -20,28 +20,26 @@ pub struct Record {
 }
 
 impl Record {
-    pub fn get_field(self: &Self, name: &str) -> &Value {
-        return &self
+    pub fn get_field(&self, name: &str) -> &Value {
+        &self
             .fields
             .iter()
             .find(|f| f.0 == name)
             .unwrap_or_else(|| panic!("Invalid: {}", name))
-            .1;
+            .1
     }
 
-    pub fn get_field_data(self: &Self, name: &str) -> Vec<u8> {
-        let data: Vec<u8> = self
-            .get_field(name)
+    pub fn get_field_data(&self, name: &str) -> Vec<u8> {
+        self.get_field(name)
             .as_array()
             .unwrap_or_else(|| panic!("Invalid: {}", name))
-            .into_iter()
+            .iter()
             .filter_map(|v| v.as_u64())
             .filter_map(|n| u8::try_from(n).ok())
-            .collect();
-        return data;
+            .collect()
     }
 
-    pub fn get_field_meta(self: &Self, name: &str) -> BTreeMap<String, String> {
+    pub fn get_field_meta(&self, name: &str) -> BTreeMap<String, String> {
         let data = self
             .get_field(name)
             .as_object()
@@ -50,38 +48,34 @@ impl Record {
         for (key, value) in data.iter() {
             result.insert(key.to_string(), value.as_str().unwrap().to_string());
         }
-        return result;
+        result
     }
 
-    pub fn get_field_str(self: &Self, name: &str) -> &str {
-        return self
-            .get_field(name)
+    pub fn get_field_str(&self, name: &str) -> &str {
+        self.get_field(name)
             .as_str()
-            .unwrap_or_else(|| panic!("Invalid: {}", name));
+            .unwrap_or_else(|| panic!("Invalid: {}", name))
     }
 
-    pub fn get_field_u16(self: &Self, name: &str) -> u16 {
-        return self
-            .get_field(name)
+    pub fn get_field_u16(&self, name: &str) -> u16 {
+        self.get_field(name)
             .as_str()
             .and_then(|s| s.parse::<u16>().ok())
-            .unwrap_or_else(|| panic!("Invalid: {}", name));
+            .unwrap_or_else(|| panic!("Invalid: {}", name))
     }
 
-    pub fn get_field_u32(self: &Self, name: &str) -> u32 {
-        return self
-            .get_field(name)
+    pub fn get_field_u32(&self, name: &str) -> u32 {
+        self.get_field(name)
             .as_str()
             .and_then(|s| s.parse::<u32>().ok())
-            .unwrap_or_else(|| panic!("Invalid: {}", name));
+            .unwrap_or_else(|| panic!("Invalid: {}", name))
     }
 
-    pub fn get_field_u64(self: &Self, name: &str) -> u64 {
-        return self
-            .get_field(name)
+    pub fn get_field_u64(&self, name: &str) -> u64 {
+        self.get_field(name)
             .as_str()
             .and_then(|s| s.parse::<u64>().ok())
-            .unwrap_or_else(|| panic!("Invalid: {}", name));
+            .unwrap_or_else(|| panic!("Invalid: {}", name))
     }
 }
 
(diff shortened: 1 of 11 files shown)


rust/examples/common/conformance_writer_spec.rs
```diff
@@ -20,28 +20,26 @@ pub struct Record {
 }
 
 impl Record {
-    pub fn get_field(self: &Self, name: &str) -> &Value {
```
james-rms, line 23: all changes in these files are due to the `clippy --all-targets` change.

rust/examples/conformance_writer.rs
```diff
@@ -38,26 +40,17 @@ fn write_file(spec: &conformance_writer_spec::WriterSpec) {
...
-                    message_encoding: message_encoding.to_string(),
-                    metadata: std::collections::BTreeMap::new(),
-                };
-                writer
-                    .add_channel(&channel)
+                let returned_id = writer
+                    .add_channel(schema_id as u16, topic, message_encoding, &BTreeMap::new())
                     .expect("Couldn't write channel");
-                channels.insert(id, channel);
+                channel_ids.insert(returned_id, id);
```
james-rms, line 46: this is backwards: need to map from input to output.

rust/src/write.rs
```diff
@@ -233,87 +249,171 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
+    // its ID is returned.
+    pub fn add_schema(&mut self, name: &str, encoding: &str, data: &[u8]) -> McapResult<u16> {
+        if let Some(&id) = self.schemas.get_by_left(&SchemaContent {
+            name: name.into(),
+            encoding: encoding.into(),
+            data: data.into(),
+        }) {
+            return Ok(id);
         }
+        let next_schema_id = self.schemas.right_values().max().unwrap_or(&0) + 1;
```
james-rms, line 272: just take len(schemas) + 1

rust/src/write.rs
```diff
@@ -233,87 +249,171 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
+    ) -> McapResult<u16> {
+        if let Some(&id) = self.channels.get_by_left(&ChannelContent {
+            topic: Cow::Borrowed(topic),
+            schema_id,
+            message_encoding: Cow::Borrowed(message_encoding),
+            metadata: Cow::Borrowed(metadata),
+        }) {
+            return Ok(id);
+        }
+        let next_channel_id = self
```
james-rms, line 330: just take len(channels) + 1

james-rms:
cc. @mrkline

rust/src/write.rs
```diff
@@ -233,87 +249,171 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
-        let schema_id = match &chan.schema {
-            Some(s) => self.add_schema(s)?,
-            None => 0,
-        };
-
-        if let Some(id) = self.channels.get(chan) {
-            return Ok(*id);
+    // Adds a schema, returning its ID. If a schema with the same content has been added already,
+    // its ID is returned.
+    pub fn add_schema(&mut self, name: &str, encoding: &str, data: &[u8]) -> McapResult<u16> {
```
jtbandes, line 264: Thoughts on having a separate struct that's like "ChannelWithoutId"? Pros: avoid having to thread multiple params through different layers, avoid possible confusion between params of the same type.

rust/src/write.rs
```diff
@@ -175,20 +176,35 @@ impl WriteOptions {
...
+struct ChannelContent<'a> {
+    topic: Cow<'a, str>,
+    schema_id: u16,
+    message_encoding: Cow<'a, str>,
+    metadata: Cow<'a, BTreeMap<String, String>>,
+}
+
+#[derive(Hash, PartialEq, Eq)]
+struct SchemaContent<'a> {
+    name: Cow<'a, str>,
```
jtbandes, line 194: Could you ELI5 why we need Cow here?

jtbandes (approved):

rust/src/write.rs
james-rms: i'll add a comment. We use CoW because we want to compare equality between all of these values without neccessarily cloning or moving all of them into a struct to do that.

jtbandes: Is there some overhead inherent in the Cow struct itself? Should I think of as being basically a union between a couple different storage options/types? or a pointer to an allocation elsewhere?

james-rms: a Cow is a union between a borrowed and owned version of a given type - so for str, it's (String | &str), for bytes it's (&[u8], Vec<u8>). It implements Deref for those types, allowing you to use their comparison, hash implementations against eachother.

  In context, this struct lets me solves this problem:
```
struct left = {
  a: String::new("abc"),
  b: String::new("def")
}

struct right = {
  a: "abc",
  b: "def",
}
if a == b {
   println!("they are the same!")
}
```

rust/src/write.rs
```diff
@@ -233,87 +249,171 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
+    // its ID is returned.
+    pub fn add_schema(&mut self, name: &str, encoding: &str, data: &[u8]) -> McapResult<u16> {
+        if let Some(&id) = self.schemas.get_by_left(&SchemaContent {
+            name: name.into(),
+            encoding: encoding.into(),
+            data: data.into(),
+        }) {
+            return Ok(id);
         }
+        let next_schema_id = self.schemas.len() as u16 + 1;
```
gasmith, line 272: Pedantic: We should probably return an error if there are more than 2^16-1 schemas.

rust/src/write.rs
```diff
@@ -233,87 +249,171 @@ impl<'a, W: Write + Seek> Writer<'a, W> {
...
+    ) -> McapResult<u16> {
+        if let Some(&id) = self.channels.get_by_left(&ChannelContent {
+            topic: Cow::Borrowed(topic),
+            schema_id,
+            message_encoding: Cow::Borrowed(message_encoding),
+            metadata: Cow::Borrowed(metadata),
+        }) {
+            return Ok(id);
+        }
+        let next_channel_id = self.channels.len() as u16;
```
gasmith, line 330: Pedantic: We should probably return an error if there are more than 2^16 channels.

gasmith (approved):

rust/src/write.rs
james-rms: The con here is that the struct would need to be defined with explicit generic lifetimes per borrowed argument. In general the borrow checker is more precise around function arguments than it is around struct members, and if such a struct is not defined in a very particular way it may constrain the API in a way we don't expect.

Aside: Iave a personal bias against deploying this pattern too early. IMO a struct should justify its own existence. If it only exists to pack arguments for one particular function, I don't think it's useful.

PR #1299, opened 2024-12-19
rust: read: handle padded compressed chunks properly

james-rms:
### Changelog
- rust: fixed case where a compressed chunk with padding bytes added by the compressor at the end would break decompression.

### Docs

None.

### Description

When MCAPs are written, the compressor may insert padding bytes at the end of a compressed block after all useful data bytes are written. This means that when the decompressor is reading messages out, it will have decompressed all message bytes out of a compressed chunk _before_ it reaches the end of the compressed data. Right now, the rust MCAP library will see that there is more unused data in the compressed chunk, try to decompress another message, but get no output from of the decompressor when it tries.  Right now this causes the reader to loop infinitely. This PR adds a file to LFS which triggers this condition and fixes it.

</td><td>

</td></tr></table>

### DIFF
--- a/rust/src/sans_io/read.rs
+++ b/rust/src/sans_io/read.rs
@@ -595,8 +595,14 @@ impl LinearReader {
                             if state.uncompressed_remaining == 0
                                 && self.decompressed_content.len() == 0
                             {
+                                // We've consumed all compressed data. It's possible for there to
+                                // still be data left in the chunk that has not yet been read into
+                                // `self.file_data`. This can happen when a compressor adds extra
+                                // bytes after its last frame. We need to treat this as "padding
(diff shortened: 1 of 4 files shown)


rust/src/sans_io/read.rs
```diff
@@ -1101,4 +1102,31 @@ mod tests {
         }
         Ok(())
     }
+
+    #[test]
+    fn test_decompression_does_not_fail() {
+        let mut f = std::fs::File::open("tests/data/break_zstd_decompression.mcap")
```
gasmith, line 1108: Nit: Could we rename this test and/or mcap file to more clearly indicate the corner case that it's tickling?

gasmith (approved):
LG

PR #1310, opened 2025-01-14
@mcap/core: Bump to v2.1.6

achim-k:


defunctzombie (approved):
We want this to be a patch release or a minor? I have forgotten what has changed.

achim-k:
> We want this to be a patch release or a minor? I have forgotten what has changed.

https://github.com/foxglove/mcap/pull/1304 fixes that the indexed reader chokes on empty chunks without message indexes. I think a patch release is fine for that?

PR #1313, opened 2025-01-17
Enable most writer conformance tests for Rust

Muon:
### Changelog
The DataEnd CRC calculation and missing writer options have been added to allow the Rust writer to pass the conformance tests.

### Docs

https://linear.app/foxglove/issue/FG-9567/rust-mcap-writer-doesnt-pass-or-run-conformance-tests

### Description

Previously, the Rust writer could not pass the conformance tests because it was not possible to control which summary record types would be included and because it would not compute the CRC for the DataEnd record.

Now, all current writer conformance tests pass, except the ones using the "pad" option. This brings the Rust crate up to parity with the C++ library (and possibly others).

No writer conformance tests run or pass

</td><td>

All writer conformance tests except those with "pad" pass

</td></tr></table>

Fixes: FG-9567


rust/src/write.rs
```diff
@@ -249,14 +364,17 @@ impl<W: Write + Seek> Writer<W> {
...
 
         Ok(Self {
-            writer: Some(WriteMode::Raw(writer, chunk_mode)),
+            writer: Some(WriteMode::Raw(writer)),
             options: opts,
+            chunk_mode,
             schemas: Default::default(),
             channels: Default::default(),
-            next_channel_id: 0,
+            next_channel_id: 1,
```
james-rms, line 372: why is this change neccessary?

rust/src/write.rs
```diff
@@ -672,7 +793,7 @@ impl<W: Write + Seek> Writer<W> {
     const WHERE_WRITER: &'static str = "Trying to write a record on a finished MCAP";
 
     /// Starts a new chunk if we haven't done so already.
-    fn chunkin_time(&mut self) -> McapResult<&mut ChunkWriter<W>> {
+    fn start_chunk(&mut self) -> McapResult<&mut ChunkWriter<W>> {
```
james-rms, line 796: 🥹

rust/src/write.rs
```diff
@@ -175,6 +205,87 @@ impl WriteOptions {
         self
     }
 
+    /// Specifies whether the writer should output statistics automatically.
```
james-rms, line 208: I don't understand the use of "automatically" - there's not a manual alternative. Maybe:
```suggestion
    /// Specifies whether the writer should output a statistics record in the summary section.
```

rust/src/write.rs
```diff
@@ -175,6 +205,87 @@ impl WriteOptions {
...
+    ///
+    /// By default the writer will output a statistics record to the MCAP file when it is finished.
+    /// Set this to `false` to disable this behavior.
+    pub fn output_statistics(mut self, output_statistics: bool) -> Self {
+        self.output_statistics = output_statistics;
+        self
+    }
+
+    /// Specifies whether the writer should output a summary record automatically.
```
james-rms, line 217: There is no summary record, there is a summary section (the records between the data end record and the footer record) which may be empty. https://mcap.dev/spec#summary-section

rust/src/write.rs
```diff
@@ -175,6 +205,87 @@ impl WriteOptions {
...
+    ///
+    /// By default the writer will output a summary record to the MCAP file when it is finished.
+    /// Set this to `false` to disable this behavior.
+    pub fn output_summary(mut self, output_summary: bool) -> Self {
+        self.output_summary = output_summary;
+        self
+    }
+
+    /// Specifies whether the writer should output summary offsets automatically.
```
james-rms, line 226: The use of "automatically" here and throughout confuses me. the caller is not choosing between automatic and manual, they're choosing between present and absent.

james-rms (commented):
The new write option docstrings need updates, but the other changes look good.

rust/src/write.rs
Muon: Good question. I was going to say something about the other writers starting from 1, but they don't. Actually, it's because the Rust interface tries to be clever and doesn't let you pick your own channel IDs, but the tests start from 1. If the tests started from a different number things would break.

Changing `add_channel()` to let you pick your own would be a breaking change. Or we could add a new method?

rust/src/write.rs
Muon: I am a killjoy who prioritizes code readability.

Muon:
I'll fix up the docstrings.

rust/src/write.rs
james-rms: You can choose your own channel IDs if you call write(Message{}) and set the channel ID in the Message.channel.id field.

rust/src/write.rs
bennetthardwick: I felt the itch to remove but ultimately felt the love when I saw this last time 🥹. Had to be done though.

rust/src/write.rs
Muon: The problem is not writing the Message record, but the Channel record. We can't currently choose the ID when creating a Channel.

james-rms: you don't need to call `add_channel` before calling `Write(Message{channel: Some(Channel{...}))`. This will write the channel record with the ID you specify, then the message record.

james-rms: You can avoid writing a channel record when you see a `Channel` object in the spec JSON, instead storing it in a map somewhere. Then you call Write(Message) when the message comes up in the spec.

Muon: Right, but if I wrote a test now that includes a channel with no messages, or that creates multiple channels all at once and then writes some messages to each of them, then that test would fail under your current proposal. The writer conformance test harness is checking whether all the records come out in the same order and with the same data.

rust/src/write.rs
Muon: Fixed in updated docs.

rust/src/write.rs
Muon: Fixed in updated docs.

rust/src/write.rs
Muon: Fixed in updated docs.

james-rms (approved):
I'm happy with this. I'll make a ticket to add an explicit `create_channel_with_id` method.

PR #1314, opened 2025-01-20
Add into_inner() on Rust writer

Muon:
### Changelog
Add an `into_inner()` method on the Rust writer to get back the underlying stream.

### Docs

https://linear.app/foxglove/issue/FG-9969/[rust-sdk]-cant-close-mcap-writer

### Description

Some I/O errors only show up when the output file is closed. This change allows users to check for such errors.

Fixes: FG-9969

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -10,6 +10,8 @@ use std::{
 use bimap::BiHashMap;
 use binrw::prelude::*;
 use byteorder::{WriteBytesExt, LE};
+#[cfg(feature = "zstd")]
+use zstd::stream::{raw as zraw, zio};
 
 use crate::{
     chunk_sink::{ChunkMode, ChunkSink},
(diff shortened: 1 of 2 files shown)


rust/src/write.rs
```diff
@@ -1053,14 +1051,38 @@ impl<W: Write + Seek> Writer<W> {
...
+        self.is_finished = true;
+        match self.writer.take().expect(Self::WHERE_WRITER) {
+            WriteMode::Raw(w) => w.finalize().0,
+            WriteMode::Attachment(w) => w.writer.finalize().0.finalize().0,
+            WriteMode::Chunk(w) => {
+                w.compressor
+                    .finalize()
+                    .0
+                    .finish()
+                    .expect("compression error")
```
bennetthardwick, line 1079: Did you mean to have this here instead of returning a result?

Muon: Yes. It's because of the zstd compressor. I'll add a comment to clarify.

james-rms: what if we swallowed the error instead? The caller is already expecting to not get a complete MCAP.

Muon: I've rejigged it to avoid the unwrap entirely by using the interface in zio::Writer instead.

Muon: @james-rms swallowing wasn't an option because the wrapper interface wouldn't even let you get the W back unless finishing succeeded.

james-rms (approved):

PR #1300, opened 2024-12-23
fix: add typings to package output

dnikitop:
### Description

Getting these errors in when importing

```
node_modules/@mcap/support/dist/esm/src/protobufDescriptors.d.ts:1:22 - error TS6053: File '/home/dimitri/src/rosbags2js/node_modules/@mcap/support/typings/protobufjs.d.ts' not found.

1 /// <reference path="../../../typings/protobufjs.d.ts" />
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@mcap/support/dist/esm/src/protobufDescriptors.d.ts:3:61 - error TS2339: Property 'toDescriptor' does not exist on type 'Root'.

3 export type ProtobufDescriptor = ReturnType<protobufjs.Root["toDescriptor"]>;                                       
```

PR adds `typings/protobufjs.d.ts` to package

<table><tr><th>Before</th><th>After</th></tr><tr><td>

</td><td>

</td></tr></table>

PR #1318, opened 2025-01-22
@mcap/support: remove .d.ts file in favor of module augmentation in source file

jtbandes:
### Changelog
@mcap/support: Fixed a TS error from an import that did not work in the published package

### Docs

None

### Description

Supersedes / closes #1300 
See also https://github.com/protobufjs/protobuf.js/issues/1499

This area of TypeScript arcana is confusing, but I don't think there is anything about [module augmentations](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) that requires them to be in a `.d.ts` file.

The way it's currently done was generally working, except as pointed out in #1300 we didn't include the `typings` dir in the published package.

I believe we didn't run into this problem in our own code because we usually turn on `"skipLibCheck": true`.


dnikitop (approved):
LGTM. Thanks!

defunctzombie (approved):
I would not bump the package version in the same PR - a habit I try to avoid should we decide on some other change being part of a new release

jtbandes:
> should we decide on some other change being part of a new release

Even if I intend to release immediately after merging the PR? I find this slows down the ability to release since you have to wait for review again 🤷

typescript/support/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mcap/support",
-  "version": "1.0.2",
+  "version": "1.0.3",
```
jtbandes, line 3:
```suggestion
  "version": "1.0.2",
```

PR #1320, opened 2025-01-24
rust reader: unexpected EOF in chunk is EOF

james-rms:
### Description

</td><td>

</td></tr></table>


bennetthardwick (approved):

jtbandes:
What's this for? Looks like maybe just tweaking which error message occurs in a given situation? Can it be unit tested?

PR #1321, opened 2025-01-26
BUGFIX: Add support for nested definitions in Python mcap-protobuf-support

methylDragon:
### Changelog

Bugfix: Support nested message definitions in mcap-protobuf-support

### Description

Consider the following Protobuf definition:

```
message NestedTypeMessage {
  message NestedType {
      string data = 1;
    }

  NestedType nested = 1;
}
```

Currently the decoder's use of the `protobuf.GetMessageClassesForFiles()` misses out the inner `NestedType` definition, causing decoding failures from missing nested descriptors, even though those descriptors would have been populated as part of the encoded file descriptors in an MCAP file.

The error that is raised is something like: `FileDescriptorSet for type {schema.name} is missing that schema`

This issue persists to an arbitrary level of nesting.

This PR fixes that issue (including for deeply nested definitions) and adds a unit test to prevent regression.

### PS:
How soon after merging this will it make it to PyPI?


python/mcap-protobuf-support/mcap_protobuf/decoder.py
```diff
@@ -59,6 +59,26 @@ def _add(fd: FileDescriptorProto):
...
+            # We must traverse the top-level message definitions to recursively find and append any
+            # nested message definitions.
+            nested_messages = {}
+            for message_class in messages.values():
+                nested_descriptions = list(
+                    message_class.DESCRIPTOR.nested_types_by_name.values()
+                )
+                while nested_descriptions:
+                    nested_desc = nested_descriptions.pop()
+                    nested_messages[nested_desc.full_name] = GetMessageClass(
```
james-rms, line 73: why does this not mutate `messages` directly?

james-rms (approved):
Looks good to me, thanks for the fix!

python/mcap-protobuf-support/mcap_protobuf/decoder.py
methylDragon: We will otherwise be mutating it while we are iterating through the dictionary.

PR #1328, opened 2025-02-05
writer: add options to switch off CRC calculation

james-rms:
### Changelog
- Added: rust writeOptions now include `calculate_*_crc[s]` options to enable or disable CRC calculation.
### Docs

Check docstrings.
### Description

CRC calculation can have a meaningful impact on CPU load and throughput, and is not appropriate for some use-cases. 

</td><td>

</td></tr></table>

### DIFF
--- a/rust/src/io_utils.rs
+++ b/rust/src/io_utils.rs
@@ -4,20 +4,24 @@ use crc32fast::Hasher;
 
 pub struct CountingCrcWriter<W> {
     inner: W,
-    hasher: Hasher,
+    hasher: Option<Hasher>,
     count: u64,
 }
 
 impl<W> CountingCrcWriter<W> {
-    pub fn new(inner: W) -> Self {
+    pub fn new(inner: W, calculate_crc: bool) -> Self {
         Self {
             inner,
-            hasher: Hasher::new(),
+            hasher: if calculate_crc {
+                Some(Hasher::new())
+            } else {
+                None
+            },
             count: 0,
         }
     }
 
-    pub fn with_hasher(inner: W, hasher: Hasher) -> Self {
+    pub fn with_hasher(inner: W, hasher: Option<Hasher>) -> Self {
         Self {
             inner,
             hasher,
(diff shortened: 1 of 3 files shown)


rust/src/io_utils.rs
```diff
@@ -4,20 +4,24 @@ use crc32fast::Hasher;
 
 pub struct CountingCrcWriter<W> {
```
jtbandes, line 5: rather than making this optional, would it make sense to dynamically use or not use this CountingCrcWriter wrapper?

rust/src/write.rs
```diff
@@ -290,6 +298,30 @@ impl WriteOptions {
     pub fn create<W: Write + Seek>(self, w: W) -> McapResult<Writer<W>> {
         Writer::with_options(w, self)
     }
+
+    /// Specifies whether to calculate and write CRCs for chunk records. This is on by default.
+    pub fn calculate_chunk_crcs(mut self, calculate_chunk_crcs: bool) -> Self {
```
jtbandes, line 303: wondering if there are any commonly-used ways of reducing boilerplate, like a macro to make a setter for the builder pattern?

rust/src/io_utils.rs
```diff
@@ -34,20 +38,25 @@ impl<W> CountingCrcWriter<W> {
     }
 
     /// Consumes the reader and returns the inner writer and the checksum
-    pub fn finalize(self) -> (W, Hasher) {
+    pub fn finalize(self) -> (W, Option<Hasher>) {
         (self.inner, self.hasher)
```
jtbandes, line 42: why do we return `hasher` rather than `hasher.finalize()`? And why do we need to return `inner` at all?

jtbandes (approved):

rust/src/io_utils.rs
james-rms: There is a case here where we re-use the hasher in another CountingCrcWriter. We need inner because generally we are not computing the CRC for an entire file that can be dropped immediately, we usually need to write something else after we use the CRC (like the end magic, or the next chunk)

Muon: We return the hasher so we can use its methods. In particular, we need `combine()`.

We need to still be able to write things to the underlying stream after computing a CRC.  (Such as the CRC itself.)

rust/src/io_utils.rs
```diff
@@ -4,20 +4,24 @@ use crc32fast::Hasher;
...
-    pub fn new(inner: W) -> Self {
+    pub fn new(inner: W, calculate_crc: bool) -> Self {
         Self {
             inner,
-            hasher: Hasher::new(),
+            hasher: if calculate_crc {
+                Some(Hasher::new())
+            } else {
+                None
+            },
```
bennetthardwick, line 19: Just a nit, but you could have a `NoopHasher` instead of an if statement so that you don't need to branch on every write.

bennetthardwick (approved):
Makes sense to me. Tested locally and it does indeed not crc.

rust/src/write.rs
james-rms: we could certainly add a macro - I can take a look at this in a later PR.

PR #1331, opened 2025-02-08
cli: Support multiple inputs in 'cat'

wkalt:
This adds support to the `mcap cat` command for multiple input files. Multiple files can be directly supplied, like

    mcap cat file1.mcap file2.mcap

or, the list of files may be auto-populated by your shell, enabling you to type

    mcap cat **/*.mcap

### DIFF
--- a/go/cli/mcap/cmd/cat.go
+++ b/go/cli/mcap/cmd/cat.go
@@ -318,13 +318,15 @@ var catCmd = &cobra.Command{
 		if err != nil {
 			die("Failed to stat() stdin: %s", err)
 		}
-		// read stdin if no filename has been provided and data is available on stdin.
-		readingStdin := (stat.Mode()&os.ModeCharDevice == 0 && len(args) == 0)
-		// stdin is a special case, since we can't seek
 
 		output := bufio.NewWriter(os.Stdout)
 		defer output.Flush()
 
+		// read stdin if no filename has been provided and data is available on
+		// stdin.
+		readingStdin := (stat.Mode()&os.ModeCharDevice == 0 && len(args) == 0)
+
+		// stdin is a special case, since we can't seek
 		if readingStdin {
 			reader, err := mcap.NewReader(os.Stdin)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


go/cli/mcap/cmd/cat.go
```diff
@@ -318,13 +318,15 @@ var catCmd = &cobra.Command{
...
-		// stdin is a special case, since we can't seek
 
 		output := bufio.NewWriter(os.Stdout)
 		defer output.Flush()
 
+		// read stdin if no filename has been provided and data is available on
+		// stdin.
+		readingStdin := (stat.Mode()&os.ModeCharDevice == 0 && len(args) == 0)
+
+		// stdin is a special case, since we can't seek
```
wkalt, line 329: I rearranged this comment and moved the stdin-related stuff closer together.

james-rms:
This seems ok, but I think it deserves a CLI doc change to describe what the behavior is. Before looking at the code and just looking at the PR title, I assumed you were extending `cat` to stream-merge input files and produce a time-ordered output stream, but that's not what's happening here.

Related - do you think it'd be worth supporting a `--mcap` output format so that there's a CLI tool which can logically concatenate the messages in several MCAP files into a single output MCAP, without merging?

wkalt:
@james-rms thanks, just expanded/clarified the doc.

I don't have an opinion on that question really. I think where concatenation of mcap files could be interesting is if you could do it in a way that allowed you to avoid decompressing any chunks, but I think to support that there would need to be some extensions of the format, or else you'd only be able to do it in some narrow circumstances where you control the data (e.g single-topic queries in the DP).

For me, the functionality in this patch is useful because it lets me turn the mcap files on my filesystem into a kind of JSON database for importing random message types into other systems.

So I can do
```
mcap cat ~/data/**/*.mcap --topics /diagnostics --json | ingest_to_lance.py
```

or things like that. Don't even need to know where the file is :)

defunctzombie:
> I assumed you were extending `cat` to stream-merge input files and produce a time-ordered output stream, but that's not what's happening here.

Why is that not what happens here? Does the existing `mcap cat` command produce data in time-order? If it does then it seems strange to me that this would not? What's the workflow where you want this but don't care about time order?

james-rms:
The existing `cat` produces a text line for every message in the file in whatever order they appear in the file.

james-rms (approved):

PR #1333, opened 2025-02-09
Speed up indexed iterator on negative search by topic

wkalt:
Prior to this commit, the indexed iterator did not rule out chunks based on topic specification, which meant it still opened chunks when no topics would match. With this change the tool does less IO when there is no matching data.

### DIFF
--- a/go/mcap/indexed_message_iterator.go
+++ b/go/mcap/indexed_message_iterator.go
@@ -165,7 +165,12 @@ func (it *indexedMessageIterator) parseSummarySection() error {
 			}
 			// if the chunk overlaps with the requested parameters, load it
 			if (it.end == 0 && it.start == 0) || (idx.MessageStartTime < it.end && idx.MessageEndTime >= it.start) {
-				it.chunkIndexes = append(it.chunkIndexes, idx)
+				for chanID := range idx.MessageIndexOffsets {
+					if it.channels.Get(chanID) != nil {
+						it.chunkIndexes = append(it.chunkIndexes, idx)
+						break
+					}
+				}
 			}
 		case TokenStatistics:
 			stats, err := ParseStatistics(record)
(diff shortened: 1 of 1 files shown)


wkalt:
timings:
```
[~/repos/mcap/go/cli/mcap] (main) $ time mcap cat ~/data/bags/cal_loop.mcap --topics /foobar

real    0m0.969s
user    0m0.911s
sys     0m0.163s
[~/repos/mcap/go/cli/mcap] (main) $ git checkout -
Switched to branch 'indexed-iterator-negative-lookup'
[~/repos/mcap/go/cli/mcap] (indexed-iterator-negative-lookup) $ go install
[~/repos/mcap/go/cli/mcap] (indexed-iterator-negative-lookup) $ time mcap cat ~/data/bags/cal_loop.mcap --topics /foobar

real    0m0.157s
user    0m0.227s
sys     0m0.024s

```

james-rms (changes requested):
Right now this reader works with files with chunk indexes but no message indexes, which is a case I'd like to continue to support. You could check if the chunk index has any message index offsets before ruling topics out based on their absence.

wkalt:
@james-rms thanks, updated. No speed change from before but here's a more extreme test example:
```
[~/repos/mcap/go/cli/mcap] (main) $ time ./mcap cat ~/data/Run10.bag.mcap --topics /foobar

real    0m5.548s
user    0m5.185s
sys     0m0.394s
[~/repos/mcap/go/cli/mcap] (main) $ git checkout -
Switched to branch 'indexed-iterator-negative-lookup'
Your branch is up to date with 'wkalt/indexed-iterator-negative-lookup'.
[~/repos/mcap/go/cli/mcap] (indexed-iterator-negative-lookup) $ go build
[~/repos/mcap/go/cli/mcap] (indexed-iterator-negative-lookup) $ time ./mcap cat ~/data/Run10.bag.mcap --topics /foobar

real    0m0.009s
user    0m0.006s
sys     0m0.006s

```

I think the negative case is quicker because the other test was on an old laptop.

james-rms (approved):

james-rms (approved):

PR #1338, opened 2025-02-25
rust: add sans-io indexed reader 

james-rms:
### Changelog
- Added: `mcap::sans_io::SummaryReader` and `mcap::sans_io::IndexedReader` structs, which can be used to a) parse out summary information and b) read messages efficiently using that summary information.

### Docs

See `cargo doc` for usage info.

### Description

Adds a sans-io indexed reading implementation, including conformance tests and benchmarks.

### Benchmarks

The indexed reader is suspiciously faster than MessageStream by quite a bit. I suspect the major difference is that the indexed reader allows the caller to re-use the same buffer for every message yielded, rather than allocating a new one.

Raw `cargo bench` outputs (with changes removed because they compare to the previous run, which is confusing)
```
mcap_read_linear/MessageStream_1M_uncompressed
                        time:   [165.82 ms 167.64 ms 169.53 ms]
                        thrpt:  [5.8986 Melem/s 5.9653 Melem/s 6.0306 Melem/s]
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
mcap_read_linear/MessageStream_1M_lz4
                        time:   [203.02 ms 204.87 ms 207.63 ms]
                        thrpt:  [4.8163 Melem/s 4.8811 Melem/s 4.9255 Melem/s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
mcap_read_linear/MessageStream_1M_zstd
                        time:   [213.31 ms 213.74 ms 214.26 ms]
                        thrpt:  [4.6672 Melem/s 4.6787 Melem/s 4.6881 Melem/s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

mcap_read_indexed/IndexedReader_1M_uncompressed
                        time:   [42.935 ms 43.215 ms 43.655 ms]
                        thrpt:  [22.907 Melem/s 23.140 Melem/s 23.291 Melem/s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
mcap_read_indexed/IndexedReader_1M_zstd
                        time:   [65.055 ms 65.616 ms 66.254 ms]
                        thrpt:  [15.093 Melem/s 15.240 Melem/s 15.372 Melem/s]
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) low mild
  1 (10.00%) high severe
mcap_read_indexed/IndexedReader_1M_lz4
                        time:   [60.758 ms 60.977 ms 61.139 ms]
                        thrpt:  [16.356 Melem/s 16.400 Melem/s 16.459 Melem/s]

### DIFF
--- a/rust/src/read.rs
+++ b/rust/src/read.rs
@@ -502,7 +502,9 @@ impl Iterator for MessageStream<'_> {
     }
 }
 
-const FOOTER_LEN: usize = 20; // 20 bytes + 8 byte len + 1 byte opcode
+const FOOTER_LEN: usize = 8 // summary start
+ + 8 // summary offset start
+ + 4; // summary section CRC
 
 /// Read the MCAP footer.
 ///
@@ -510,7 +512,12 @@ const FOOTER_LEN: usize = 20; // 20 bytes + 8 byte len + 1 byte opcode
 /// then index into the rest of the file with
 /// [`Summary::stream_chunk`], [`attachment`], [`metadata`], etc.
 pub fn footer(mcap: &[u8]) -> McapResult<records::Footer> {
-    if mcap.len() < MAGIC.len() * 2 + (9 + 1 + FOOTER_LEN) {
+    let footer_record_len = 1 // opcode
+     + 8 // record length
+     + FOOTER_LEN;
+    // an MCAP must be at least large enough to accomodate a header magic, a footer record and a
+    // footer magic.
+    if mcap.len() < (MAGIC.len() + (footer_record_len) + MAGIC.len()) {
         return Err(McapError::UnexpectedEof);
     }
 
--- a/rust/src/read.rs
+++ b/rust/src/read.rs
@@ -505,6 +505,9 @@ impl Iterator for MessageStream<'_> {
 const FOOTER_LEN: usize = 8 // summary start
  + 8 // summary offset start
  + 4; // summary section CRC
+const FOOTER_RECORD_LEN: usize = 1 // opcode
+     + 8 // record length
+     + FOOTER_LEN;
 
 /// Read the MCAP footer.
 ///
@@ -512,12 +515,9 @@ const FOOTER_LEN: usize = 8 // summary start
 /// then index into the rest of the file with
 /// [`Summary::stream_chunk`], [`attachment`], [`metadata`], etc.
 pub fn footer(mcap: &[u8]) -> McapResult<records::Footer> {
-    let footer_record_len = 1 // opcode
-     + 8 // record length
-     + FOOTER_LEN;
     // an MCAP must be at least large enough to accomodate a header magic, a footer record and a
     // footer magic.
-    if mcap.len() < (MAGIC.len() + (footer_record_len) + MAGIC.len()) {
+    if mcap.len() < (MAGIC.len() + FOOTER_RECORD_LEN + MAGIC.len()) {
         return Err(McapError::UnexpectedEof);
     }
 
--- a/rust/src/read.rs
+++ b/rust/src/read.rs
@@ -563,11 +563,11 @@ impl Summary {
         let mut summary_reader = SummaryReader::new();
         while let Some(event) = summary_reader.next_event() {
             match event? {
-                SummaryReadEvent::Read(n) => {
+                SummaryReadEvent::ReadRequest(n) => {
                     let read = cursor.read(summary_reader.insert(n))?;
                     summary_reader.notify_read(read);
                 }
-                SummaryReadEvent::Seek(to) => {
+                SummaryReadEvent::SeekRequest(to) => {
                     let pos = cursor.seek(to)?;
                     summary_reader.notify_seeked(pos);
                 }
--- a/rust/src/read.rs
+++ b/rust/src/read.rs
@@ -16,11 +16,13 @@ use binrw::prelude::*;
 use byteorder::{ReadBytesExt, LE};
 use crc32fast::hash as crc32;
 use enumset::{enum_set, EnumSet, EnumSetType};
-use log::*;
 
 use crate::{
-    records::{self, op, Record},
-    sans_io::{LinearReadEvent, LinearReader as SansIoReader, LinearReaderOptions},
+    records::{self, op, Footer, Record},
+    sans_io::{
+        LinearReadEvent, LinearReader as SansIoReader, LinearReaderOptions, SummaryReadEvent,
+        SummaryReader,
+    },
     Attachment, Channel, McapError, McapResult, Message, Schema, MAGIC,
 };
 
@@ -287,13 +289,17 @@ impl<'a> Iterator for ChunkFlattener<'a> {
 
 /// Parses schemas and channels and wires them together
 #[derive(Debug, Default)]
-struct ChannelAccumulator<'a> {
-    schemas: HashMap<u16, Arc<Schema<'a>>>,
-    channels: HashMap<u16, Arc<Channel<'a>>>,
+pub(crate) struct ChannelAccumulator<'a> {
+    pub(crate) schemas: HashMap<u16, Arc<Schema<'a>>>,
+    pub(crate) channels: HashMap<u16, Arc<Channel<'a>>>,
 }
 
 impl<'a> ChannelAccumulator<'a> {
-    fn add_schema(&mut self, header: records::SchemaHeader, data: Cow<'a, [u8]>) -> McapResult<()> {
+    pub(crate) fn add_schema(
+        &mut self,
+        header: records::SchemaHeader,
+        data: Cow<'a, [u8]>,
+    ) -> McapResult<()> {
         if header.id == 0 {
             return Err(McapError::InvalidSchemaId);
         }
@@ -315,7 +321,7 @@ impl<'a> ChannelAccumulator<'a> {
         Ok(())
     }
 
-    fn add_channel(&mut self, chan: records::Channel) -> McapResult<()> {
+    pub(crate) fn add_channel(&mut self, chan: records::Channel) -> McapResult<()> {
         // The schema ID can be 0 for "no schema",
         // Or must reference some previously-read schema.
         let schema = if chan.schema_id == 0 {
@@ -346,7 +352,7 @@ impl<'a> ChannelAccumulator<'a> {
         Ok(())
     }
 
-    fn get(&self, chan_id: u16) -> Option<Arc<Channel<'a>>> {
+    pub(crate) fn get(&self, chan_id: u16) -> Option<Arc<Channel<'a>>> {
         self.channels.get(&chan_id).cloned()
     }
 }
@@ -496,7 +502,7 @@ impl Iterator for MessageStream<'_> {
     }
 }
 
-const FOOTER_LEN: usize = 20 + 8 + 1; // 20 bytes + 8 byte len + 1 byte opcode
+const FOOTER_LEN: usize = 20; // 20 bytes + 8 byte len + 1 byte opcode
 
 /// Read the MCAP footer.
 ///
@@ -504,7 +510,7 @@ const FOOTER_LEN: usize = 20 + 8 + 1; // 20 bytes + 8 byte len + 1 byte opcode
 /// then index into the rest of the file with
 /// [`Summary::stream_chunk`], [`attachment`], [`metadata`], etc.
 pub fn footer(mcap: &[u8]) -> McapResult<records::Footer> {
-    if mcap.len() < MAGIC.len() * 2 + FOOTER_LEN {
+    if mcap.len() < MAGIC.len() * 2 + (9 + 1 + FOOTER_LEN) {
         return Err(McapError::UnexpectedEof);
     }
 
@@ -513,27 +519,25 @@ pub fn footer(mcap: &[u8]) -> McapResult<records::Footer> {
     }
 
     let footer_buf = &mcap[mcap.len() - MAGIC.len() - FOOTER_LEN..];
+    let mut cursor = std::io::Cursor::new(footer_buf);
 
-    match LinearReader::sans_magic(footer_buf).next() {
-        Some(Ok(Record::Footer(f))) => Ok(f),
-        _ => Err(McapError::BadFooter),
-    }
+    Ok(Footer::read_le(&mut cursor)?)
 }
 
 /// Indexes of an MCAP file parsed from its (optional) summary section
 #[derive(Default, Eq, PartialEq)]
-pub struct Summary<'a> {
+pub struct Summary {
     pub stats: Option<records::Statistics>,
     /// Maps channel IDs to their channel
-    pub channels: HashMap<u16, Arc<Channel<'a>>>,
+    pub channels: HashMap<u16, Arc<Channel<'static>>>,
     /// Maps schema IDs to their schema
-    pub schemas: HashMap<u16, Arc<Schema<'a>>>,
+    pub schemas: HashMap<u16, Arc<Schema<'static>>>,
     pub chunk_indexes: Vec<records::ChunkIndex>,
     pub attachment_indexes: Vec<records::AttachmentIndex>,
     pub metadata_indexes: Vec<records::MetadataIndex>,
 }
 
-impl fmt::Debug for Summary<'_> {
+impl fmt::Debug for Summary {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // Keep the actual maps as HashMaps for constant-time lookups,
         // but order everything up before debug printing it here.
@@ -551,69 +555,37 @@ impl fmt::Debug for Summary<'_> {
     }
 }
 
-impl<'a> Summary<'a> {
+impl Summary {
     /// Read the summary section of the given mapped MCAP file, if it has one.
-    pub fn read(mcap: &'a [u8]) -> McapResult<Option<Self>> {
-        let foot = footer(mcap)?;
-
-        // A summary start offset of 0 means there's no summary.
-        if foot.summary_start == 0 {
-            return Ok(None);
-        }
-
-        if foot.summary_crc != 0 {
-            // The checksum covers the entire summary _except_ itself, including other footer bytes.
-            let calculated =
-                crc32(&mcap[foot.summary_start as usize..mcap.len() - MAGIC.len() - 4]);
-            if foot.summary_crc != calculated {
-                return Err(McapError::BadSummaryCrc {
-                    saved: foot.summary_crc,
-                    calculated,
-                });
+    pub fn read(mcap: &[u8]) -> McapResult<Option<Self>> {
+        use std::io::{Read, Seek};
+        let mut cursor = std::io::Cursor::new(mcap);
+        let mut summary_reader = SummaryReader::new();
+        while let Some(event) = summary_reader.next_event() {
+            match event? {
+                SummaryReadEvent::Read(n) => {
+                    let read = cursor.read(summary_reader.insert(n))?;
+                    summary_reader.notify_read(read);
+                }
+                SummaryReadEvent::Seek(to) => {
+                    let pos = cursor.seek(to)?;
+                    summary_reader.notify_seeked(pos);
+                }
             }
         }
 
-        let mut summary = Summary::default();
-        let mut channeler = ChannelAccumulator::default();
-
-        let summary_end = match foot.summary_offset_start {
-            0 => MAGIC.len() - FOOTER_LEN,
-            sos => sos as usize,
-        };
-        let summary_buf = &mcap[foot.summary_start as usize..summary_end];
-
-        for record in LinearReader::sans_magic(summary_buf) {
-            match record? {
-                Record::Statistics(s) => {
-                    if summary.stats.is_some() {
-                        warn!("Multiple statistics records found in summary");
-                    }
-                    summary.stats = Some(s);
-                }
-                Record::Schema { header, data } => channeler.add_schema(header, data)?,
-                Record::Channel(c) => channeler.add_channel(c)?,
-                Record::ChunkIndex(c) => summary.chunk_indexes.push(c),
-                Record::AttachmentIndex(a) => summary.attachment_indexes.push(a),
-                Record::MetadataIndex(i) => summary.metadata_indexes.push(i),
-                _ => {}
-            };
-        }
-
-        summary.schemas = channeler.schemas;
-        summary.channels = channeler.channels;
-
-        Ok(Some(summary))
+        Ok(summary_reader.finish())
     }
 
     /// Stream messages from the chunk with the given index.
     ///
     /// To avoid having to read all preceding chunks first,
     /// channels and their schemas are pulled from this summary.
-    pub fn stream_chunk(
-        &self,
+    pub fn stream_chunk<'a, 'b: 'a>(
+        &'b self,
         mcap: &'a [u8],
         index: &records::ChunkIndex,
-    ) -> McapResult<impl Iterator<Item = McapResult<Message<'a>>> + '_> {
+    ) -> McapResult<impl Iterator<Item = McapResult<Message<'a>>> + 'a> {
         let end = (index.chunk_start_offset + index.chunk_length) as usize;
         if mcap.len() < end {
             return Err(McapError::BadIndex);
@@ -631,7 +603,7 @@ impl<'a> Summary<'a> {
         // Chunks from the LinearReader will always borrow from the file.
         // (Getting a normal reference to the underlying data back
         // frees us from returning things that reference this local Cow.)
-        let d: &'a [u8] = match d {
+        let d: &[u8] = match d {
             Cow::Borrowed(b) => b,
             Cow::Owned(_) => unreachable!(),
         };
@@ -734,7 +706,7 @@ impl<'a> Summary<'a> {
     /// Compressed chunks aren't random access -
     /// this decompresses everything in the chunk before
     /// [`message.offset`](records::MessageIndexEntry::offset) and throws it away.
-    pub fn seek_message(
+    pub fn seek_message<'a>(
         &self,
         mcap: &'a [u8],
         index: &records::ChunkIndex,
--- a/rust/src/records.rs
+++ b/rust/src/records.rs
@@ -414,6 +414,23 @@ pub struct ChunkIndex {
     pub uncompressed_size: u64,
 }
 
+impl ChunkIndex {
+    /// Returns the offset in the file to the start of compressed chunk data.
+    /// This can be useful for retrieving just the compressed content of a chunk given its index.
+    pub fn compressed_data_offset(&self) -> u64 {
+        self.chunk_start_offset
+        + 1 // opcode
+        + 8 // chunk record length
+        + 8 // start time
+        + 8 // end time
+        + 8 // uncompressed size
+        + 4 // CRC
+        + 4 // compression string length
+        + (self.compression.len() as u64) // compression string
+        + 8 // compressed size
+    }
+}
+
 #[derive(Debug, Clone, Eq, PartialEq, BinRead, BinWrite)]
 pub struct AttachmentHeader {
     pub log_time: u64,
--- a/rust/src/records.rs
+++ b/rust/src/records.rs
@@ -14,6 +14,8 @@ use std::{
 
 use binrw::*;
 
+use crate::{McapError, McapResult};
+
 /// Opcodes for MCAP file records.
 ///
 /// "Records are identified by a single-byte opcode.
@@ -417,17 +419,23 @@ pub struct ChunkIndex {
 impl ChunkIndex {
     /// Returns the offset in the file to the start of compressed chunk data.
     /// This can be useful for retrieving just the compressed content of a chunk given its index.
-    pub fn compressed_data_offset(&self) -> u64 {
-        self.chunk_start_offset
-        + 1 // opcode
-        + 8 // chunk record length
-        + 8 // start time
-        + 8 // end time
-        + 8 // uncompressed size
-        + 4 // CRC
-        + 4 // compression string length
-        + (self.compression.len() as u64) // compression string
-        + 8 // compressed size
+    /// Returns [`McapError::TooLong`] if the resulting offset would be greater than [`u64::MAX`].
+    pub fn compressed_data_offset(&self) -> McapResult<u64> {
+        let res = self.chunk_start_offset.checked_add(
+            1 // opcode
+            + 8 // chunk record length
+            + 8 // start time
+            + 8 // end time
+            + 8 // uncompressed size
+            + 4 // CRC
+            + 4 // compression string length
+            + (self.compression.len() as u64) // 32-bit compression string length
+            + 8, // compressed size
+        );
+        match res {
+            Some(n) => Ok(n),
+            None => Err(McapError::TooLong(self.chunk_start_offset)),
+        }
     }
 }
 
--- a/rust/src/sans_io/indexed_reader.rs
+++ b/rust/src/sans_io/indexed_reader.rs
@@ -415,7 +415,7 @@ impl IndexedReader {
                 self.cur_compressed_chunk.len()
             };
             assert!(
-                buffer_length > self.cur_compressed_chunk_loaded_bytes + n,
+                buffer_length >= self.cur_compressed_chunk_loaded_bytes + n,
                 "notify_read called with n > last inserted length"
             );
             self.cur_compressed_chunk_loaded_bytes += n;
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

File: rust/src/sans_io/util.rs [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+    Read(usize),
+    SeekedTo(u64),
+}
+
+/// casts a 64-bit length value from an MCAP into a [`usize`].
+pub fn len_as_usize(len: u64) -> McapResult<usize> {
+    len.try_into().map_err(|_| McapError::TooLong(len))
+}
+
+pub fn get_decompressor(
```
james-rms, line 22: this didn't need to move in the end

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+                            self.state = State::SeekingToChunk;
+                            continue;
+                        }
+                    }
+                    self.cur_message_index += 1;
+                    self.chunk_slots[message_index.chunk_slot_idx].message_count -= 1;
+                    let record =
+                        &self.chunk_slots[message_index.chunk_slot_idx].buf[message_index.offset..];
+                    // This should not happen - we failed in the indexing process somehow.
+                    if record[0] != op::MESSAGE {
```
Muon, line 446: Use `assert!()`.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+    /// The reader needs more data to provide the next record. Call [`IndexedReader::insert`] then
+    /// [`IndexedReader::notify_read`] to load more data. The value provided here is a hint for how
+    /// much data to insert.
+    ReadRequest(usize),
+    /// The reader needs to seek to a different position in the file. Call
+    /// [`IndexedReader::notify_seeked`] to inform the reader of the result of the seek.
+    SeekRequest(SeekFrom),
+    /// Get a new message from the reader.
+    Message {
+        header: crate::records::MessageHeader,
```
Muon, line 27: This currently necessitates copies, but could be handled in a zero-copy fashion.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+///             IndexedReadEvent::Message{ header, data } => {
+///                 let channel = summary.channels.get(&header.channel_id).unwrap();
+///                 // do something with the message header and data
+///             }
+///         }
+///     }
+///     Ok(())
+/// }
+/// ```
+pub struct IndexedReader {
```
Muon, line 93: Move closer to the corresponding impl block.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+
+    /// Configure the reader to yield messages in the specified order (defaults to log-time order).
+    pub fn with_order(mut self, order: ReadOrder) -> Self {
+        self.order = order;
+        self
+    }
+
+    /// Configure the reader to yield only messages from topics matching this set of strings.
+    /// By default, all topics will be yielded.
+    pub fn include_topics<T: IntoIterator<Item = impl Deref<Target = str>>>(
```
Muon, line 168: `Item = impl AsRef<str>`

rust/src/sans_io/indexed_reader.rs
james-rms: This doesn't feel like a big deal to me, since the header is  2 + 4 + 8 + 8 bytes in total.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+        if let State::LoadingChunkData { into_slot } = &self.state {
+            let buffer_length = if self.chunk_indexes[self.cur_chunk_index]
+                .compression
+                .is_empty()
+            {
+                self.chunk_slots[*into_slot].buf.len()
+            } else {
+                self.cur_compressed_chunk.len()
+            };
+            if buffer_length < self.cur_compressed_chunk_loaded_bytes + n {
```
Muon, line 479: Use `assert!()`.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+    /// Inform the reader of the result of the latest seek of the underlying stream.
+    pub fn notify_seeked(&mut self, pos: u64) {
+        if self.at_eof && self.pos != pos {
+            self.at_eof = false;
+        }
+        // If we're currently loading data, we need to reset and start loading from the beginning.
+        if self.pos != pos && matches!(self.state, State::LoadingChunkData { .. }) {
+            let mut state = State::SeekingToChunk;
+            std::mem::swap(&mut state, &mut self.state);
+            let State::LoadingChunkData { .. } = state else {
```
Muon, line 496: Why this? We checked this in the `if` condition.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+    cur_message_index: usize,
+) -> McapResult<usize> {
+    let mut offset = 0usize;
+    let mut sorting_required = cur_message_index != 0;
+    let mut latest_timestamp = 0;
+    let new_message_index_start = message_indexes.len();
+    while offset < chunk_data.len() {
+        let record = &chunk_data[offset..];
+        let opcode = record[0];
+        if record.len() < 9 {
```
Muon, line 545: Magic number 9?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+            }
+        }
+        if !filter.channel_ids.is_empty() && !filter.channel_ids.contains(&msg.channel_id) {
+            offset = next_offset;
+            continue;
+        }
+        if !sorting_required {
+            sorting_required = msg.log_time < latest_timestamp
+        }
+        latest_timestamp = std::cmp::max(latest_timestamp, msg.log_time);
```
Muon, line 574: `latest_timestamp = latest_timestamp.max(msg.log_time)`

rust/src/sans_io/indexed_reader.rs
james-rms: No memory of why i wrote that. removed.

rust/src/sans_io/indexed_reader.rs
james-rms: 🤦 i've been annoyed that I had to reach into this weird namespace to call `max()` for like a year now.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+                        if self.at_eof {
+                            return Err(McapError::UnexpectedEof);
+                        }
+                        return Ok(Some(IndexedReadEvent::ReadRequest(need)));
+                    }
+                    // decompress the chunk into the current slot. For un-compressed chunks, we do
+                    // nothing, because we already loaded the "compressed" data into the chunk slot.
+                    let slot = &mut self.chunk_slots[*into_slot];
+                    //
+                    slot.buf.resize(uncompressed_size, 0);
```
Muon, line 291: Change the capacity, not the length. `zstd_safe` implements [WriteBuf](https://docs.rs/zstd-safe/7.2.3/zstd_safe/trait.WriteBuf.html) which can deal with uninitialized data properly. Also, why is this necessary for the case that the current chunk is uncompressed?

james-rms: If the current chunk is compressed, we need to give the caller somewhere to write compressed data, and then we need a separate buffer to hold the uncompressed data. We could get them to write compressed data directly into a chunk slot, but we'd need another chunk slot to decompress into later.

james-rms: Regarding changing the capacity - I still need to hand the caller a mutable buffer to write into, and there's nothing to stop them from reading from it, which I think prevents me from handing them a mutable slice past `len`.

Muon: Right, but I was asking about the first match case, where we have uncompressed data. Also re: capacity, for zstd-safe you just pass a mutable reference to the whole Vec, not a mutable slice.

rust/src/sans_io/indexed_reader.rs
Muon: It's not necessarily a big deal! Just a potential improvement.

rust/src/sans_io/indexed_reader.rs
james-rms: I don't want to pass a reference to the whole vec out to the caller, because I want them to append writes to the end and not mess with any of the prior data.

james-rms: When we have uncompressed data, we sitll need to resize back to the uncompressed size after all of their writes are done, because they may have written more into the slot than we actually requested.

Muon: I'm talking about this. I don't know what caller you're referring to. Here you're passing a full slice to zstd-safe instead of just the vec: https://github.com/foxglove/mcap/pull/1338/files/6fdbc9711d672d6479e6836d8c93550461b7555a#diff-dc5f3105e529bb7e824c94dc7f2f9c763db53f901d5a6325d63f5e27b5d6a59fR362-R365

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,867 @@
...
+        } else {
+            BTreeSet::new()
+        };
+
+        // filter out chunks that we won't use
+        let mut chunk_indexes: Vec<crate::records::ChunkIndex> = summary
+            .chunk_indexes
+            .clone()
+            .into_iter()
+            .filter(|chunk_index| {
```
bennetthardwick, line 152: nit: you could do `iter().filter().cloned().collect()` to avoid a double allocation

rust/src/sans_io/indexed_reader.rs
james-rms: I think I get what you're getting at re: zstd, i'll take a look.

rust/src/sans_io/indexed_reader.rs
Muon: Just write down a `const` for this please!

rust/src/read.rs
```diff
@@ -631,7 +603,7 @@ impl<'a> Summary<'a> {
         // Chunks from the LinearReader will always borrow from the file.
```
Muon, line 603: Should this be addressed first? Why does it return a Cow then?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+                    }
+                }
+                false
+            })
+            .collect();
+
+        // put the chunk indexes in the order that we want to read them
+        match options.order {
+            ReadOrder::File => {
+                chunk_indexes.sort_by(|a, b| a.chunk_start_offset.cmp(&b.chunk_start_offset))
```
Muon, line 183:
```suggestion
                chunk_indexes.sort_by_key(|x| x.chunk_start_offset)
```

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+            }
+            ReadOrder::LogTime => {
+                chunk_indexes.sort_by(|a, b| {
+                    match a.message_start_time.cmp(&b.message_start_time) {
+                        std::cmp::Ordering::Equal => {
+                            a.chunk_start_offset.cmp(&b.chunk_start_offset)
+                        }
+                        other => other,
+                    }
+                });
```
Muon, line 193:
```suggestion
                chunk_indexes.sort_by_key(|x| (x.message_start_time, x.chunk_start_offset));
```

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+            }
+            ReadOrder::ReverseLogTime => {
+                chunk_indexes.sort_by(|a, b| match b.message_end_time.cmp(&a.message_end_time) {
+                    std::cmp::Ordering::Equal => b.chunk_start_offset.cmp(&a.chunk_start_offset),
+                    other => other,
+                });
```
Muon, line 199: As above.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+            }
+        };
+
+        // check through all chunk indexes once to ensure that we have address space for an
+        // uncompressed chunk.
+        for chunk_index in chunk_indexes.iter() {
+            if chunk_index.compressed_size > usize::MAX as u64 {
```
Muon, line 206: I am not sure what this is supposed to be testing, but if `usize` is bigger than `u64` then this is going to just be `u64::MAX` and if `usize` is smaller than `u64` this is going to be `usize::MAX`. Is that the intent? Maybe make a `const MAX_SUPPORTED_CHUNK_SIZE` for this if so and add a comment explaining.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+                        self.order,
+                        &self.filter,
+                        &mut self.message_indexes,
+                        self.cur_message_index,
+                    )?;
+                    slot.message_count = message_count;
+                    // If there is more dead space at the front of `self.message_indexes` than the
+                    // set of new message indexes, compact the message index array now.
+                    if message_count < (self.cur_message_index) {
+                        self.message_indexes.drain(0..self.cur_message_index);
```
Muon, line 344: Should message_indexes be a `VecDeque` instead?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+                        &self.chunk_slots[message_index.chunk_slot_idx].buf[message_index.offset..];
+                    assert_eq!(
+                        record[0],
+                        op::MESSAGE,
+                        "invariant: message indexes should point to message records"
+                    );
+                    let len = u64::from_le_bytes(record[1..9].try_into().unwrap()) as usize; // size checked when indexing
+                    let mut cursor = std::io::Cursor::new(&record[9..9 + len]);
+                    let header = MessageHeader::read_le(&mut cursor)?;
+                    let header_end = cursor.position() as usize; // we can assume position <= record.len() <= usize::MAX here
```
Muon, line 390: "We assume" or "it must be the case that"?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+                            compressed_data_start,
+                        ))));
+                    }
+                    // Seek is done, time to load the compressed chunk data.
+                    self.cur_compressed_chunk.clear();
+                    self.cur_compressed_chunk_loaded_bytes = 0;
+                    self.state = State::LoadingChunkData {
+                        into_slot: find_or_make_chunk_slot(
+                            &mut self.chunk_slots,
+                            cur_chunk.uncompressed_size as usize, // size checked in new()
```
Muon, line 263: Maybe this could be cleaner if there was a newtype wrapper for "lesser of `u64` or `usize`" providing `From` impls for both `u64` and `usize`.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+/// }
+/// ```
+pub struct IndexedReader {
+    // This MCAP's chunk indexes, pre-filtered by time range and topic and sorted in the order
+    // they should be visited.
+    chunk_indexes: Vec<ChunkIndex>,
+    // The index in `chunk_indexes` of the current chunk to be loaded. cur_chunk_index >=
+    // chunk_indexes.len() means that all chunks have been loaded.
+    cur_chunk_index: usize,
+    // A set of decompressed chunks. Slots are re-used when their message count reaches zero.
```
Muon, line 100: Some explanation on why there are multiple slots would be good.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+    cur_compressed_chunk_loaded_bytes: usize,
+    // The current known position of the reader in the underlying file.
+    pos: u64,
+    // describes what the indexed reader is currently trying to do.
+    state: State,
+    // What order messages should be yielded
+    order: ReadOrder,
+    // Criteria for what messages from the MCAP should be yielded
+    filter: Filter,
+    at_eof: bool,
```
Muon, line 123: Should this be encoded in `State`?

rust/src/sans_io/summary_reader.rs
```diff
@@ -223,9 +223,10 @@ impl SummaryReader {
         self.at_eof = n == 0;
         match &mut self.state {
             State::ReadingFooter { loaded_bytes, .. } => {
-                if self.footer_buf.len() < *loaded_bytes + n {
-                    panic!("notify_read called with n > last inserted length");
-                }
+                assert!(
+                    self.footer_buf.len() >= *loaded_bytes + n,
+                    "notify_read called with n > last inserted length",
```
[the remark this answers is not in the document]
james-rms: If you hit this assertion, you're calling `notify_read` with a bogus `n`, because you can't have read more bytes into our buffer than the size we gave you.
```rust
let buf = reader.insert(1024); // returns a buffer of size 1024;
let num_bytes_read = file.read(buf)?; // num_bytes_read should always by <= 1024
reader.notify_read(num_bytes_read); // asserts if num_bytes_read > 1024
```

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+        }
+        // If we're currently loading data, we need to reset and start loading from the beginning.
+        if self.pos != pos && matches!(self.state, State::LoadingChunkData { .. }) {
+            self.state = State::SeekingToChunk;
+        }
+        self.pos = pos;
+    }
+
+    /// Get a mutable buffer of size `n` to read new MCAP data into from the stream.
+    pub fn insert(&mut self, n: usize) -> &mut [u8] {
```
Muon, line 439: I'm a bit confused here. If we're getting a buffer, what's being inserted?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+fn index_messages(
+    chunk_slot_idx: usize,
+    chunk_data: &[u8],
+    order: ReadOrder,
+    filter: &Filter,
+    message_indexes: &mut Vec<MessageIndex>,
+    cur_message_index: usize,
+) -> McapResult<usize> {
+    let mut offset = 0usize;
+    let mut sorting_required = cur_message_index != 0;
```
Muon, line 535: What does the index have to do with sorting?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+        offset = next_offset
+    }
+    match order {
+        ReadOrder::File => {
+            // in file order, message indexes do not need sorting
+        }
+        ReadOrder::LogTime => {
+            if sorting_required {
+                let unread_message_indexes = &mut message_indexes[cur_message_index..];
+                unread_message_indexes.sort_by(|a, b| a.log_time.cmp(&b.log_time));
```
Muon, line 589: `sort_by_key`

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+            }
+        }
+        ReadOrder::ReverseLogTime => {
+            // first, reverse the order of the new message indexes. This will make the sort much faster.
+            let new_message_indexes = &mut message_indexes[new_message_index_start..];
+            new_message_indexes.reverse();
+            if sorting_required {
+                let unread_message_indexes = &mut message_indexes[cur_message_index..];
+                unread_message_indexes.sort_by(|a, b| b.log_time.cmp(&a.log_time));
```
Muon, line 598: As above.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+    chunk_index.chunk_start_offset
+    + 1 // opcode
+    + 8 // chunk record length
+    + 8 // start time
+    + 8 // end time
+    + 8 // uncompressed size
+    + 4 // CRC
+    + 4 // compression string length
+    + (chunk_index.compression.len() as u64) // compression string
+    + 8 // compressed size
```
Muon, line 632: Overflow considered?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+                    self.state = State::LoadingChunkData {
+                        into_slot: find_or_make_chunk_slot(
+                            &mut self.chunk_slots,
+                            cur_chunk.uncompressed_size as usize, // size checked in new()
+                        ),
+                    };
+                    continue;
+                }
+                State::LoadingChunkData { into_slot } => {
+                    // This is a defensive check - the reader should not enter this state if there
```
Muon, line 269: Defending against what?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,870 @@
...
+                            let mut decoder = lz4::Decoder::new(std::io::Cursor::new(
+                                &self.cur_compressed_chunk[..compressed_size],
+                            ))?;
+                            decoder.read_exact(&mut slot.buf[..])?;
+                        }
+                        other => return Err(McapError::UnsupportedCompression(other.into())),
+                    }
+                    // index the current chunk slot
+                    // before starting, check if all existing message indexes have been exhausted -
+                    // if so, clear them out now.
```
Muon, line 326: Why do we do this?

rust/src/sans_io/summary_reader.rs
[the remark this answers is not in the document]
Muon: Could we instead pass owned buffers around?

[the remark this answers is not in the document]
james-rms: I don't really have a clear idea of how passing owned buffers around would/should work - if you have a concrete proposal, i could think more about it. i'll update the assertion message.

Given the current semantics of the reader, i still feel that `insert` is appropriate - you're inserting data into the reader.

rust/src/sans_io/indexed_reader.rs
james-rms: I need to do various sorting operations in-place on `message_indexes`, and VecDeque doesn't let me do that since it's not guaranteed to be a contiguous slice.

rust/src/sans_io/indexed_reader.rs
james-rms: Probably, but I think it wouldn't reflect my intent as well. The reader maintains its best knowledge of whether it's at EOF, regardless of what it's trying to do right now.

Muon: I'm not sure what you mean by that. Are all possible states valid when combined with either value of `at_eof`? And what does `at_eof` mean precisely?

rust/src/sans_io/summary_reader.rs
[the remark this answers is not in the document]
Muon: We could return a `Vec<u8>` or `ChunkSlot`, and leave it to the user to fill it up and pass it back.

rust/src/sans_io/indexed_reader.rs
Muon: Resolved but not done?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,868 @@
...
+        ] {
+            let mut expected: Vec<(u16, u64)> = chunks.iter().cloned().flatten().cloned().collect();
+            match order {
+                ReadOrder::File => {}
+                // sort in log time order (stable, so that file order is preserved) for equal values
+                ReadOrder::LogTime => expected.sort_by(|a, b| a.1.cmp(&b.1)),
+                // sort in log time order then reverse
+                ReadOrder::ReverseLogTime => {
+                    expected.sort_by(|a, b| a.1.cmp(&b.1));
+                    expected.reverse();
```
Muon, line 732:
```suggestion
                    expected.sort_by(|a, b| a.1.cmp(&b.1).reverse());
```

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,868 @@
...
+    }
+
+    fn test_read_order(chunks: &[&[(u16, u64)]]) {
+        let mcap = make_mcap(None, chunks);
+        for order in [
+            ReadOrder::LogTime,
+            ReadOrder::ReverseLogTime,
+            ReadOrder::File,
+        ] {
+            let mut expected: Vec<(u16, u64)> = chunks.iter().cloned().flatten().cloned().collect();
```
Muon, line 724: `cloned()` twice?

rust/src/sans_io/indexed_reader.rs
james-rms: > Are all possible states valid when combined with either value of at_eof?
yes. The underlying file's read position is either at the end of the file or not, regardless of our internal state.

rust/src/records.rs
```diff
@@ -414,6 +414,23 @@ pub struct ChunkIndex {
...
+    pub fn compressed_data_offset(&self) -> u64 {
+        self.chunk_start_offset
+        + 1 // opcode
+        + 8 // chunk record length
+        + 8 // start time
+        + 8 // end time
+        + 8 // uncompressed size
+        + 4 // CRC
+        + 4 // compression string length
+        + (self.compression.len() as u64) // compression string
```
Muon, line 429: Overflow?

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,841 @@
...
+
+    /// Configure the reader to yield messages in the specified order (defaults to log-time order).
+    pub fn with_order(mut self, order: ReadOrder) -> Self {
+        self.order = order;
+        self
+    }
+
+    /// Configure the reader to yield only messages from topics matching this set of strings.
+    /// By default, all topics will be yielded.
+    pub fn include_topics<T: IntoIterator<Item = impl Deref<Target = str>>>(
```
Muon, line 414:
```suggestion
    pub fn include_topics<T: IntoIterator<Item = impl ToOwned<Owned = String>>>(
```

Muon, line 414:
```suggestion
    pub fn include_topics<T: IntoIterator<Item = impl Into<String>>>(
```

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,841 @@
...
+                    |err| {
+                        McapError::DecompressionError(zstd::zstd_safe::get_error_name(err).into())
+                    },
+                )?;
+                if n != uncompressed_size {
+                    return Err(McapError::DecompressionError(format!(
+                        "zstd decompression error: expected {uncompressed_size}, got {n}"
+                    )));
+                }
+            }
```
Muon, line 311: This got moved around, but it's still unnecessarily zeroing stuff out in the zstd case.

rust/src/read.rs
Muon: This is still relevant.

rust/src/sans_io/indexed_reader.rs
Muon: Still relevant.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,841 @@
...
+            ReadOrder::File => {
+                chunk_indexes.sort_by_key(|chunk_index| chunk_index.chunk_start_offset);
+            }
+            ReadOrder::LogTime => {
+                chunk_indexes.sort_by_key(|chunk_index| chunk_index.message_start_time);
+            }
+            ReadOrder::ReverseLogTime => {
+                // load chunks in reverse order of their _end_ time.
+                chunk_indexes.sort_by_key(|chunk_index| chunk_index.message_end_time);
+                chunk_indexes.reverse();
```
Muon, line 186:
```suggestion
                chunk_indexes.sort_by_key(|chunk_index| std::cmp::Reverse(chunk_index.message_end_time));
```

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,841 @@
...
+                    (
+                        index.log_time,
+                        chunk_slots[index.chunk_slot_idx].data_start,
+                        index.offset,
+                    )
+                });
+            }
+        }
+        ReadOrder::ReverseLogTime => {
+            // first, reverse the order of the new message indexes. This will make the sort much faster.
```
Muon, line 525: Why does this make the sort much faster?

rust/src/sans_io/indexed_reader.rs
Muon: Can this be `chunks.iter().flatten().copied().collect()` instead?

rust/src/sans_io/indexed_reader.rs
james-rms: OK, this might not be true for the rust version, and I should check, but it was true for the Go sort implementation. Reversing then sorting an array that is probably already in reverse order is usually faster than just sorting it.

james-rms (commented):
Need to make a comment to get my responses back

rust/src/read.rs
```diff
@@ -496,15 +502,15 @@ impl Iterator for MessageStream<'_> {
...
+const FOOTER_LEN: usize = 20; // 20 bytes + 8 byte len + 1 byte opcode
 
 /// Read the MCAP footer.
 ///
 /// You'd probably prefer to use [`Summary::read`] to parse the whole summary,
 /// then index into the rest of the file with
 /// [`Summary::stream_chunk`], [`attachment`], [`metadata`], etc.
 pub fn footer(mcap: &[u8]) -> McapResult<records::Footer> {
-    if mcap.len() < MAGIC.len() * 2 + FOOTER_LEN {
+    if mcap.len() < MAGIC.len() * 2 + (9 + 1 + FOOTER_LEN) {
```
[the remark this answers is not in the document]
Muon: Good catch. No more manual size calculations please. If you're doing this, please declare some `#[repr(packed)]` struct types and use `size_of`.

rust/src/sans_io/indexed_reader.rs
```diff
@@ -0,0 +1,872 @@
...
+                        "zstd decompression error: expected {uncompressed_size}, got {n}"
+                    )));
+                }
+            }
+            #[cfg(feature = "lz4")]
+            "lz4" => {
+                slot.buf.resize(uncompressed_size, 0);
+                use std::io::Read;
+                let mut decoder = lz4::Decoder::new(std::io::Cursor::new(compressed_data))?;
+                decoder.read_exact(&mut slot.buf[..])?;
```
bennetthardwick, line 340: Is there some cost to creating a new decoder every time? Like setting up a thread pool or something?

rust/src/read.rs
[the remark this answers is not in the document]
james-rms: Yep, good catch.

I don't want to declare some #[repr(packed)] struct types and use size_of(). They wouldn't be used as part of decoding the file, so they would only exist as a more verbose and roundabout way of saying "add these numbers together and use them as an offset"

I will lay out the sum and comment with its meaning.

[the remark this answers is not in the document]
Muon: Verbose, yes, but not roundabout. IMO correctness is more obvious that way. Expanding out the sum for each field would be an OK compromise though.

rust/src/read.rs
```diff
@@ -496,15 +502,22 @@ impl Iterator for MessageStream<'_> {
...
 /// Read the MCAP footer.
 ///
 /// You'd probably prefer to use [`Summary::read`] to parse the whole summary,
 /// then index into the rest of the file with
 /// [`Summary::stream_chunk`], [`attachment`], [`metadata`], etc.
 pub fn footer(mcap: &[u8]) -> McapResult<records::Footer> {
-    if mcap.len() < MAGIC.len() * 2 + FOOTER_LEN {
+    let footer_record_len = 1 // opcode
+     + 8 // record length
+     + FOOTER_LEN;
```
Muon, line 517: Maybe
```rust
const RECORD_HEADER_LEN: usize = 1 /* opcode */ + 8 /* length */;
const FOOTER_RECORD_LEN: usize = RECORD_HEADER_LEN + FOOTER_LEN;
```
instead?

rust/src/sans_io/summary_reader.rs
```diff
@@ -0,0 +1,356 @@
...
+
+use crate::{
+    parse_record,
+    records::{Footer, Record},
+    sans_io::linear_reader::{LinearReadEvent, LinearReader, LinearReaderOptions},
+    McapError, McapResult, Summary, MAGIC,
+};
+use std::io::SeekFrom;
+
+const FOOTER_RECORD_AND_END_MAGIC: usize = 1 + 8 + 20 + 8;
```
Muon, line 11: Expand.

Muon (approved):

jtbandes:
🎉

PR #1339, opened 2025-02-25
rust: fix removal of identical channel data

james-rms:
### Changelog
- Fixed: fixed a bug where when writing messages using `mcap::Writer::write(&Message{...})`, introduced in #1297 . Writing messages on separate channels with identical channel data would cause channel data for some channels not to be written to the MCAP, resulting in an invalid MCAP.
### Docs

None.

### Description

MCAP as specified allows a 1:N relationship between channel content and channel IDs. This is intentional - there may be multiple sources publishing the exact same topic, and the output MCAP can and should contain one channel per publisher, not per topic.

#1297 introduced the use of a bimap to track the relationship between channel / schema IDs and their content. however, I overlooked at the time a key constraint of the bimap implementation, in that all values on the left and right side are unique.
 This means it could not track multiple identical channels or schemas with the same ID. This PR introduces an additional (u16, u16) map for channels and schemas which tracks the mapping of all IDs to their "canonical" IDs in the bimap. This allows us to track the presence of multiple records with the same content but different IDs, while also preserving the quick lookups that the bimap was introduced for.

</td><td>

</td></tr></table>

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -453,6 +453,9 @@ impl<W: Write + Seek> Writer<W> {
         Ok(id)
     }
 
+    /// Write a schema record into the MCAP, and adds this schema to `canonical_schemas` and
+    /// `all_schema_ids`. The caller is responsible for ensuring this schema does not conflict with
+    /// a previously-written schema with the same ID.
     fn write_schema(&mut self, schema: Schema) -> McapResult<()> {
         let content = SchemaContent {
             name: Cow::Owned(schema.name.clone()),
@@ -460,16 +463,24 @@ impl<W: Write + Seek> Writer<W> {
             data: Cow::Owned(schema.data.clone().into_owned()),
         };
         if let Some(canonical_id) = self.canonical_schemas.get_by_left(&content) {
-            assert!(self
-                .all_schema_ids
-                .insert(schema.id, *canonical_id)
-                .is_none());
+            if let Some(old_canonical_id) = self.all_schema_ids.insert(schema.id, *canonical_id) {
+                assert!(
+                    old_canonical_id == *canonical_id,
+                    "caller must ensure that a second schema record with the same ID has identical content",
+                );
+            }
         } else {
-            assert!(!self
-                .canonical_schemas
-                .insert(content, schema.id)
-                .did_overwrite());
-            assert!(self.all_schema_ids.insert(schema.id, schema.id).is_none());
+            assert!(
+                !self
+                    .canonical_schemas
+                    .insert(content, schema.id)
+                    .did_overwrite(),
+                "caller must ensure that schema IDs do not conflict"
+            );
+            assert!(
+                self.all_schema_ids.insert(schema.id, schema.id).is_none(),
+                "caller must ensure that schema IDs do not conflict"
+            );
         }
 
         if self.options.use_chunks {
@@ -539,6 +550,9 @@ impl<W: Write + Seek> Writer<W> {
         Ok(id)
     }
 
+    /// Write a channel record into the MCAP, and adds this channel to `canonical_channels` and
+    /// `all_channel_ids`. The caller is responsible for ensuring this channel does not conflict
+    /// with a previously-written channel with the same ID.
     fn write_channel(&mut self, channel: records::Channel) -> McapResult<()> {
         let content = ChannelContent {
             topic: Cow::Owned(channel.topic.clone()),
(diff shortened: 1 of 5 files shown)


rust/src/write.rs
```diff
@@ -953,37 +964,42 @@ impl<W: Write + Seek> Writer<W> {
...
+            .all_channel_ids
             .iter()
-            .map(|(content, &id)| records::Channel {
-                id,
-                schema_id: content.schema_id,
-                topic: content.topic.clone().into(),
-                message_encoding: content.message_encoding.clone().into(),
-                metadata: content.metadata.clone().into_owned(),
+            .map(|(&id, canonical_id)| {
+                let content = self.canonical_channels.get_by_right(canonical_id).unwrap();
```
bennetthardwick, line 970: Might be worth using `expect` to explain why this will always be Some

rust/src/write.rs
```diff
@@ -526,15 +534,18 @@ impl<W: Write + Seek> Writer<W> {
...
+            topic: Cow::Owned(channel.topic.clone()),
+            schema_id: channel.schema_id,
+            message_encoding: Cow::Owned(channel.message_encoding.clone()),
+            metadata: Cow::Owned(channel.metadata.clone()),
+        };
+        if let Some(canonical_id) = self.canonical_channels.get_by_left(&content) {
+            self.all_channel_ids.insert(channel.id, *canonical_id);
+        } else {
+            self.canonical_channels.insert(content, channel.id);
+            self.all_channel_ids.insert(channel.id, channel.id);
```
bennetthardwick, line 547: Should these assert `.is_none()` just to be safe?

rust/src/write.rs
```diff
@@ -449,15 +453,36 @@ impl<W: Write + Seek> Writer<W> {
         Ok(id)
     }
 
+    /// Write a schema record into the MCAP, and adds this schema to `canonical_schemas` and
```
Muon, line 456: Those fields are private. What does it mean for the caller that this happens?

rust/src/write.rs
```diff
@@ -449,15 +453,36 @@ impl<W: Write + Seek> Writer<W> {
...
+    /// `all_schema_ids`. The caller is responsible for ensuring this schema does not conflict with
```
Muon, line 457: Or else what happens?

rust/src/write.rs
```diff
@@ -348,8 +348,10 @@ pub struct Writer<W: Write + Seek> {
     is_finished: bool,
     chunk_mode: ChunkMode,
     options: WriteOptions,
-    schemas: BiHashMap<SchemaContent<'static>, u16>,
-    channels: BiHashMap<ChannelContent<'static>, u16>,
+    canonical_schemas: BiHashMap<SchemaContent<'static>, u16>,
+    all_channel_ids: BTreeMap<u16, u16>,
+    canonical_channels: BiHashMap<ChannelContent<'static>, u16>,
+    all_schema_ids: BTreeMap<u16, u16>,
```
Muon, line 354: This sounds like a set of IDs. What does it map from/to?

rust/src/write.rs
```diff
@@ -525,16 +550,38 @@ impl<W: Write + Seek> Writer<W> {
         Ok(id)
     }
 
+    /// Write a channel record into the MCAP, and adds this channel to `canonical_channels` and
```
Muon, line 553: See comments above.

rust/src/write.rs
bennetthardwick: This method is private too. It only gets called when using the `write` API which takes a channel + schema in their "denormalized" format in Arcs.

rust/src/write.rs
```diff
@@ -549,19 +554,37 @@ impl<W: Write + Seek> Writer<W> {
...
         if let Some(schema) = message.channel.schema.as_ref() {
-            match self.schemas.get_by_right(&schema.id) {
-                Some(previous) => {
+            let content = SchemaContent {
+                name: Cow::Borrowed(&schema.name),
+                encoding: Cow::Borrowed(&schema.encoding),
+                data: Cow::Borrowed(&schema.data),
+            };
+            let canonical_schema_id = self.canonical_schemas.get_by_left(&content);
+            match self.all_schema_ids.get(&schema.id) {
```
Muon, line 563: Use entry API.

rust/src/write.rs
```diff
@@ -570,21 +593,48 @@ impl<W: Write + Seek> Writer<W> {
...
-                };
-                if *previous != current {
+        let channel_content = ChannelContent {
+            topic: Cow::Borrowed(&message.channel.topic),
+            schema_id: schema_id,
+            message_encoding: Cow::Borrowed(&message.channel.message_encoding),
+            metadata: Cow::Borrowed(&message.channel.metadata),
+        };
+        let canonical_channel_id = self.canonical_channels.get_by_left(&channel_content);
+        match self.all_channel_ids.get(&message.channel.id) {
```
Muon, line 603: Also a good spot for the entry API.

rust/src/write.rs
```diff
@@ -527,6 +527,18 @@ impl<W: Write + Seek> Writer<W> {
...
+                ChannelContent {
+                    topic: Cow::Owned(topic.to_owned()),
+                    schema_id,
+                    message_encoding: Cow::Owned(message_encoding.to_owned()),
+                    metadata: Cow::Owned(metadata.clone()),
+                },
+                id,
+            )
+            .expect("buggus");
+        self.all_channel_ids.insert(id, id);
```
Muon, line 541: Maybe assert this is None too?

rust/src/write.rs
```diff
@@ -527,6 +527,18 @@ impl<W: Write + Seek> Writer<W> {
...
+            .insert_no_overwrite(
+                ChannelContent {
+                    topic: Cow::Owned(topic.to_owned()),
+                    schema_id,
+                    message_encoding: Cow::Owned(message_encoding.to_owned()),
+                    metadata: Cow::Owned(metadata.clone()),
+                },
+                id,
+            )
+            .expect("buggus");
```
Muon: I assume this message is temporary? :P

rust/src/write.rs
```diff
@@ -549,19 +566,35 @@ impl<W: Write + Seek> Writer<W> {
...
+                    // this is OK.
+                    if let Some(canonical_schema_id) = canonical_schema_id {
+                        entry.insert(*canonical_schema_id);
+                    } else {
+                        self.canonical_schemas.insert_no_overwrite(SchemaContent{
+                            name: Cow::Owned(schema.name.clone()),
+                            data: Cow::Owned(schema.data.clone().into()),
+                            encoding: Cow::Owned(schema.encoding.clone()),
+                        }, schema.id).expect(
+                            "all right values in canonical_schemas should correspond to a key in all_schema_ids");
```
bennetthardwick, line 595: rustfmt makes me 😠

bennetthardwick (approved):

rust/src/write.rs
james-rms: Ill unfold it manually .

PR #1340, opened 2025-02-25
Optional only install `lz4` or `zstandard`

mrkbac:
### Changelog

Install optional only zstandard or lz4

### Docs

None

### Description

Zstandard and LZ4 are not available on all platforms and are not always required. To make it easier to install MCAP on platforms where these dependencies are not required, this PR makes these dependencies optional, without changing the current installation workflow.

`pip install mcap` installs `zstandard` and `lz4`
`pip install mcap --no-deps` does not install any child deps -> package still works


python/mcap/mcap/writer.py
```diff
@@ -427,9 +434,13 @@ def __finalize_chunk(self):
 
         chunk_data = self.__chunk_builder.end()
         if self.__compression == CompressionType.LZ4:
+            if lz4 is None:
+                raise RuntimeError("lz4 is not installed")
             compression = "lz4"
-            compressed_data: bytes = lz4.frame.compress(chunk_data)  # type: ignore
```
james-rms, line 431: Why did this switch away from frame compression?

python/mcap/mcap/stream_reader.py
```diff
@@ -70,9 +77,13 @@ def get_chunk_data_stream(
     chunk: Chunk, validate_crc: bool = False
 ) -> Tuple[ReadDataStream, int]:
     if chunk.compression == "zstd":
+        if zstandard is None:
+            raise ImportError("zstandard is required to decompress zstd chunks")
```
james-rms, line 81: I'm not sure about the convention for exceptions of this nature, but assuming ImportError is the right choice here, I would set the [keyword arguments it takes](https://docs.python.org/3/library/exceptions.html#ImportError) here.

python/mcap/mcap/writer.py
```diff
@@ -427,9 +434,13 @@ def __finalize_chunk(self):
 
         chunk_data = self.__chunk_builder.end()
         if self.__compression == CompressionType.LZ4:
+            if lz4 is None:
+                raise RuntimeError("lz4 is not installed")
```
james-rms, line 438: This is a configuration error, the error should be raised at configuration-time, not here when finalizing the chunk.

python/mcap/mcap/stream_reader.py
james-rms: In general I'd be surprised if I saw an import error at runtime. I would define a new error class for `UnsupportedCompressionError` and raise that here. I would also raise the same exception if the compression string was something other than "zstd", "lz4", or "".

python/mcap/mcap/writer.py
jtbandes: The import changed: `import lz4.frame as lz4`

james-rms: Right :+1:

james-rms (approved):

PR #1343, opened 2025-03-06
Call std::fflush() and ::fsync() before std::fclose() in FileWriter::end()

jhurliman:
### Changelog
- [C++] Ensure data is flushed to physical media before `FileWriter::end()` returns

### Docs

None

### Description

I've observed periodic data loss writing MCAP files where I destroy the `McapWriter` (which triggers `FileWriter::end()` under the hood) then immediately call `std::rename()` to remove a `.active` extension from the written MCAP file. Most of the time this works, but maybe ~5% of my MCAPs are missing data at the end where the final bytes of the file are a partial Chunk record.

From my reading it looks like calling `std::fclose()` immediately followed by `std::rename()` may not be safe if there are writes that haven't flushed to disk. This change adds additional calls to `std::fflush()` and `::fsync()` to guarantee the OS has delivered all pending writes to the physical media before `FileWriter::end()` returns, and hence before my `std::rename()` call.

This change was manually tested, I am still monitoring but haven't seen any partial writes since deploying this fix locally.

### DIFF
--- a/cpp/mcap/include/mcap/writer.inl
+++ b/cpp/mcap/include/mcap/writer.inl
@@ -69,6 +69,20 @@ void FileWriter::flush() {
 
 void FileWriter::end() {
   if (file_) {
+    std::fflush(file_);
+#if defined(_WIN32) || defined(_WIN64)
+    // Windows platform
+    const int fd = _fileno(file_);
+    if (fd != -1) {
+      _commit(fd);
+    }
+#elif defined(__unix__) || defined(__APPLE__) || defined(__linux__)
+    // Unix/Linux/macOS platforms
+    const int fd = ::fileno(file_);
+    if (fd != -1) {
+      ::fsync(fd);
+    }
+#endif
     std::fclose(file_);
     file_ = nullptr;
   }


cpp/mcap/include/mcap/writer.inl
```diff
@@ -69,6 +69,20 @@ void FileWriter::flush() {
 
 void FileWriter::end() {
   if (file_) {
+    std::fflush(file_);
+#if defined(_WIN32) || defined(_WIN64)
```
james-rms, line 73: we have a different set of platform-detection ifdefs in `visibility.hpp` - we don't check for win64, we do check for cygwin. 

From https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170, it looks like `_WIN64` is redundant with `_WIN32`.

Also, `::fsync` is only defined in `unistd.h`, which we don't include in `mcap`, so this code only works when built into another object that includes it. We probably need a conditional include at the top of the file.

james-rms: Similar with `_fileno`, which microsoft says is defined in `stdio.h` - we only include `cstdio`, which may resolve to the same thing? I haven't tested. https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fileno?view=msvc-170

defunctzombie:
@james-rms sounds like this is a bug in behavior? Or was that not verified? We should decide what action to take here.

PR #1344, opened 2025-03-06
Add information regarding rosbag2 conversion compression

scrouthtv:
### Changelog
Added documentation for using rosbag2 convert with mcap compression.

### Docs
None

### Description
When creating a rosbag with `ros2 bag convert` and this configuration:
```yaml
output_bags:
- uri: out
  storage_id: mcap
  compression_mode: file
  compression_format: zstd
  all_topics: true
```

A `out_0.mcap.zstd` file is created, which is unreadable by `ros2 play` -- the file must be first be fully inflated (taking up 4x the original space and long time) and can then be read.

Thanks to a contributor at rosbag2 (https://github.com/ros2/rosbag2/issues/1920) I was able to figure out a working configuration for converting to a compressed rosbag, which involves two separate configuration files:
```yaml
output_bags:
- uri: out
  storage_id: mcap
  storage_config_uri: mcap_config.yaml
  all_topics: true
```
and `mcap_config.yaml` as used by `ros2 bag record -s mcap --all --storage-config-file mcap_config.yaml`.

I think it would be nice to also have this feature documented in the mcap docs.

### DIFF
--- a/website/docs/guides/getting-started/ros-2.md
+++ b/website/docs/guides/getting-started/ros-2.md
@@ -39,10 +39,15 @@ $ cat << EOF > convert.yaml
 output_bags:
   - uri: ros2_output
     storage_id: mcap
+    storage_config_uri: my_storage_config.yaml
     all: true
 EOF
+$ cat << EOF > my_storage_config.yaml
+compression: Zstd
 $ ros2 bag convert -i ros2_input.db3 -o convert.yaml
 ```
+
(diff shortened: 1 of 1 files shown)


website/docs/guides/getting-started/ros-2.md
```diff
@@ -39,11 +39,16 @@ $ cat << EOF > convert.yaml
 output_bags:
   - uri: ros2_output
     storage_id: mcap
+    storage_config_uri: my_storage_config.yaml
     all: true
 EOF
+$ cat << EOF > my_storage_config.yaml
+compression: Zstd
```
james-rms, line 46: needs an EOF

james-rms (approved):

PR #1345, opened 2025-03-11
go: reject invalid utf-8 strings in MCAP

james-rms:
### Description

MCAP [strings](https://mcap.dev/spec#string) should be UTF-8 formatted, we should check for this when parsing.

We've seen some users provide invalid-formatted strings in schema names or in channel metadata, and we need to catch this at import time, rather than have the invalid data in storage and be impossible to stream.

</td><td>

</td></tr></table>

### DIFF
--- a/go/mcap/reader.go
+++ b/go/mcap/reader.go
@@ -6,7 +6,10 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"unicode/utf8"
 )
+
+var ErrInvalidString = errors.New("string is not valid UTF-8")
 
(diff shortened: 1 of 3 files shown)


bennetthardwick (approved):

james-rms:
I've decided not to merge this. The MCAP spec says strings "should" be Utf-8 formatted - it doesn't seem fair to fail to read a file when a string does not conform to this recommendation.

I'll add a check to `doctor` and into the inbox listener instead.

Muon:
Should we tweak stuff on our end to use `&[u8]` or `BStr` then?

PR #1347, opened 2025-03-18
Improve releases link

bryfox:
### Changelog
None
### Docs
None
### Description

This updates the CLI documentation to point to the releases page, filtered for CLI releases. The existing link (`/latest`) may or may not point to a CLI release.


eloff (approved):
Yes, that works! Thanks

PR #1350, opened 2025-03-24
replace `memmap` with `memmap2`

eric-seppanen:
### Changelog
Replace (unmaintained) `memmap` dev-dependency with `memmap2`.

### Docs

None, other than the corresponding doc-comment change in `lib.rs`.

### Description

`memmap` has been unmaintained for many years; there has been a RustSec [advisory](https://rustsec.org/advisories/RUSTSEC-2020-0077.html) about this for 4 years. As a dev-dependency this isn't critical, but it seems likely that users of this library will copy/paste example code. Setting a good example for them seems worthwhile.


bennetthardwick (approved):
Awesome thanks.

PR #1352, opened 2025-03-28
doctor: add check for non-utf8-formatted strings

james-rms:
### Description

the MCAP spec defines strings as valid utf-8, but we don't check that in Doctor (until now).

Doctor would not inspect strings for validity</td><td>

`mcap doctor` on an MCAP file with MCAP strings that are not valid utf-8 will produce an error.
</td></tr></table>

### DIFF
--- a/go/cli/mcap/cmd/doctor.go
+++ b/go/cli/mcap/cmd/doctor.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"os"
 	"reflect"
+	"unicode/utf8"
 
 	"github.com/fatih/color"
 	"github.com/foxglove/mcap/go/cli/mcap/utils"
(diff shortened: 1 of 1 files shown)


go/cli/mcap/cmd/doctor.go
```diff
@@ -649,7 +704,7 @@ func main(_ *cobra.Command, args []string) {
 	err := utils.WithReader(ctx, filename, func(remote bool, rs io.ReadSeeker) error {
 		doctor := newMcapDoctor(rs)
 		if remote {
-			doctor.warn("Will read full remote file")
+			color.Yellow("Will read full remote file")
```
Muon, line 707: What's going on here?

Muon (approved):

go/cli/mcap/cmd/doctor.go
james-rms: doctor.warn indicates something is wrong with the file that the caller needs to be aware of, and it increments a counter of warnings which is displayed at the end of the run. Warning the user that the tool will read the whole file isn't a warning about the file content itself.

jtbandes: could probably use a comment to that effect

PR #1356, opened 2025-04-10
fix: allow re-using writer for multiple files

sofuture:
### Changelog

Clear schemas in `McapWriter::terminate()`

### Docs

Fixes: FG-10773

### Description

Some internal state was left on McapWriter::terminate() contrary to docs.

### DIFF
--- a/cpp/examples/CMakeLists.txt
+++ b/cpp/examples/CMakeLists.txt
@@ -22,5 +22,8 @@ target_link_libraries(bag2mcap ${CONAN_LIBS})
 add_executable(mcapdump mcapdump.cpp)
 target_link_libraries(mcapdump ${CONAN_LIBS})
 
+add_executable(rotatemcap rotatemcap.cpp)
+target_link_libraries(rotatemcap ${CONAN_LIBS})
+
 add_subdirectory(protobuf)
 add_subdirectory(jsonschema)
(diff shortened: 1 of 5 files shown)


eloff (changes requested):
At first glance this appears to be the right fix.

But after *a LOT* of digging, I don't think it is.

I think schemas_ are meant to persist beyond just the mcap file currently being written.  open can be called again on the same writer (e.g. logging to rotating mcap files) and it's fine to re-use the same list of schemas with the same id mappings. A feature, not a bug.

The schemas written in each file are tracked separately in writtenSchemas_, which is cleared in terminate, which makes sense.

But the reported problem is still an issue. And I think it stems from other parts of the writer referencing schemas_ when the should be using writtenSchemas_, thus causing schemas to show up in files they weren't present in, in the summary section.

This happens in [close() when repeating the schemas](https://github.com/foxglove/mcap/blob/main/cpp/mcap/include/mcap/writer.inl#L390)

I think that should reference writtenSchemas_ instead.

@james-rms could you double-check my logic here? I can take over fixing this, but first I want to make sure I understand it correctly.

jtbandes:
Just want to interject: I don't think we have documented anywhere whether you are even supposed to be able to reuse the same writer for multiple files 😊

james-rms:
My answer to this is that it could go either way. We could
1. Document that the writer is not to be re-used between files. This leaves the question of what `terminate()` is even for - I don't know.
2. Make the change that @eloff suggests, and explicitly document that the writer can be re-used between files if you call `terminate` or `end`. We should add an example / test for this.

eloff:
Closing this in favor of a new PR

PR #1360, opened 2025-04-21
Add support for recording audio in RawAudio messages to mcap recording demo.

nidanin:
### Changelog
Add microphone support to the the MCAP demo page.

### Docs

None

### Description

Screenshot 2025-04-21 at 12 06 58 PM

</td><td>
Screenshot 2025-04-24 at 10 48 33 AM

</td></tr></table>


defunctzombie:
"JPEG" isn't video.

nidanin:
> "JPEG" isn't video.

Relabel as `Images` or something like that?

nidanin:
> Hm - Could "Camera" work? I know its not quite like "Audio" but maybe it works - you could also maybe flip around the Audio one as well and do "Microphone" and offer a dropdown to select which microphone?

Yes definitely doable. I know from experience that what browsers offer as "microphone" is a can of wormholes but we could untangle it.

defunctzombie:
One loss here is that before I could record multiple formats to the mcap (vp9, jpeg) - why did we remove that ability?

nidanin:
> One loss here is that before I could record multiple formats to the mcap (vp9, jpeg) - why did we remove that ability?

I can easily put it back if we think it's useful.

defunctzombie:
> I can easily put it back if we think it's useful.

I've found it useful to produce files with multiple formats for comparison.

website/package.json
```diff
@@ -22,11 +22,13 @@
     "@docusaurus/preset-classic": "2.4.1",
     "@foxglove/eslint-plugin": "1.0.1",
     "@foxglove/rostime": "1.1.2",
-    "@foxglove/schemas": "1.6.2",
+    "@foxglove/schemas": "github:foxglove/foxglove-sdk#workspace=@foxglove/schemas&commit=44efba283a4c2f7118c2546a7e70ea5692e2dbdd",
```
defunctzombie, line 25: Do we have a new version of schemas we can reference?

nidanin: I guess we need to cut a new release? I'm not sure exactly what's involved there.

defunctzombie: Yea - new release. Though I think its a new release of sdk now? Or maybe we have a way to do schemas? Ask the sdk folks.

defunctzombie (approved):
We need the updated schemas from npm (rather than the source reference) but seems good otherwise.

jtbandes (commented):
🙏 thanks for the refactoring, it was long overdue

PR #1364, opened 2025-04-23
rust: writer: return summary from finish()

james-rms:
### Changelog
- Changed: the `mcap::writer::Writer::finish()` method now returns a summary of the files' content.

### Docs

None.

### Description

We have a few cases internally where we want to write out an MCAP to some destination, then also track the summary information for that MCAP somewhere else. Cases include:
- log rotation, where we want to track schemas and channels written and would rather not do that externally
- maintaining an external index of MCAP chunks, which we do at Foxglove
- Writing then immediately reading back chunk data without needing to re-parse the just-written summary

This PR enables that by returning a summary from `finish()`.

Note: similar functionality is available from the Go writer, which exposes statistics and indexes as public fields from the Writer struct, see here: https://pkg.go.dev/github.com/foxglove/mcap/go/mcap#Writer

</td><td>

</td></tr></table>

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -17,10 +17,7 @@ use zstd::stream::{raw as zraw, zio};
 use crate::{
     chunk_sink::{ChunkMode, ChunkSink},
     io_utils::CountingCrcWriter,
-    records::{
-        self, op, AttachmentHeader, AttachmentIndex, MessageHeader, Record, SchemaHeader,
-        Statistics,
-    },
+    records::{self, op, AttachmentHeader, AttachmentIndex, MessageHeader, Record},
     Attachment, Channel, Compression, McapError, McapResult, Message, Schema, Summary, MAGIC,
 };
 
@@ -969,6 +966,7 @@ impl<W: Write + Seek> Writer<W> {
         self.finish_chunk()?;
 
         let summary = self.summarize();
+        self.finished_summary = Some(summary.clone());
 
         // Grab the writer - self.writer becoming None makes subsequent writes fail.
         let writer = match &mut self.writer {
@@ -1185,7 +1183,7 @@ impl<W: Write + Seek> Writer<W> {
                     id: *schema_id,
                     name: schema_content.name.clone().into(),
                     encoding: schema_content.encoding.clone().into(),
-                    data: schema_content.data.clone().into(),
+                    data: schema_content.data.clone(),
                 }),
             );
         }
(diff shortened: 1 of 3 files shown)


rust/src/write.rs
```diff
@@ -1151,19 +1141,76 @@ impl<W: Write + Seek> Writer<W> {
...
+        let stats = records::Statistics {
+            message_count: self.channel_message_counts.values().sum(),
+            schema_count: self.all_schema_ids.len() as u16,
+            channel_count: self.all_channel_ids.len() as u32,
+            attachment_count: self.attachment_count,
+            metadata_count: self.metadata_count,
+            chunk_count: self.chunk_indexes.len() as u32,
+            message_start_time: message_bounds.0,
+            message_end_time: message_bounds.1,
+            channel_message_counts: self.channel_message_counts.clone(),
```
Muon, line 1170: We're still using `take()` in other places but we're using `clone()` here. Why?

rust/src/write.rs
```diff
@@ -951,19 +952,21 @@ impl<W: Write + Seek> Writer<W> {
...
 
-    /// Finishes any current chunk and writes out the rest of the file.
+    /// Finishes any current chunk and writes out the summary section of the file.
     ///
-    /// Subsequent calls to other methods will panic.
-    pub fn finish(&mut self) -> McapResult<()> {
-        if self.is_finished {
+    /// Returns a [`Summary`] of data written to the file.  Subsequent calls to other methods will
+    /// panic.
+    pub fn finish(&mut self) -> McapResult<Summary> {
```
Muon, line 959: Maybe make this a `Cow<'_, Summary>` to avoid the clones?

Muon: (Though at some point in the future I'd prefer to change the API so finish() takes `mut self` instead...)

james-rms: yeah - I think the right move is to not implement Drop on the writer and for finish to take self here.

james-rms: That's an annoying (silent and fatal) breaking change though, so i'd prefer to do that along with a broader API rethink that forces users to re-write their code.

james-rms: Addressing the Cow<'_> idea: I don't think wrapping a reference or an owned value in Cow is useful unless we intend to return one or the other based on runtime state. I tried returning a reference to the summary, and I could probably get it to work, but it's very difficult to get the lifetimes to work out. I'm not that worried about the clone cost of a summary - finish is not expected to be called often for a given Writer (zero or once).

Muon: Yeah I wasn't that worried about the cost. Feel free to ignore this. And right, returning a reference or `Cow` in `finish()` means that the `Writer` has to be borrowed (and thus cannot be dropped) until that reference goes away. Hmm, annoying.

Muon (approved):

PR #1370, opened 2025-04-30
Updating Hero Logo and Dexterity

admfrk:
Updating the hero MCAP logo to reflect the new Foxglove brand along with updating Dexterity's new logo.


website/src/pages/index.tsx
```diff
@@ -102,8 +102,8 @@ export default function Home(): JSX.Element {
           <div className={styles.heroLogo}>
             <h1>{siteConfig.title}</h1>
             <img
-              src="/img/mcap720.webp"
```
jtbandes, line 105: @admfrk The old `mcap720.webp` file can be deleted if it's no longer used. Also this made me realize that there's also a `mcap240.webp` that needs to be updated too.

amacneil: this repo hosts the official high-res mcap logo. the new version in this PR is a lot lower quality. is there a higher quality image we can upload?

that's why there is both a high quality PNG + lower resolution webp's. it would be good to stick to the existing pattern unless there is good reason to change it.

admfrk: Let me ask Ellis. I used the png he provided. The webp is the same quality. Was the 240 v 720 the high-res v lower-res you're referring to @amacneil ?

website/static/img/mcap.png
jtbandes, line 1: The readme looks wrong now:
image

website/src/pages/index.tsx
jtbandes: I think my original comment here still applies – as far I can see this means `mcap720.webp` is no longer used, yet the file is being updated **and** a new one added at the same time in this PR. I think you will have to choose one: just update the old file & keep using it, or add a new file & delete the old one.

website/static/img/og-image.jpeg
jtbandes, line 1: According to github's diff, this file size has increased quite a bit (used to be 24kb, now 112kb) – not a super big deal and it does look like the new one may be higher quality on the hat. But just wanted to mention it. (btw it looks like both new & old have some vertical stripes in the background gradient - maybe we need to remake this image completely? cc @ellisneder)

website/static/img/mcap.png
admfrk: Which file does the readme use and why would it squish like that when everything is fine?

defunctzombie (approved):

PR #1371, opened 2025-05-01
cpp: add missing cstdint include (gcc15 build failure)

oysstu:
### Changelog
Add a missing cstdint include that causes a build error with GCC 15.1.1.


jtbandes (approved):
Thanks!

PR #1373, opened 2025-05-06
typescript: fix duplicated words in error message

achim-k:
### Changelog
None

### Docs
None

### Description
`... with logTime with logTime ...` -> `... with logTime ...`


jtbandes (approved):
I guess the error message was already so long a couple extra words were no big deal 😆

PR #1376, opened 2025-05-09
cpp: Bump to v2.0.1

gasmith:
### Changelog
- Fix a few bugs related to message-less chunks
- Add a missing cstdint include that causes a build error with GCC 15.1.1

### Docs
None

### Description
Just bumping the version, using #1305 as an example.


jtbandes (approved):
we may also have to manually make a PR to https://github.com/conan-io/conan-center-index/tree/master/recipes/mcap – not sure if these are able to be auto-updated

gasmith:
> we may also have to manually make a PR to https://github.com/conan-io/conan-center-index/tree/master/recipes/mcap – not sure if these are able to be auto-updated

https://github.com/conan-io/conan-center-index/pull/27390

PR #1379, opened 2025-05-11
Update mcap logo

amacneil:
Add vector logo


website/static/img/genthumbs.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -eux
```
james-rms, line 2: kinda weird that this shell script is now served from  https://6125524d.mcap.pages.dev/img/genthumbs.sh . Maybe move it to `website` and put a comment at the top explaining how/why to use it?

james-rms (commented):
`mcap.png` is still referenced from the top-level `README.md`.

website/static/img/genthumbs.sh
amacneil: yeah, its a bit weird I agree. but I feel like if I put it anywhere else, people are not going to find it (unless I put a readme in this directory? but then that readme would be served publicly and we are back where we started).

also the repo is open source, so its not like we are unintentionally exposing source code or something.

james-rms (approved):

PR #1381, opened 2025-05-12
Fix issue with re-using McapWriter

eloff:
### Changelog

Fix an issue where re-using a writer leads to schemas appearing in mcap files where they were not used. Documented how to re-use a writer correctly and added an example.

### Docs

I made changes to the doc comments, how do I deploy those?

### Description

There were some issues in terminate() not clearing all the state it needs to, as well as in writing the summary where schemas and channels that don't appear in the file can be written out.

I fixed those issues, added a test for them, added docs for how to re-use a McapWriter safely, and added an example.

Linear ticket: FG-10773

### DIFF
--- a/cpp/examples/rotatemcap.cpp
+++ b/cpp/examples/rotatemcap.cpp
@@ -1,0 +1,86 @@
+#define MCAP_IMPLEMENTATION
+#include <mcap/writer.hpp>
+
+#include <array>
+#include <chrono>
+#include <cstring>
+#include <fstream>
+#include <iostream>
+
+constexpr char StringSchema[] = "string data";
+constexpr char NumberSchema[] = "number data";
+
+mcap::Timestamp now() {
+  return mcap::Timestamp(std::chrono::duration_cast<std::chrono::nanoseconds>(
+                           std::chrono::system_clock::now().time_since_epoch())
+                           .count());
+}
+
+int main() {
+  mcap::McapWriter writer;
+
+  auto options = mcap::McapWriterOptions("ros1");
+  options.compression = mcap::Compression::Zstd;
+
+  std::ofstream out("output.mcap", std::ios::binary);
+  writer.open(out, options);
+
+  // Here we add both all the schemas and channels up front, but we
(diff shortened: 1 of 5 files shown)


cpp/mcap/include/mcap/writer.inl
```diff
@@ -389,16 +395,19 @@ void McapWriter::close() {
     ByteOffset schemaStart = fileOutput.size();
     if (!options_.noRepeatedSchemas) {
       // Write all schema records
-      for (const auto& schema : schemas_) {
-        write(fileOutput, schema);
+      for (const auto& schemaId : writtenSchemas_) {
+        write(fileOutput, schemas_[schemaId - 1]);
```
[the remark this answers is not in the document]
eloff: They are sequential and start from 1 :man_shrugging:

cpp/test/unit_tests.cpp
```diff
@@ -1056,3 +1056,116 @@ TEST_CASE("parsing", "header") {
...
+
+    bool hasSchema1 = false;
+    bool hasSchema2 = false;
+    for (const auto& [id, schema] : schemas) {
+      if (schema->name == "Schema1") hasSchema1 = true;
+      if (schema->name == "Schema2") hasSchema2 = true;
+    }
+
+    REQUIRE(hasSchema1);
+    REQUIRE(!hasSchema2);
```
jtbandes, line 1116: If `size() == 1` do we need all this boilerplate around finding schema1 and 2 individually? Can we just do something like `REQUIRE(schemas.front().second->topic == "Schema1")`?

cpp/examples/rotatemcap.cpp
jtbandes, line 1: What is the purpose of this example? We could use a comment or readme somewhere to describe what it's for.

jtbandes (approved):
Approach seems fine. Trying to read the tests I felt like the boilerplate outweighs the content, which to me means it's likely easier to introduce copy/paste errors (or AI errors?) without noticing since the boilerplate is so distracting – hence I left a suggestion for refactoring.

PR #1385, opened 2025-05-20
go: fix iteration edge case with overlapping chunks

### LINKED ISSUE #1374: go: mcap.InOrder(mcap.LogTimeOrder)，The message is not in log_time order
I called `reader.Messages()` with the `mcap.InOrder(mcap.LogTimeOrder)` option, but the data I read is still not fully ordered—some frames have a log_time that is smaller than the previous frame’

**I read the mcap file in time segments.**

`reader, err := mcap.NewReader(file)`
`it, err = reader.Messages(mcap.AfterNanos(uint64(start)*1e6), mcap.BeforeNanos(uint64(end)*1e6),
			mcap.InOrder(mcap.LogTimeOrder), mcap.WithTopics(topics))`

- Version:  mcap 1.7.1

**I don't have this problem with the py or C version**

Image

jtbandes:
### Changelog
go: fixed a bug with InOrder message reading when using additional filters with overlapping chunks.

### Docs

None

### Description

Previously, during `loadChunk()`, `sortingRequired` would only be set to true if the new chunk internally had some messages out of order (or if some messages had been read already?). However, the messages also need to be sorted even if none have been read yet and each chunk is individually ordered, to properly handle overlapping chunks.

Fixed by starting with `sortingRequired = true` and only setting it to false if no message indexes are outstanding. It kinda looks like the existing logic was already supposed to work this way but I'm not totally sure.

Fixes #1374

### DIFF
--- a/go/mcap/indexed_message_iterator.go
+++ b/go/mcap/indexed_message_iterator.go
@@ -287,12 +287,16 @@ func (it *indexedMessageIterator) loadChunk(chunkIndex *ChunkIndex) error {
 	}
 	// produce message indexes for the newly decompressed chunk data.
 	var maxLogTime uint64
+	// Always assume we need to sort newly added message indexes unless there are no outstanding
+	// messages. Even though chunks indexes are in order of messageStartTime, the message indexes
+	// within them may overlap or be out of order (especially when time/topic filters are considered).
+	sortingRequired := true
 	// if there are no message indexes outstanding, truncate now.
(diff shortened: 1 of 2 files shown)


james-rms (approved):

jtbandes:
@james-rms do you have any more context or thoughts from the original implementation as to whether there may be other similar bugs lurking? Or more testing we should do before merging? I guess these days this is mostly affecting CLI for us?

PR #1393, opened 2025-06-04
Python: Fix array b to bytes

AnthonyCvn:
### Changelog

Fix encoding issue when passing `array("B")` to `uint8[]` fields in ROS 2 messages

### Docs

None

### Description

Previously, users could not serialize `uint8[]` fields using `array("B", ...)`, which is a standard Python type for byte arrays. This caused compatibility issues when writing ROS 2 messages where the data was already in that form (e.g., images or raw sensor buffers).

This PR adds support for handling `array("B")` in addition to `bytes`, converting them appropriately for serialization. 

Passing `array("B", range(10))` would raise a `ValueError` and fail serialization

</td><td>

Correctly encodes the array and reads back the values without error

</td></tr></table>


gasmith (approved):
Looks reasonable to me. Thanks for the contribution!

PR #1394, opened 2025-06-12
rust: configurable compression level and threads

chriso:
### Changelog

This PR adds two new write options:
* `compression_level: u32` – compression level to use for zstd/lz4, or zero to use the default compression level.
* `compression_threads: u32` - number of threads to use for compression, or zero to disable multithreaded compression.

### Docs

None? Let me know if I need to update docs somewhere.

### Description

Prior to this PR, it wasn't possible to set the compression level; the default lz4 and zstd compression levels were used. Now, it's possible to set the lz4 and zstd compression levels.

Prior to this PR, non-wasm targets were forced to use multithreaded zstd compression ([reference](https://github.com/foxglove/mcap/blob/b404e42bd9ca6e9193ae994f7f3d752248ea8b06/rust/src/write.rs#L1368)). Now, it's possible to disable multithreaded compression, or to use a different number of compression threads.

lz4 supports multithreaded compression as of v1.10.0 (see [[1]](https://github.com/lz4/lz4/releases/tag/v1.10.0) and [[2]](https://github.com/lz4/lz4/pull/1336)), but this feature doesn't seem to be accessible from the [lz4](https://crates.io/crates/lz4) / [lz4-sys](https://crates.io/crates/lz4-sys) crates. For now, the `compression_level` option is ignored when using lz4 compression. Let me know if we should print a warning or fail.

I made these write options rather than parameterizing the [`Compression`](https://github.com/foxglove/mcap/blob/b404e42bd9ca6e9193ae994f7f3d752248ea8b06/rust/src/lib.rs#L158) enum variants, because these particular options are orthogonal to the compression algorithm, and because they're only applicable to writers.

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -313,7 +313,7 @@ impl WriteOptions {
     }
 
     /// Specifies how many threads to use for compression. A value of zero
-    /// to disables multithreaded compression. The default number of threads
+    /// disables multithreaded compression. The default number of threads
     /// is equal to the number of physical CPUs.
     #[cfg(feature = "zstd")]
     pub fn compression_threads(mut self, compression_threads: u32) -> Self {
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -119,13 +119,8 @@ pub struct WriteOptions {
     calculate_data_section_crc: bool,
     calculate_summary_section_crc: bool,
     calculate_attachment_crcs: bool,
-
-    /// Compression level, or zero to use the default compression level.
     #[cfg(any(feature = "zstd", feature = "lz4"))]
     compression_level: u32,
-
-    /// Threads to use for compression, or zero to disable
-    /// multithreaded compression.
     #[cfg(feature = "zstd")]
     compression_threads: u32,
 }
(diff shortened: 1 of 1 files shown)


rust/src/write.rs
```diff
@@ -1362,15 +1381,20 @@ impl<W: Write + Seek> ChunkWriter<W> {
             #[cfg(feature = "zstd")]
             Some(Compression::Zstd) => {
                 #[allow(unused_mut)]
-                let mut enc = zraw::Encoder::with_dictionary(0, &[])?;
+                let mut enc = zraw::Encoder::with_dictionary(compression_level as i32, &[])?;
                 // Enable multithreaded encoding on non-WASM targets.
                 #[cfg(not(target_arch = "wasm32"))]
-                enc.set_parameter(zraw::CParameter::NbWorkers(num_cpus::get_physical() as u32))?;
+                enc.set_parameter(zraw::CParameter::NbWorkers(compression_threads))?;
```
chriso, line 1399: A value of 1 means one background thread. Zero disables extra threads entirely. See https://github.com/gyscos/zstd-rs/blob/229054099aa73f7e861762f687d7e07cac1d9b3b/zstd-safe/src/lib.rs#L2228

rust/src/write.rs
```diff
@@ -1362,15 +1381,20 @@ impl<W: Write + Seek> ChunkWriter<W> {
...
+
                 Compressor::Zstd(zio::Writer::new(sink, enc))
             }
             #[cfg(feature = "lz4")]
             Some(Compression::Lz4) => Compressor::Lz4(
+                // Note: lz4-1.10.0 supports multithreaded compression
+                // (github.com/lz4/lz4/pull/1336), but this is not yet
+                // available through the lz4 / lz4-sys crates.
                 lz4::EncoderBuilder::new()
+                    .level(compression_level)
```
chriso, line 1409: With both lz4 and zstd,  `level=0` means use the default compression level. References:
* https://github.com/10XGenomics/lz4-rs/blob/287dbd063a53738917f1078192185839215d10a0/src/encoder.rs#L20
* https://github.com/facebook/zstd/blob/5e6bdf5e3dee6d977ea87b2c50c3bcb14fed3603/lib/zstd.h#L359

chriso:
Thanks for the quick review @bennetthardwick! Out of interest, how often is the crate updated? I'm happy to submit a PR to bump the `Cargo.toml` version if that helps.

bennetthardwick:
That'd be great - I should have asked you to chuck it in this one. We don't have any specific release cycles, we'll just put out a new release as things accumulate.

PR #1396, opened 2025-06-13
swift: use next tag for install action

bennetthardwick:
### Description

</td><td>

</td></tr></table>


bennetthardwick:
Jeez that did nothing.

PR #1399, opened 2025-06-13
typescript: support yielding unknown records from stream reader

bennetthardwick:
### Changelog

- typescript: add option to include unknown records in stream reader output, instead of skipping

### Docs

None

### Description

We're currently skipping unknown records in the Typescript `McapStreamReader`. It would be great to be able to get access to those unknown records in order to implement the application extensions / user proposals section of the spec.

I'm not sure whether the default behaviour should be skipping (since Typescript makes you handle the `Unknown` record anyway), but I've left it as is for backwards compatibility.

`McapStreamReader.nextRecord()` wouldn't yield unknown records

</td><td>

Setting `includeUnknownRecords` as a reader arg will yield them

</td></tr></table>

### DIFF
--- a/typescript/core/src/McapStreamReader.ts
+++ b/typescript/core/src/McapStreamReader.ts
@@ -29,6 +29,12 @@ type McapReaderOptions = {
    * reading starts in the middle of an MCAP file.
    */
   noMagicPrefix?: boolean;
+
+  /**
+   * When set to true, Unknown records will be returned from `nextRecord` instead of being skipped.
+   * This is useful for handling application extensions and user records.
+   */
+  includeUnknownRecords?: boolean;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


typescript/core/src/McapStreamReader.ts
```diff
@@ -29,6 +29,12 @@ type McapReaderOptions = {
    * reading starts in the middle of an MCAP file.
    */
   noMagicPrefix?: boolean;
+
+  /**
+   * When set to true, Unknown records will be returned from `nextRecord` instead of being skipped.
+   * This is useful for handling application extensions and user records.
+   */
+  includeUnknownRecords?: boolean;
```
jtbandes: I don't think we need this option. It should be fine to always yield the unknown records. We can maybe make a minor version bump but I wouldn't see this as a breaking change since the return type is already `TypedMcapRecord`.

bennetthardwick: Updated to remove the option

PR #1401, opened 2025-06-16
rust: support writing private records

bennetthardwick:
### Changelog

- rust: support writing private records

### Docs

None

### Description

This change adds support for a simple `write_private_record` method for writing user-defined private records as per the spec. Since private records may also appear in chunks the method has a boolean "chunkable" flag.

users cannot write records with custom opcodes

</td><td>

users can write private records

</td></tr></table>

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -111,7 +111,7 @@ pub enum McapError {
     #[error("Schema length ({header}) exceeds space in record ({available})")]
     BadSchemaLength { header: u32, available: u32 },
     #[error("Private records must have an opcode >= 0x80, got {opcode:#04x}")]
-    BadPrivateRecordOpcode { opcode: u8 },
+    PrivateRecordOpcodeIsReserved { opcode: u8 },
     #[error("Channel `{0}` has mulitple records that don't match.")]
     ConflictingChannels(String),
     #[error("Schema `{0}` has mulitple records that don't match.")]
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -110,8 +110,8 @@ pub enum McapError {
     BadChunkLength { header: u64, available: u64 },
     #[error("Schema length ({header}) exceeds space in record ({available})")]
     BadSchemaLength { header: u32, available: u32 },
-    #[error("Extension records must have an opcode >= 0x80, got {opcode:#04x}")]
-    BadExtensionOpcode { opcode: u8 },
+    #[error("Private records must have an opcode >= 0x80, got {opcode:#04x}")]
+    BadPrivateRecordOpcode { opcode: u8 },
     #[error("Channel `{0}` has mulitple records that don't match.")]
     ConflictingChannels(String),
     #[error("Schema `{0}` has mulitple records that don't match.")]
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -110,6 +110,8 @@ pub enum McapError {
     BadChunkLength { header: u64, available: u64 },
     #[error("Schema length ({header}) exceeds space in record ({available})")]
     BadSchemaLength { header: u32, available: u32 },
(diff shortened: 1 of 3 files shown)


james-rms:
We need to come up with a consistent name for "custom" records within the code and the spec - we refer to user records, extensions, unknown opcodes in various places. Place your vote here for what you want.

My personal taste doesn't love the pattern where we require the user to implement a trait on their own struct for our API to be satisfied.It feels more flexible to take these data as arguments:

```rust
writer.add_custom_record(opcode, data, chunkable);
```

bennetthardwick:
> My personal taste doesn't love the pattern where we require the user to implement a trait on their own struct for our API to be satisfied.It feels more flexible to take these data as arguments:

I've updated to have a method called `write_private_record` instead.

rust/src/write.rs
```diff
@@ -94,7 +94,12 @@ fn write_record<W: Write>(mut w: &mut W, r: &Record) -> io::Result<()> {
         Record::MetadataIndex(mi) => record!(op::METADATA_INDEX, mi),
         Record::SummaryOffset(so) => record!(op::SUMMARY_OFFSET, so),
         Record::DataEnd(eod) => record!(op::DATA_END, eod),
-        _ => todo!(),
+        Record::Unknown { opcode, data } => {
```
james-rms, line 104: private?

james-rms: Maybe its' fine as it is, as this covers both private records and unknown future-spec records.

bennetthardwick: Yeah if I did anything I'd split it up into two but I don't really want to break the API just yet.

rust/src/write.rs
```diff
@@ -732,6 +732,33 @@ impl<W: Write + Seek> Writer<W> {
...
 
+    /// Write a private record. If the record can be present in chunks then the `chunkable` flag
+    /// should be set to true.
+    ///
+    /// Private records must have an opcode >= 0x80.
+    pub fn write_private_record(
+        &mut self,
+        opcode: u8,
+        data: &[u8],
+        chunkable: bool,
```
james-rms: consider a two-value enum rather than a bool here?

bennetthardwick: I've added `PrivateRecordKind::Chunkable`. Happy to change if you have a better name.

rust/src/lib.rs
```diff
@@ -110,6 +110,8 @@ pub enum McapError {
     BadChunkLength { header: u64, available: u64 },
     #[error("Schema length ({header}) exceeds space in record ({available})")]
     BadSchemaLength { header: u32, available: u32 },
+    #[error("Private records must have an opcode >= 0x80, got {opcode:#04x}")]
+    BadPrivateRecordOpcode { opcode: u8 },
```
james-rms:
```suggestion
    PrivateRecordOpcodeIsReserved { opcode: u8 },
```

rust/src/write.rs
```diff
@@ -32,6 +32,13 @@ enum WriteMode<W: Write + Seek> {
     Attachment(AttachmentWriter<CountingCrcWriter<W>>),
 }
 
+pub enum PrivateRecordKind {
```
james-rms: I think shifting to PrivateRecordOptions (maybe an enum set or a struct of bools) is the right move.

rust/src/write.rs
```diff
@@ -32,11 +33,10 @@ enum WriteMode<W: Write + Seek> {
     Attachment(AttachmentWriter<CountingCrcWriter<W>>),
 }
 
-pub enum PrivateRecordKind {
+#[derive(EnumSetType, Debug)]
+pub enum PrivateRecordOptions {
     /// The private record can appear in chunks.
```
james-rms:
```suggestion
    /// If set and chunking is enabled, the private record will be written into a chunk. Otherwise, the record will be written directly to the file.
```

james-rms (approved):

PR #1402, opened 2025-06-16
docs: clarify extension record wording

bennetthardwick:
### Description

The docs aren't super clear about what "user proposals", "user extensions" or "application extensions" are, and why the opcodes from 0x80 are reserved for this purpose.

I propose we change this to simply say "private records". There is some precedent for this around; PNG defines "private chunks", MP4 defines "private streams" or "private extensions", TIFF defines "private tags". Other formats mention "application" or "custom" data or extensions.


james-rms (commented):
Private records sounds fine to me - i think it's worth a sentence of definition under ## records about this, like:

"Private records are application-specific and not defined within the MCAP spec."

bennetthardwick:
> Private records sounds fine to me - i think it's worth a sentence of definition under ## records about this, like:
> 
> "Private records are application-specific and not defined within the MCAP spec."

Added that line 👍🏻

PR #1406, opened 2025-06-18
python: bump package version

rwadman:
### Changelog
None

### Docs

None

### Description

Bump python mcap package version so that the latest changes to the package can be released to PyPi. There has been a few PRs merged in the package which haven't found their way into a release yet, and it would be nice to get those out.

I'm not at all sure that this is the right way to go about doing it, but mostly wanted to ask for the latest changes to be released.

These are the unreleased fixes:

foxglove/mcap#1316
foxglove/mcap#1340


python/mcap/mcap/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "1.2.2"
+__version__ = "1.2.3"
```
bennetthardwick:
```suggestion
__version__ = "1.3.0"
```

bennetthardwick:
Hey @rwadman thanks for this! I've just changed to a minor version bump because the changes added new behaviour in a backwards compatible way. But once this passes, I'll merge and push out the new version.

bennetthardwick (approved):

PR #1407, opened 2025-06-18
rust: add option to summary reader to error on large records

james-rms:
### Changelog
- Adds a record length limit check when reading summaries using SummaryReader.

### Description

</td><td>

</td></tr></table>


Muon (approved):

gasmith (commented):
This looks useful for constraining memory usage.

For dealing with corrupt files in particular, we might be able to get a better out-of-the-box experience, since in many situations (`SummaryReader` in particular), the file length is known information.

PR #1419, opened 2025-07-15
ci: Add armv7 build

patrickelectric:
### Changelog
Added CI support for ARM 32-bit (arm-linux-gnueabihf) cross-compilation.

### Docs
None

### Description
This change extends the CI matrix to include a new build target for 32-bit ARM using `arm-linux-gnueabihf-gcc`. This is useful for testing compatibility with devices running 32-bit ARM Linux, such as Raspberry Pi models or embedded systems, and ensures build correctness across more deployment environments.

Manually validated by verifying package installation (`gcc-arm-linux-gnueabihf`) and confirming CI matrix executes successfully.

CI did not test for 32-bit ARM targets.

</td><td>

CI includes a job with `arm-linux-gnueabihf-gcc` to build for 32-bit ARM.

</td></tr></table>

Fix #1416

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -425,6 +425,13 @@ jobs:
             env:
               CC: aarch64-linux-gnu-gcc
               CXX: aarch64-linux-gnu-g++
+          - os: linux
+            image: ubuntu-latest
+            arch: arm
+            setup: sudo apt-get update && sudo apt-get install -qq gcc-arm-linux-gnueabihf
+            env:
+              CC: arm-linux-gnueabihf-gcc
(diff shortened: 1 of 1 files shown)


amacneil:
Thanks! Do you think there is any need for ARMv6 support? I recall needing that to run on an older generation raspberry pi, maybe no one uses them. But it's also easy enough to support (I've got a test build running [here](https://github.com/foxglove/mcap/pull/1420)).

patrickelectric:
For my use case ARMv7 is enough, I only care for Raspberry Pi 3. The Raspberry Pi 2 is already 10 years old.

patrickelectric:
Thank you very much! It would be nice to have a release for it!

PR #1426, opened 2025-07-30
mcap CLI: Allow pulling latched topics to the start of the filtered segment

DaleKoenig:
### Changelog

Add `--latched-topic-regex` option to `mcap filter`, which pulls the most recent message of each matching topic to the beginning of the filtered timespan.

### Docs

Description is added through the command line parser and available with `mcap filter -h`

### Description

In ROS and similar systems, it is common for certain topics like `/tf_static` to be latched, where they are published once (or infrequently) and expected to be available to subscribers that come later.  rosbags/mcaps without these messages without these messages can often be difficult to analyze or visualize due to the lack of critical messages.  For this reason, ROS1 had [a flag for rosbag recorder](https://github.com/ros/ros_comm/pull/1850) to repeat such messages when the rosbag recorder split the rosbag, and [a similar feature may be planned for ROS2](https://github.com/ros2/rosbag2/issues/1159).  It is convenient to also have such a feature when snipping a small time segment of a large mcap, to ensure that the filtered mcap still has the necessary information for analysis and visualization.

Other notes:
- I have not tried to optimize too much, as the assumption is that messages on latched topics are not too frequent. Performance may be quite bad if someone uses this flag on a large high volume topic.
- DDS and other communication systems may support things like higher depth queues of latched topics (transient_local) or latched topics from multiple publishers on the same topic.  Handling these cases is I believe not feasible in this tool, and the case of depth=1, single publisher is by far the most prevalent.

Tested with --start-secs specified and -l flag on:
- Nonexistent topic (no effect on final result)
- Topic with multiple messages (Only the most recent message before start time was pulled to the beginning of the filtered mcap, as expected
- /tf_static -> Pulled to beginning of the filtered mcap

### DIFF
--- a/go/cli/mcap/cmd/filter.go
+++ b/go/cli/mcap/cmd/filter.go
@@ -20,6 +20,7 @@ type filterFlags struct {
 	output             string
 	includeTopics      []string
 	excludeTopics      []string
+	latchedTopics      []string
 	startSec           uint64
 	endSec             uint64
 	startNano          uint64
@@ -37,6 +38,7 @@ type filterOpts struct {
 	output             string
 	includeTopics      []regexp.Regexp
 	excludeTopics      []regexp.Regexp
+	latchedTopics      []regexp.Regexp
 	start              uint64
 	end                uint64
 	includeMetadata    bool
@@ -123,6 +125,13 @@ func buildFilterOptions(flags *filterFlags) (*filterOpts, error) {
 		return nil, err
 	}
 	opts.excludeTopics = excludeTopics
+
+	latchedTopics, err := compileMatchers(flags.latchedTopics)
+	if err != nil {
+		return nil, err
+	}
+	opts.latchedTopics = latchedTopics
+
 	opts.chunkSize = flags.chunkSize
 	return opts, nil
 }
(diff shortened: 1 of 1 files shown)


go/cli/mcap/cmd/filter.go
```diff
@@ -346,12 +363,60 @@ func filter(
...
+						mostRecentLatched[message.ChannelID].Data = make([]byte, len(message.Data))
+						copy(mostRecentLatched[message.ChannelID].Data, oldData)
+					}
+				}
 				continue
 			}
 			if message.LogTime >= opts.end {
 				continue
 			}
+			if !handledLatched {
```
james-rms, line 381:
```suggestion
			if !latchedMessagesWritten {
```

go/cli/mcap/cmd/filter.go
```diff
@@ -346,12 +363,60 @@ func filter(
 			if err != nil {
 				return err
 			}
+			mostRecent, ok := mostRecentLatched[message.ChannelID]
```
james-rms, line 366:
```suggestion
			mostRecent, topicIsLatched := mostRecentLatched[message.ChannelID]
```

PR #1431, opened 2025-08-05
cpp: allow application to determine visibility attributes

james-rms:
### Description

Proposed as an alternative to https://github.com/foxglove/mcap/pull/1380.

Adds an `#ifndef` guard around the existing definition logic in `visibility.hpp`. This allows application code to determine the meaning of MCAP_PUBLIC directly, if the implementation in `visibility.hpp` doesn't work for them.

</td><td>

</td></tr></table>

### DIFF
--- a/cpp/mcap/include/mcap/visibility.hpp
+++ b/cpp/mcap/include/mcap/visibility.hpp
@@ -5,21 +5,20 @@
  */
 #ifndef MCAP_PUBLIC
 #  if defined _WIN32 || defined __CYGWIN__
-#    ifdef __GNUC__
-#      define MCAP_EXPORT __attribute__((dllexport))
-#      define MCAP_IMPORT __attribute__((dllimport))
-#    else
-#      define MCAP_EXPORT __declspec(dllexport)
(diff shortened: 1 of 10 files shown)


cpp/mcap/include/mcap/visibility.hpp
Muon, line 1: `MCAP_EXPORT` and `MCAP_IMPORT` should probably not be guarded on `MCAP_PUBLIC` since I think they're intended to correspond to specific visibility attributes. Perhaps they could be guarded separately though? I am not sure off the top of my head whether that is desirable though since I am not very familiar with the DLL linking process.

james-rms: MCAP_IMPORT and MCAP_EXPORT aren't used anywhere outside of this file, so i feel safe guarding them / treating them however I want in this file. This isn't strictly true as there could be someone using these macros in their own code after including `mcap.hpp`, but that feels pretty unlikely.

james-rms: Removed MCAP_IMPORT and MCAP_EXPORT in this PR to remove the ambiguity.

PR #1441, opened 2025-08-22
rust: do not panic on next write if chunk closure fails

james-rms:
### Changelog
- rust: removed the condition for a panic if a subsequent write failed.
### Docs

None.

### Description

Previously, if we attempted to close a chunk and this failed, the writer would be left in an invalid state, with `self.writer = None`. This is no longer the case. Instead we return an explicit error value for cases where we can no longer write because the underlying file is invalid in an (unknown) state.

</td><td>

</td></tr></table>

### DIFF
--- a/rust/src/write.rs
+++ b/rust/src/write.rs
@@ -31,6 +31,7 @@ enum WriteMode<W: Write + Seek> {
     Raw(CountingCrcWriter<W>),
     Chunk(ChunkWriter<W>),
     Attachment(AttachmentWriter<CountingCrcWriter<W>>),
+    Failed(W),
 }
 
 #[derive(EnumSetType, Debug)]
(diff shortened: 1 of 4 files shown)


rust/src/write.rs
```diff
@@ -2154,4 +2214,51 @@ mod tests {
...
+    fn test_write_failure_does_not_cause_panic() {
+        #[derive(Default)]
+        struct FailingWriter {
+            write_count: i32,
+        }
+
+        impl std::io::Write for FailingWriter {
+            fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+                if self.write_count > 5 {
+                    return Err(std::io::Error::other("writes now fail"));
```
james-rms, line 2228:
```suggestion
                    return Err(io::Error::other("writes now fail"));
```

rust/src/write.rs
```diff
@@ -1545,26 +1585,45 @@ impl<W: Write + Seek> ChunkWriter<W> {
...
         let mut writer = CountingCrcWriter::with_hasher(writer, post_chunk_header_crc);
-
         // Write our message indexes
-        let data_end = writer.stream_position()?;
+        let data_end = match writer.stream_position() {
+            Ok(v) => v,
+            Err(err) => {
+                return Err((writer.finalize().0, err.into()));
+            }
+        };
```
bennetthardwick, line 1594: nit: if you wanted you could use a macro for these:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e131ddb259c38f3fb68e16318f559a32

PR #1442, opened 2025-08-27
cpp: Document the mcap vcpkg port

ethanmusser:
### Changelog

Add a brief blurb and badge referencing the new vcpkg port for mcap in the `README.md`.

### Docs

None

### Description

A vcpkg port is in the works in microsoft/vcpkg#47084.  This change calls out that port in the `README.md` so that new users are aware that it exists.

</td><td>

</td></tr></table>

### Draft TODOs

- [x] Wait for microsoft/vcpkg#47084 to land.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,14 +15,14 @@ MCAP is a modular container format and logging library for pub/sub messages with
 
 MCAP libraries are provided in the following languages. For guidance on each language, see its corresponding README:
 
-| Language              | Readme                 | API docs                                                        | Package name | Version                                                                              |
-| --------------------- | ---------------------- | --------------------------------------------------------------- | ------------ | ------------------------------------------------------------------------------------ |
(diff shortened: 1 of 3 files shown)


james-rms:
@ethanmusser not at this time. CMake wrappers are difficult to provide a good user experience for, since everyone needs something a little different there. The goal of the first-party MCAP C++ library is to be easy to integrate into whatever build system you need, CMake or otherwise.

ethanmusser:
> @ethanmusser not at this time. CMake wrappers are difficult to provide a good user experience for, since everyone needs something a little different there. The goal of the first-party MCAP C++ library is to be easy to integrate into whatever build system you need, CMake or otherwise.

Darn, figured it was worth asking.  I will say that the approach in [`olympus-robotics/mcap_builder`](https://github.com/olympus-robotics/mcap_builder) is pretty close to standard as far as vendoring a header-only CMake library like this one (with the `#define MCAP_IMPLEMENTATION` define).  But totally understood, thanks for the consideration and for calling out the third-party vcpkg port!

Issue #1454, opened 2025-09-17
`mcap filter` can break Foxglove User Scripts


philsc:
**Description**

I'm trying to cut the size of an MCAP file down to a small size. This is the command I'm using:
```console
$ mcap filter --start-secs 6114 --end-secs 6144 --include-topic-regex '.*' --output /home/philipp.schrader/logs/pro-35553/from_stop_increase.before.trimmed.mcap /home/philipp.schrader/logs/pro-35553/from_stop_increase.before.mcap
```

The command finishes successfully and I can look at the trimmed MCAP file.

The problem comes from `mcap filter` appearing to drop topics that have no messages during the specified time range.
This means that User Scripts on Foxglove stop working, erroring out with messages like this:
```
Type '"/spray/a brt.vpu.planner.PlannerDebug"' does not satisfy the constraint 'keyof MessageTypeByTopic'.
'event.message' is of type 'unknown'.
Parameter 'command' implicitly has an 'any' type.
Parameter 'index' implicitly has an 'any' type.
'event.message' is of type 'unknown'.
Parameter 'command' implicitly has an 'any' type.
Parameter 'index' implicitly has an 'any' type.
'event.message' is of type 'unknown'.
```

- Version:
```console
$ mcap version
v0.0.54
```

- Platform:
```console
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.5 LTS
Release:        22.04
Codename:       jammy

```

**Steps To Reproduce**

Trim a log file in a range where a topic doesn't have any messages.
This also happens with topics that never had any messages to begin with.

**Expected Behavior**

I would expect the empty channels to be present.

PR #1456, opened 2025-09-23
`mcap filter`: add last-per-channel filtering semantics

james-rms:
### Changelog
`mcap filter`: added `--include-last-per-channel-topic-regex` option. This includes the last message before the filter start on selected topics. This can be used to retain infrequently-logged topics in filtered MCAPs.
### Docs

None.

### Description

Fixes https://github.com/foxglove/mcap/issues/1454

This PR adds last-per-channel filtering semantics to `mcap filter`. This adds a new command-line option to select a set of topics to include the last message before the start time from.

This only works if the input file is being read in log-time order, so we add a new codepath that implements filtering using the indexed reader. This codepath is used for all seekable inputs. This should also improve performance when filtering by log time or topic, since these commands no longer need to read the entire file.

---

> [!NOTE]
> Enables retaining the most recent pre-start message per topic when filtering, with improved, index-backed filtering and stricter reader semantics.
> 
> - **CLI**: New `--last-per-channel-topic-regex` flag to include the last message before `start` for matching topics; only supported on seekable inputs.
> - **Filter implementation**: Added seekable path using `mcap.Reader` for topic/time filtering and log-time-ordered output; streaming path unchanged and rejects last-per-channel usage. Attachments/metadata handled via indexes.
> - **Reader behavior**: `reader.Messages` now errors if a non-file order is requested without indexes; improved error messages for invalid regexes.
> - **Tests**: Comprehensive coverage for seekable vs streaming paths, `includeTopic` precedence, last-per-channel behavior and ordering, and reader index requirement.
> - **Misc**: Version bump to `v1.7.4`; minor typos config update.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0f21eb521dc06ae4e88aad6795af13b448b1d713. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/go/cli/mcap/cmd/filter.go
+++ b/go/cli/mcap/cmd/filter.go
@@ -263,15 +263,12 @@ func filter(
 	} else {
 		err = filterStreaming(r, mcapWriter, opts)
 	}
-	closeErr := mcapWriter.Close()
-	if closeErr != nil {
-		if err != nil {
-			return fmt.Errorf("filter failed: %w, also failed to close writer: %w", err, closeErr)
-		}
-		return fmt.Errorf("failed to close writer: %w", closeErr)
-	}
 	if err != nil {
 		return fmt.Errorf("filter failed: %w", err)
+	}
+	err = mcapWriter.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close writer: %w", err)
 	}
 	return nil
 }
@@ -512,14 +509,6 @@ func filterStreaming(
 	if err != nil {
 		return err
 	}
-
-	defer func() {
-		err := mcapWriter.Close()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to close mcap writer: %v\n", err)
-			return
-		}
-	}()
 
 	buf := make([]byte, 1024)
 	schemas := make(map[uint16]markableSchema)


go/cli/mcap/cmd/filter.go
```diff
@@ -327,12 +344,58 @@ func filter(
...
+					}
+				}
 				continue
 			}
 			if message.LogTime >= opts.end {
 				continue
 			}
+			if !messagesBeforeRangeStartWritten {
+				messagesBeforeRangeStartWritten = true
+				// We have reached the start of the record, so here we need to add the latched topics
```
DaleKoenig: Maybe better to update the terminology "latched" in this comment as well

defunctzombie:
👍 to the semantics and having this be part of the CLI.

Do we need the `-topic-regex` suffix rather than having a separate `--include-last-per-channel` which would indicate that for any topic selected by the filter this rule would apply? So you would do:

```
mcap filter --include-last-per-channel -y /tf_static -y /imu
```

My thinking is that the typical desire would be to have any of the topics you wanted to filter on present in the truncated file without having to remember special ones to state separately in the flag.

Code review I leave to someone more familiar with the structure.

defunctzombie:
> For example, if a publish on a topic triggers some event, then pulling a message from that topic a minute forward in order to add it to the record could trigger events to happen during record playback that would not be expected

We no longer "pull" the topic forward and keep the original timestamp. Does this comment still apply in that context?

go/cli/mcap/cmd/filter.go
```diff
@@ -298,6 +309,12 @@ func filter(
 			if err != nil {
 				return err
 			}
+			for i := range opts.includeLastPerChannelTopics {
+				matcher := opts.includeLastPerChannelTopics[i]
```
jtbandes: nit: would this work? (same could be applied to other matchers below)

```suggestion
			for _, matcher := range opts.includeLastPerChannelTopics {
```

go/cli/mcap/cmd/filter.go
```diff
@@ -406,6 +469,12 @@ usage:
 			[]string{},
 			"messages with topic names matching this regex will be excluded, can be supplied multiple times",
 		)
+		includeLastPerChannelTopics := filterCmd.PersistentFlags().StringArray(
+			"include-last-per-channel-topic-regex",
+			[]string{},
+			"For included topics matching this regex, the most recent message previous to the start time"+
```
jtbandes: nit: prior sounds more natural to me
```suggestion
			"For included topics matching this regex, the most recent message prior to the start time"+
```

go/cli/mcap/cmd/filter.go
```diff
@@ -327,12 +344,58 @@ func filter(
...
+				for _, mostRecent := range mostRecentMessageBeforeRangeStart {
+					if mostRecent == nil {
+						continue
+					}
+					// We might still need to write the channel
+					channel, ok := channels[mostRecent.ChannelID]
+					if !ok {
+						continue
+					}
+					if !channel.written {
```
jtbandes: nit: is there a way to avoid the duplication here with the channel-writing code that already exists immediately below this block? At first glance, it looks like we should be able to take advantage of `mostRecent.ChannelID` being the same as `message.ChannelID` and avoid doing this twice.

go/cli/mcap/cmd/filter.go
```diff
@@ -327,12 +344,58 @@ func filter(
 			if err != nil {
 				return err
 			}
+			mostRecent, ok := mostRecentMessageBeforeRangeStart[message.ChannelID]
```
jtbandes: nit: Should we use `messagesBeforeRangeStartWritten` to short-circuit this logic if the messages have already been written, since they will never be written again?

go/cli/mcap/cmd/filter.go
```diff
@@ -327,12 +344,58 @@ func filter(
...
 			if message.LogTime < opts.start {
+				if ok {
+					if mostRecent == nil || mostRecent.LogTime <= message.LogTime {
+						mostRecentMessageBeforeRangeStart[message.ChannelID] = message
+						// Copy the data buffer explicitly, to avoid keeping a reference to the greater
+						// `buf` array that underlies `message.Data`.
+						mostRecentMessageBeforeRangeStart[message.ChannelID].Data = append([]byte{}, message.Data...)
```
jtbandes: caveat that I am not super familiar with this filter code and it's been a long time since I looked at it...but...are we assuming here that the file is ordered by log time?   I'm not sure if repeated calls to `lexer.Next(buf)` is giving us in-order reading using the index, but I suspect not. If I'm right, then it means this feature might not do what it claims to do if the file is disordered. I'm not sure what it would take to fix that, but if we don't think it's worth it to fully support disordered files, should we at least try to detect and warn/error when it happens, and maybe also document the limitation in the cli help?

go/cli/mcap/cmd/filter_test.go
```diff
@@ -311,3 +311,95 @@ func TestBuildFilterOptions(t *testing.T) {
...
+				1: 51,
+				2: 51,
+				3: 50,
+			},
+		},
+		{name: "does not override include topics",
+			flags: &filterFlags{
+				startNano:                   50,
+				includeLastPerChannelTopics: []string{"camera_.*"},
+				includeTopics:               []string{"camera_a"},
```
jtbandes, line 572: would it be worth adding a test case for excludeTopics as well?

jtbandes (approved):

go/cli/mcap/cmd/filter.go
jtbandes: on another note, the pre-emptive copying strikes me as possibly pessimizing, since we will create new copies of almost every preceding message until we reach the desired range. I see that @DaleKoenig already mentioned this too :)

I'm not sure I understand exactly what the copy is fixing -- would storing a reference to the underlying `buf` mostly be a concern if the number of channels is high (or total message size across channels is large?)  Are we specifically concerned that we might end up holding multiple preceding whole chunks in memory at once? At minimum, I think this comment could be expanded to clarify what we are avoiding (assuming you did some performance testing and found that this solution is best)

Another quick note that we could probably improve it a bit by re-using the buffer if we are replacing an item in the map. Since it's already been copied, if the buffer is large enough we should be able to copy into it without allocating again.

DaleKoenig: In my implementation, the reasoning was that the filter reads through the cli sequentially and old chunks are not kept in memory, so it is necessary to keep a copy of anything we might want to hold onto to only write at a later time.  So it did not seem feasible to keep a reference to the old data without copying it.  Implementing a method of keeping the old chunks around when they contain messages we want to write later seemed too invasive/difficult.

go/cli/mcap/cmd/filter.go
```diff
@@ -221,8 +269,214 @@ func filter(
...
+			writtenChannels[channel.ID] = true
+		}
+		if channel.SchemaID != 0 && !writtenSchemas[channel.SchemaID] {
+			// This invariant should be upheld by the reader, assert on it here.
+			die(
+				"found repeated channel record with ID %d and differing schema ID %d",
+				channel.ID,
+				channel.SchemaID,
+			)
+		}
```
Muon, line 332: This message isn't quite right, this check will pass as long as the schema referenced by this channel record's ID has been previously written. So e.g. this sequence of records might be accepted:

schema 1
schema 2
channel 1 with schema 1
channel 2 with schema 2
message on channel 1
message on channel 2
channel 1 with schema 2
message on channel 1

 I don't know if `mcap filter` should *die* if the source MCAP is malformed in this way, given it doesn't interpret the message contents. It should definitely at least emit a warning and probably exit with a nonzero status code by default though.

go/cli/mcap/cmd/filter.go
james-rms: I've updated this to iterate backwards from the start time to get the last-per-channel messages, which should mean no extra buffering required.

go/cli/mcap/cmd/filter.go
james-rms: It works, but triggers a lint error. When you do this, the go runtime copies the value out of the array on every iteration, which for this particular struct is 160B. indexing into the array within the loop avoids the copy.

go/cli/mcap/cmd/filter.go
```diff
@@ -221,8 +269,223 @@ func filter(
...
+		return mcapWriter.WriteMessage(msg)
+	}
+
+	// If any lastPerChannelTopics are specified, we iterate backwards from the start time to find them.
+	if len(opts.includeLastPerChannelTopics) > 0 {
+		channelsToWrite := map[uint16]bool{}
+		for _, ch := range info.Channels {
+			// make sure the topic is not separately excluded by topic filters
+			if includeAll || includeTopic(ch.Topic, opts) {
```
jtbandes: nit: could/should this be using the `topicSet` instead of re-running includeTopic?

go/cli/mcap/cmd/filter.go
```diff
@@ -212,6 +247,19 @@ func filter(
...
+	}
+	return filterStreaming(r, w, opts)
+}
+
+func filterSeekable(
+	rs io.ReadSeeker,
+	w io.Writer,
+	opts *filterOpts,
 ) error {
 	mcapWriter, err := mcap.NewWriter(w, &mcap.WriterOptions{
```
jtbandes, line 251: nit: I wonder if it would be helpful to factor out the conversion of `filterOpts` into `WriterOptions` (or into an mcap.Writer) so that we don't end up with two places we need to change if we add more filter options in the future

jtbandes (approved):
Looks good.

I am not sure I saw any test cases showing that the last per channel behavior actually works when reading from an out-of-order file using an index -- have you at least manually tested that? (or maybe we should add a test case for it -- might involve updating the writeFilterTestInput function)

PR #1465, opened 2025-10-06
cpp: enable & fix -Wshadow for GCC

jtbandes:
### Changelog
cpp: fixed `-Wshadow` warnings when building with GCC.

### Docs

None

### Description

Fixes #1462 by renaming some variables to avoid shadowing.

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -1,9 +1,11 @@
+.PHONY: default
 default: build
 
 .PHONY: dev-image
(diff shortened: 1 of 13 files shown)


clalancette (approved):
The changes to the code look fine.  I'm less certain about the removal of `CONAN_SYSTEM_INCLUDES` in the CMakeLists.txt, but I take it that you've tested it and it's OK.

PR #1472, opened 2025-10-15
Add S3 support and refactor related code for improved modularity

asokharev:
### Changelog
Add modular reader system with support for S3 and GCS, replacing the old GCS-only implementation with a unified `readers` registry.

### Docs
None

### Description
This change refactors how remote and local files are accessed in the MCAP CLI.  
Previously, only Google Cloud Storage (`gs://`) was supported via a custom `GCSReadSeekCloser`.  
Now, a **modular reader architecture** provides unified support for:
- Local files  
- Google Cloud Storage (GCS)  
- Amazon S3  

Key updates:
- Added `utils/readers` package containing:
  - `registry.go` — manages reader registration and lookup  
  - `local.go` — handles local file reading  
  - `gcs.go` — refactored GCS reader (replaces old `gcs_read_seek_closer.go`)  
  - `s3.go` — new AWS S3 reader using AWS SDK v2  
- Updated `utils.go` to delegate all reading logic to the new registry  
- Removed the legacy GCS-only implementation  
- Added AWS SDK v2 dependencies

### DIFF
--- a/website/docs/guides/cli.md
+++ b/website/docs/guides/cli.md
@@ -137,7 +137,7 @@ Echo messages for a specific topic to stdout as JSON:
 
 ### Remote file support
 
-All commands except `convert` support reading from remote files stored in **GCS** and **S3**:
+All commands except `convert` support reading from remote files stored in **GCS** or **S3**:
 
 <!-- cspell: disable -->
 


go/cli/mcap/utils/readers/registry.go
```diff
@@ -0,0 +1,45 @@
...
+
+var factories = map[string]ReaderFactory{}
+
+// RegisterReader registers a reader factory for a given scheme.
+func RegisterReader(scheme string, factory ReaderFactory) {
+	factories[scheme] = factory
+}
+
+// GetReader returns an io.ReadSeekCloser for a given filename.
+func GetReader(ctx context.Context, scheme, bucket, path string) (func() error, io.ReadSeekCloser, error) {
```
bennetthardwick, line 21: nit: it's a little confusing swapping the order of these vs the `ReaderFactory`

bennetthardwick:
@asokharev testing with s3 I noticed that I have to set `AWS_REGION` to get it to work. Doing `aws s3 cp` doesn't require setting the region so I think it'd be good to have parity there (not sure how they do it).

I also needed to be authenticated. If I try and access a public bucket it will time out with the following message:

```
Failed to read file <file>: failed to create S3 reader: failed to head S3 object: operation error S3: HeadObject, get identity: get credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, request canceled, context deadline exceeded
```

The AWS CLI doesn't seem to support copying from public buckets so I think it is fine not to support this (if it is too hard), but I think a more friendly error message would be great at least.

Everything else looks good!

website/docs/guides/cli.md
```diff
@@ -137,7 +137,7 @@ Echo messages for a specific topic to stdout as JSON:
 
 ### Remote file support
 
-All commands except `convert` support reading from remote files stored in GCS:
+All commands except `convert` support reading from remote files stored in **GCS** and **S3**:
```
bennetthardwick:
```suggestion
All commands except `convert` support reading from remote files stored in **GCS** or **S3**:
```

website/docs/guides/cli.md
```diff
@@ -162,6 +162,17 @@ All commands except `convert` support reading from remote files stored in GCS:
...
 
+S3 support now includes **public buckets**.  
+When accessing S3 objects, the program first attempts an **anonymous connection**.  
+If that fails (for example, for private buckets), it will automatically try to use **local AWS credentials** found on your system (as configured for the AWS CLI).
+
+For public buckets, you still need to specify the correct AWS region.  
+Example:
+
+```bash
+AWS_REGION=eu-north-1 mcap info s3://my-public-bucket/demo.mcap
```
bennetthardwick: To keep this evergreen I think it'd be better just to say something simple like this:

```suggestion
#### AWS S3

When reading from S3 you must specify the region of the bucket:

```bash
AWS_REGION=eu-north-1 mcap info s3://my-public-bucket/demo.mcap
```

bennetthardwick (approved):

PR #1473, opened 2025-10-22
Clarify schema and message encodings

Muon:
### Changelog

- Clarified definitions of `ros1msg`, `ros2msg` and `ros2idl` schema encoding and `cdr` message encoding.

### Docs
None

### Description

The `cdr` message encoding linked to the wrong document and didn't actually explain that it was specifically XCDR1/XCDR2 that it contained.

The `ros1msg` schema encoding (and by extension `ros2msg`) left open a possible ambiguity with multiple messages sharing the same combination of package and message name. This is now forbidden. 

The `ros2idl` schema encoding did not specify how it was to be interpreted. I have not been able to find any official ROS2 documentation about this, so I have added what I believe to be the most reasonable interpretation. 

Fixes: FIRE-97
Fixes: FIRE-180

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -49,6 +49,7 @@ words:
   - crc
   - crcs
   - datetime
+  - DDSI
   - deserialization
   - distro
   - elems
@@ -90,6 +91,7 @@ words:
   - rosbag
   - rosmsg
   - rostime
+  - RTPS
   - Saronic
   - schemaless
   - serde


website/docs/spec/registry.md
```diff
@@ -122,22 +130,22 @@ The `omgidl` schema encoding stores [OMG IDL](https://www.omg.org/spec/IDL/4.2/A
...
 
 ```bash
-# -x c: treat input as C to avoid guessing based on the extension.
+# -x c++: treat input as C++ to avoid guessing based on the extension.
 # -E: run only the preprocessor.
 # -P: remove all #-directives from the output.
 # Run from the root of your IDL definition tree, or change the `-I` flag to the
 # correct directory.
-$ cc -x c -E -P -I . top_level_module/my_module/MyType.idl > combined.idl
+$ cc -x c++ -E -P -I . top_level_module/my_module/MyType.idl > combined.idl
```
clalancette, line 148: I don't have any particular objection to treating these includes as C++ rather than C, I'm just curious why the change was made?

clalancette (commented):
This seems like reasonable clarifications to me.  I left one question inline.

website/docs/spec/registry.md
james-rms: I can take this one: [IDL spec section 7.3](https://www.omg.org/spec/IDL/4.2/PDF) starts with:
> IDL shall be preprocessed according to the specification of the preprocessor in ISO/IEC 14882:2003. The preprocessor may be implemented as a separate process or built into the IDL compiler.

ISO/IEC 14882:2003 is the C++ standard, not the C standard.

clalancette: Makes sense, thanks for the clarification.

PR #1480, opened 2025-11-04
rust: fix int underflow when chunk wasn't closed

### LINKED ISSUE #1479: rust: Integer overflow with the linear reader
**Description**

- Version: 0.23.3
- Platform: x86 linux

The StreamedReader panics in debug mode due to integer overflow [here](https://sourcegraph.com/github.com/foxglove/mcap@c7ae597cf60e56ed8858fbae248f8ab59f185399/-/blob/rust/src/sans_io/linear_reader.rs?L513). The `len` and `header.compressed_size` variables have the u64 max value of `18446744073709551615` when the reader is in this state (verified in debugger). The cart runs off the tracks as early as [here](https://sourcegraph.com/github.com/foxglove/mcap@c7ae597cf60e56ed8858fbae248f8ab59f185399/-/blob/rust/src/sans_io/linear_reader.rs?L410-413).

**Steps To Reproduce**

Added a test in my fork [here](https://github.com/nionata/mcap/commit/23bfa39b95a581c85ce16eb7f3d768cd86746f32).

```rust
#[test]
#[cfg_attr(debug_assertions, should_panic)]
fn bad_reader() {
    let tmpdir = tempfile::tempdir().unwrap();
    let input_path = tmpdir.path().join("input.mcap");

    let file = std::fs::File::create(&input_path).unwrap();
    assert!(input_path.exists());
    let writer = std::io::BufWriter::new(file);
    let mut writer = mcap::WriteOptions::new().create(writer).unwrap();

    let schema_id = writer.add_schema("my-schema", "", &[]).unwrap();
    let _channel_id = writer
        .add_channel(
            schema_id,
            "my-topic",
            "json",
            &std::collections::BTreeMap::new(),
        )
        .unwrap();

    let fd = std::fs::File::open(input_path).unwrap();
    let mapped = unsafe { memmap2::Mmap::map(&fd) }.unwrap();

    let mut reader = mcap::MessageStream::new(&mapped).unwrap();

    // There is an overflow bug in the MCAP reader library. This will panic when compiled in debug mode.
    // https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-overflow
    let result = reader.next();

    // The overflow will saturate at the u64 maximum value when compiled in release mode.
    // https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=f6143a79d6ec7aa8eb0badc56569ed09
    // This results in the McapError::UnexpectedEof error being returned.
    let mcap_result = result.unwrap().unwrap_err();
    match mcap_result {
        mcap::McapError::UnexpectedEof => {}
        other => panic!("Unexpected error: {:?}", other),
    }
}
```

**Expected Behavior**

`reader.next` returns None when there is no new data to read. Or at the very least `UnexectedEof`, but certainly not an int overflow.

bennetthardwick:
### Changelog

- rust: fix int underflow on chunk with invalid size

### Docs

None

### Description

When a chunk doesn't close correctly (loss of power, crash, etc.) the size that was written will be `u64::MAX`. If you try and read a file like this with the Rust reader it breaks because the size of the chunk is expected to be valid. In debug mode this will panic due to underflow, in release mode this value will wrap and break something else down the track.

This change explicitly uses checked calculations for the chunk size and returns an `UnexpectedEoc` error if any of the calculations overflow.

Fixes #1479.
Fixes FIRE-202.

### DIFF
--- a/rust/src/sans_io/linear_reader.rs
+++ b/rust/src/sans_io/linear_reader.rs
@@ -509,8 +509,15 @@ impl LinearReader {
                         &mut self.decompressors,
                         &header.compression
                     ));
+
+                    let chunk_data_len = check!(len
+                        .checked_sub(header_len as u64)
+                        .ok_or(McapError::UnexpectedEoc));
+
                     let padding_after_compressed_data = check!(check_len(
-                        len - (header_len as u64) - header.compressed_size,
+                        check!(chunk_data_len
+                            .checked_sub(header.compressed_size)
+                            .ok_or(McapError::UnexpectedEoc)),
                         self.options.record_length_limit
                     )
                     .ok_or(McapError::ChunkTooLarge(len)));
@@ -706,6 +713,7 @@ impl LinearReader {
 }
 
 /// Events emitted by the linear reader.
+#[derive(Debug)]
 pub enum LinearReadEvent<'a> {
     /// The reader needs more data to provide the next record. Call [`LinearReader::insert`] then
     /// [`LinearReader::notify_read`] to load more data. The value provided here is a hint for how
@@ -985,6 +993,7 @@ mod tests {
         );
         Ok(())
     }
+    use assert_matches::assert_matches;
     use paste::paste;
 
     macro_rules! test_chunked_parametrized {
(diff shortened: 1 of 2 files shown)


rust/src/sans_io/linear_reader.rs
```diff
@@ -1206,4 +1215,31 @@ mod tests {
...
+            .expect("first record should be header");
+
+        let LinearReadEvent::Record { opcode: 1, .. } = next else {
+            panic!("expected first record to be header");
+        };
+
+        let next = reader.next_event().expect("there should be one event");
+
+        // test fails with unexpected EOC because sizes are u64::max
+        assert_matches!(next, Err(McapError::UnexpectedEoc));
```
nionata, line 1243: Say the file writer flushed the chunk after this call. Would `reader.next_event` yield data? Or is the reader borked after an error like this?

bennetthardwick: In general once the reader hits an error it is considered to be in an invalid state. So it will be borked after this.

jtbandes (approved):

james-rms (approved):

PR #1482, opened 2025-11-06
cpp: Remove redundant close calls in McapWriter

Junxia0Wang:
### Changelog

Remove redundant close calls in McapWriter

### Description

Each open method in McapWriter calls close, but in fact, the method that is ultimately called is the open method that takes IWritable as a parameter.

</td><td>

</td></tr></table>

### DIFF
--- a/cpp/mcap/include/mcap/writer.inl
+++ b/cpp/mcap/include/mcap/writer.inl
@@ -341,7 +341,6 @@ void McapWriter::open(IWritable& writer, const McapWriterOptions& options) {
 
 Status McapWriter::open(const std::string_view filename, const McapWriterOptions& options) {
   // If the writer was opened, close it first
-  close();
   fileOutput_ = std::make_unique<FileWriter>();
   const auto status = fileOutput_->open(filename);
   if (!status.ok()) {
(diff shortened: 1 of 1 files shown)


cpp/mcap/include/mcap/writer.inl
```diff
@@ -341,7 +341,6 @@ void McapWriter::open(IWritable& writer, const McapWriterOptions& options) {
 
 Status McapWriter::open(const std::string_view filename, const McapWriterOptions& options) {
   // If the writer was opened, close it first
```
jtbandes, line 343: These comments are now out of date

jtbandes (commented):
Thanks for the contribution! 

It looks like these close() calls were added in https://github.com/foxglove/mcap/pull/1381 by @eloff. I'm not sure but I imagine that it's intended to make sure calling open() + open() twice in a row doesn't result in a truncated file from the first writer. Do you know if that behavior still works properly with this change? What was the issue that led you to propose this change?

We should probably add something to the [unit tests](https://github.com/foxglove/mcap/blob/main/cpp/test/unit_tests.cpp) to cover this behavior. It looks like the tests were updated in https://github.com/foxglove/mcap/pull/1381 but it doesn't look like they covered the double-open behavior.

PR #1483, opened 2025-11-10
go: fix cli recover command error messages

bennetthardwick:
### Changelog

- Fixes the mcap recover command printing to stdout
- Fixes incorrect error message

### Docs

None

### Description

Currently the MCAP CLI prints error messages to stdout even when it is being used to write the "recovered" file. This leads to the file in some cases having error messages interleaved throughout it, which creates an invalid this.

This is only an issue when redirecting stdout to a file, not when the `--output` flag is used.

### DIFF
--- a/go/cli/mcap/cmd/recover.go
+++ b/go/cli/mcap/cmd/recover.go
@@ -56,11 +56,10 @@ func recoverRun(
 
 		err := mcapWriter.Close()
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to close mcap writer: %v\n", err)
+			utils.EprintF("failed to close mcap writer: %v\n", err)
 			return
 		}
-		fmt.Fprintf(
-			os.Stderr,
+		utils.EprintF(
 			"Recovered %d messages, %d attachments, and %d metadata records.\n",
 			mcapWriter.Statistics.MessageCount,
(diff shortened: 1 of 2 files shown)


go/cli/mcap/cmd/recover.go
```diff
@@ -157,14 +157,15 @@ func recoverRun(
...
 			if decodeChunk {
 				idx, err := utils.UpdateInfoFromChunk(info, chunk, nil)
 				if err != nil {
-					fmt.Printf("Failed to update info from chunk, skipping: %s\n", err)
+					fmt.Fprintf(os.Stderr, "Failed to update info from chunk, skipping: %s\n", err)
 				} else {
 					err = mcapWriter.WriteChunkWithIndexes(chunk, idx)
 					if err != nil {
-						fmt.Printf("Failed to write chunk, skipping: %s\n", err)
+						fmt.Fprintf(os.Stderr, "Failed to write chunk, skipping: %s\n", err)
```
james-rms: i think it's worth making and using a helper function for these cases.

bennetthardwick: @james-rms created a helper called `utils.EprintF` and `utils.Eprintln`

james-rms (approved):

PR #1491, opened 2025-12-02
rust: fix notifying EOF after end magic has been parsed

Muon:
### Changelog
- rust: fix error when notifying `LinearReader` of EOF after end magic has been parsed.
- rust: add `LinearReader` option to check for trailing garbage bytes after end magic.

### Docs
None

### Description

When `skip_end_magic` is false, `LinearReader::next_event()` currently returns an error when EOF is reached even if the end magic has been seen. This PR makes it return `None` instead.

An option has also been added to check that there are no trailing bytes after the end magic. Users who care about the end magic being present probably also care that there is nothing after it, though it is not on by default because that would be a breaking change.

The documentation for `skip_end_magic` has been updated to clarify that it is about expecting the end magic after the footer record, not at the end of the stream.

Unit tests have been added to cover this behavior, and I have verified that they fail with the old code.

As part of this change, `RwBuf`'s fields have been made private to protect its invariants from accidental violations. This bug was uncovered because some code I was testing found an underflow panic due to the `end - start` calculation in `tail_with_size()` (which should never happen with `LinearReader`), so as part of trying to localize that I made `RwBuf` more strictly encapsulated. I haven't yet been able to reproduce that panic in tests, but I've left in the associated changes since I think they're probably an improvement.

There are still some unfixed problems with `LinearReader`'s EOF handling, but those require deeper restructuring. Currently it eagerly fails with `UnexpectedEof` if anything has been buffered and EOF has been notified, even if it could still finish parsing the file with what has been supplied. A naive attempt by deleting the top-level EOF check and putting EOF checks into `load!()` and `consume!()` causes tests to fail, which I'm not yet sure whether it indicates a bug.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -7,7 +7,7 @@ categories = [ "science::robotics", "compression" ]
 repository = "https://github.com/foxglove/mcap"
 documentation = "https://docs.rs/mcap"
 readme = "README.md"
-version = "0.23.4"
+version = "0.24.0"
 edition = "2021"
 license = "MIT"
 


rust/src/sans_io/linear_reader.rs
```diff
@@ -60,76 +61,133 @@ struct ChunkState {
...
 
-/// A private struct that encapsulates a buffer with start and end cursors.
-#[derive(Default)]
-struct RwBuf {
-    data: Vec<u8>,
-    start: usize,
-    end: usize,
-    hasher: Option<crc32fast::Hasher>,
-}
+mod rw_buf {
```
james-rms, line 64: what's with the mod here?

Muon: To enforce the privacy of `RwBuf`'s fields, in order to protect `RwBuf`'s invariants from the outer code.

Issue #1504, opened 2025-12-16
`mcap info` frequency estimation appears to be broken


joe-saronic:
**Description**

- Version: 0.0.51
- Platform: Ubuntu

**Steps To Reproduce**

Generate an MCAP with two frames. The example here uses python for brevity, but it doesn't matter.

```python
from io import BytesIO
from time import time_ns
import numpy as np
from foxglove_schemas_protobuf.CompressedImage_pb2 import CompressedImage
from google.protobuf.timestamp_pb2 import Timestamp
from mcap_protobuf.writer import Writer
from PIL import Image

count = 2
period_nanos = 100_000_000
width = 1080
height = 920

buf = BytesIO()
Image.new("L", (width, height), color=0).save(buf, format="png", optimize=True)
png_bytes = buf.getvalue()

with open("test.mcap", "wb") as file:
    writer = Writer(file)
    nano_time = time_ns()
    for i in range(count):
        timestamp = Timestamp(seconds=nano_time // 1_000_000_000, nanos=nano_time % 1_000_000_000)
        for topic in ["video1"]:
            msg = CompressedImage(timestamp=timestamp, format="png", data=png_bytes)
            writer.write_message(topic=topic, message=msg, log_time=nano_time, publish_time=nano_time)
            nano_time += period_nanos
writer.finish()
```

Running `info` on the file shows:

```
$ mcap info test.mcap 
library:   python mcap-protobuf-support 0.5.1; mcap 1.1.1              
profile:                                                               
messages:  2                                                           
duration:  100ms                                                       
start:     2025-12-15T19:51:30.520928453-06:00 (1765849890.520928453)  
end:       2025-12-15T19:51:30.620928453-06:00 (1765849890.620928453)  
compression:
	zstd: [1/1 chunks] [2.68 KiB/496.00 B (81.94%)] [4.84 KiB/sec] 
channels:
	(1) video1  2 msgs (20.00 Hz)   : foxglove.CompressedImage [protobuf]  
channels: 1
attachments: 0
metadata: 0
```

**Expected Behavior**

Frequency of messages 100ms apart should be 10Hz = (2msgs - 1) / 100ms, but it looks like the calculation is 2msgs / 100ms.

Similar behavior if we increase `count` in the code above to `10`:

```
$ mcap info test.mcap 
library:   python mcap-protobuf-support 0.5.1; mcap 1.1.1              
profile:                                                               
messages:  10                                                          
duration:  900ms                                                       
[...]

PR #1505, opened 2025-12-16
`mcap info`: fix frequency estimation

joe-saronic:
### Changelog
Frequency of MCAP channels is displayed as `(n - 1) / dt` instead of current `n / dt`.

### Docs

None

### Description

Fixes foxglove/mcap#1504, which also describes the problem. Another confirmation that this is the consistent way to define frequency is in the example here: https://github.com/foxglove/mcap/blob/main/typescript/examples/flatbufferswriter/scripts/main.ts#L198.

Using the same 2-message MCAP as in the issue:

```
$ mcap info test.mcap 
library:   python mcap-protobuf-support 0.5.1; mcap 1.1.1              
profile:                                                               
messages:  2                                                           
duration:  100ms                                                       
start:     2025-12-15T19:51:30.520928453-06:00 (1765849890.520928453)  
end:       2025-12-15T19:51:30.620928453-06:00 (1765849890.620928453)  
compression:
	zstd: [1/1 chunks] [2.68 KiB/496.00 B (81.94%)] [4.84 KiB/sec] 
channels:
	(1) video1  2 msgs (20.00 Hz)   : foxglove.CompressedImage [protobuf]  
channels: 1
attachments: 0
metadata: 0
```

</td><td>

```
$ mcap info test.mcap 
library:   python mcap-protobuf-support 0.5.1; mcap 1.1.1              
profile:                                                               
messages:  2                                                           
duration:  100ms                                                       
start:     2025-12-15T19:51:30.520928453-06:00 (1765849890.520928453)  
end:       2025-12-15T19:51:30.620928453-06:00 (1765849890.620928453)  
compression:
	zstd: [1/1 chunks] [2.68 KiB/496.00 B (81.94%)] [4.84 KiB/sec] 
channels:
	(1) video1  2 msgs (10.00 Hz)   : foxglove.CompressedImage [protobuf]  
channels: 1
attachments: 0
metadata: 0
```

</td></tr></table>

Also ensuring that the one-frame case does no cause a crash with `count=1`:

```
$ mcap info test.mcap 
library:   python mcap-protobuf-support 0.5.1; mcap 1.1.1              
profile:                                                               
messages:  1                                                           
duration:  0s                                                          
start:     2025-12-16T13:14:39.753900004-06:00 (1765912479.753900004)  
end:       2025-12-16T13:14:39.753900004-06:00 (1765912479.753900004)  
compression:
	zstd: [1/1 chunks] [1.61 KiB/477.00 B (71.07%)] 
channels:
	(1) video1  1 msgs   : foxglove.CompressedImage [protobuf]  
channels: 1
attachments: 0
metadata: 0
```

</td><td>

```
$ ~/ots/mcap/go/cli/mcap/bin/mcap info test.mcap 
library:   python mcap-protobuf-support 0.5.1; mcap 1.1.1              
profile:                                                               
messages:  1                                                           
duration:  0s                                                          
start:     2025-12-16T13:14:39.753900004-06:00 (1765912479.753900004)  
end:       2025-12-16T13:14:39.753900004-06:00 (1765912479.753900004)  
compression:
	zstd: [1/1 chunks] [1.61 KiB/477.00 B (71.07%)] 
chunks:
	max uncompressed size: 1.61 KiB
	max compressed size: 477.00 B
	overlaps: no
channels:
	(1) video1  1 msgs   : foxglove.CompressedImage [protobuf]  
channels: 1
attachments: 0
metadata: 0
```

</table>

---

> [!NOTE]
> Compute per-channel min/max message times from indexes and use them to display channel frequency as (n-1)/dt in `mcap info`.
> 
> - **CLI (`mcap info`)**:
>   - Update channel frequency display to use `(n-1)/dt` based on per-channel start/end times; refactor message string assembly in `go/cli/mcap/cmd/info.go`.
> - **Core/Reader**:
>   - Extend `mcap.Info` with `ChannelMessageStartTimes` and `ChannelMessageEndTimes`.
>   - Populate these in `Reader.Info()` via new `getChannelTimings` that scans chunk `MessageIndex` records to find per-channel min/max timestamps.
>   - Plumb new maps when constructing `Info`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dea4e437aad1ee04327029091f188d9961699dc5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/go/mcap/reader.go
+++ b/go/mcap/reader.go
@@ -230,13 +230,21 @@ func (r *Reader) getChannelTimings(info *Info) error {
 				return err
 			}
 			if len(messageIndex.Records) > 0 {
-				firstTime := messageIndex.Records[0].Timestamp
-				lastTime := messageIndex.Records[len(messageIndex.Records)-1].Timestamp
-				if startTime, exists := info.ChannelMessageStartTimes[chanID]; !exists || firstTime < startTime {
-					info.ChannelMessageStartTimes[chanID] = firstTime
+				minTime := uint64(math.MaxUint64)
+				maxTime := uint64(0)
+				for _, record := range messageIndex.Records {
+					if record.Timestamp < minTime {
+						minTime = record.Timestamp
+					}
+					if record.Timestamp > maxTime {
+						maxTime = record.Timestamp
+					}
 				}
-				if endTime, exists := info.ChannelMessageEndTimes[chanID]; !exists || lastTime > endTime {
-					info.ChannelMessageEndTimes[chanID] = lastTime
+				if startTime, exists := info.ChannelMessageStartTimes[chanID]; !exists || minTime < startTime {
+					info.ChannelMessageStartTimes[chanID] = minTime
+				}
+				if endTime, exists := info.ChannelMessageEndTimes[chanID]; !exists || maxTime > endTime {
+					info.ChannelMessageEndTimes[chanID] = maxTime
 				}
 			}
 		}


jtbandes (commented):
While we are here, I'm not sure if `end-start` being the start/end of the whole file, vs just the start/end times of _this channel_ (or topic), is more correct? It appears the Foxglove app uses the start/end times of the topic.

joe-saronic:
@james-rms I do check `if channelMessageCount > 1` and subsequently `chanEnd > chanStart` to make sure both quantities are non-zero, so not sure how that can trigger an error, but I may very well have missed something.

Assuming that channels contain messages sampled at a more-or-less constant frequency, using the MCAP length, especially for short clips is probably not a good idea, since we don't know how the file time-span maps to the message period at all. I would argue that the channel duration is a much more accurate basis for the estimator at least in that case. For sufficiently long durations, even the current estimator will converge to the correct value, so that's not saying much.

james-rms:
The point of the `info` command is to provide fast summary information that can be derived by reading the MCAP index section. If the reader needs to seek all through the file to provide meaningful statistics, that changes the set of situations where it's appropriate to use.

> Assuming that channels contain messages sampled at a more-or-less constant frequency

Your case https://github.com/foxglove/mcap/issues/1504 describes a situation but describes the current calculation as "wrong". It's only wrong if the tool is meant to assume that messages are always published on channels at a constant frequency. I don't think that's the right assumption to make in all cases.

If you use a weaker assumption that messages times are the result of a poisson process, then the existing calculation k/T is a correct estimate of the maximum likelihood of frequency. However, since there are only two samples, the estimate has low confidence. I think what the tool is missing today is an estimate of that confidence value.

joe-saronic:
Superseded by #1506

PR #1516, opened 2026-01-08
rust: add inline documentation for record sizes and offsets

oliverdaff:
## Changelog

None (readability-only changes, no API or behaviour changes)

## Docs

None

## Description

Improves readability of MCAP Rust parsing and writing code by documenting record layouts and fixed-size calculations inline, without introducing new abstractions or changing behaviour.

Some fixed-size calculations rely on numeric literals that are correct but non-obvious at the call site. This PR replaces those with annotated sums so the structure of the underlying record is immediately clear.

Example:

```rust
impl MessageHeader {
    pub(crate) fn serialized_len(&self) -> u64 {
        2  // channel ID
        + 4  // sequence
        + 8  // log time
        + 8  // publish time
    }
}
```

**Changes**:

* Added inline comments explaining record field sizes and offset arithmetic in readers and writers
* Clarified fixed-size calculations for chunk, message, footer, and index handling
* Updated MCAP spec reference to [https://mcap.dev/spec](https://mcap.dev/spec)

**Testing**: All existing tests pass with no behaviour changes.

**Impact**: Makes MCAP binary parsing and writing logic easier to audit and safer to modify, while preserving existing structure and behaviour.

---

> [!NOTE]
> Improves readability and maintainability by documenting fixed-size/offset math and using small constants where helpful, without changing behavior.
> 
> - Inline comments/clarified arithmetic for `9`-byte opcode+length prefixes, message index minimum size, and chunk header fields in `read.rs`, `write.rs`, and `sans_io/*`
> - Footer parsing clarified to slice fixed 20-byte body; adds note about trailing 8-byte end magic
> - Introduces small constants (e.g., `MIN_MESSAGE_INDEX_BYTES`, `MIN_CHUNK_HEADER_SIZE`) and uses them in bounds checks
> - Updates MCAP spec reference to `https://mcap.dev/spec`
> - Extends `.typos.toml` identifiers with `bimap`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6d00185eddcbbccab7825d4598c576ac1110fc81. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/rust/src/records.rs
+++ b/rust/src/records.rs
@@ -35,293 +35,6 @@ pub mod op {
     pub const DATA_END: u8 = 0x0F;
 }
 
-/// Size constants for MCAP record format fields.
-///
-/// These constants define the byte sizes of fields in the MCAP binary format,
-/// as specified in the MCAP specification:
-/// <https://mcap.dev/spec>
-///
-/// Using named constants instead of magic numbers improves code clarity,
-/// maintainability, and helps prevent errors when parsing binary data.
-pub(crate) mod sizes {
-    /// Size of the opcode field in bytes (1 byte for all records)
-    pub const OPCODE: usize = 1;
-
-    /// Size of the record length field in bytes (u64 = 8 bytes)
-    pub const RECORD_LENGTH: usize = 8;
-
-    /// Combined size of opcode + record length header (9 bytes total)
-    ///
-    /// Every MCAP record starts with this 9-byte header:
-    /// - 1 byte opcode (which record type)
-    /// - 8 byte length (size of the record body)
-    pub const OPCODE_AND_LENGTH: usize = OPCODE + RECORD_LENGTH;
-
-    /// Message header field sizes (MessageHeader struct)
-    pub mod message {
-        /// Channel ID field size (u16 = 2 bytes)
-        pub const CHANNEL_ID: usize = 2;
-
-        /// Sequence number field size (u32 = 4 bytes)
-        pub const SEQUENCE: usize = 4;
-
-        /// Log time field size (u64 = 8 bytes)
-        pub const LOG_TIME: usize = 8;
-
-        /// Publish time field size (u64 = 8 bytes)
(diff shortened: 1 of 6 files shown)


rust/src/records.rs
```diff
@@ -35,6 +35,293 @@ pub mod op {
     pub const DATA_END: u8 = 0x0F;
 }
 
+/// Size constants for MCAP record format fields.
+///
+/// These constants define the byte sizes of fields in the MCAP binary format,
+/// as specified in the MCAP specification:
+/// <https://mcap.dev/spec>
+///
+/// Using named constants instead of magic numbers improves code clarity,
```
james-rms: This does not belong in the docstring, unless it helps the user understand what they're looking at or how they'd use it. This might be better placed in the PR description to argue for  the change.

rust/src/records.rs
```diff
@@ -35,6 +35,293 @@ pub mod op {
...
+                encoding::STRING_LENGTH + encoding::STRING_LENGTH
+            );
+        }
+
+        #[test]
+        fn test_mcap_format_spec_compliance() {
+            // Document and verify expected sizes per MCAP specification
+            // https://mcap.dev/spec
+            assert_eq!(OPCODE, 1, "MCAP spec: opcode is 1 byte");
+            assert_eq!(
```
james-rms: This seems really low value to me. Tests should _increase certainty_ about the behavior of they're testing. I don't need a test to show me that OPCODE_LENGTH is defined as 1 byte, I can just look at its definition for that. Tests have a real cost in terms of what a programmer has to understand and modify to get their job done, they need to pay that cost with value.

oliverdaff:
Updated the PR text to remove the shared constants refactor and instead document offsets and record layouts inline at the call sites, keeping the change readability-only and behaviour-neutral.

james-rms (approved):

PR #1524, opened 2026-02-03
Align build system with Conan 2.x

### LINKED ISSUE #1489: [cpp] Library does not build with Conan 2
**Description**

The build scripts use Conan 1 commands, which are not compatible with Conan 2. Conan 1 is deprecated and more difficult to install (`pip install conan` currently gets you Conan 2), so we should update.

Ext3h:
### Changelog

Support for building and consuming `mcap` from Conan 2.x as well as standard CMake based build systems.

Enable consumption of `mcap` as a true header-only library without linker conflicts.
### Description

The existing Conan recipes had a couple of severe issues:
- Incompatible with Conan 2.x due to the use of deprecated features in recipes.
  - Now compatible with both Conan 1.x and Conan 2.x.
- The package could never produce a deployable Conan package, was only used with `conan editable`.
  - `conan editable` still works (with Conan 2.x due to bugs in the Conan 1.x layout system), but formost `conan export` and `conan create` workflows are now supported.
- Configuration options were not mapped to Conan options. 
  - `with_lz4` and `with_zstd` are now exposed as Conan options.
- Upstream library versions were tied. This resulted in having to rebuild i.e. `Protobuf`, `abseil` and alike as there were no more ready-built upstream packages.
  - Relaxed version ranges. This increases the chances of hitting a version for which ready-made builds exist upstream in conan-center by automatically following stable latest. Note that this will only apply once switching to Conan 2, as the Conan 1 mirror of conancenter no longer has pre-built binaries.
- Shared and static library build types were not usable via Conan.
  - `shared` and `header_only` options are exposed as Conan options.
  - This also configures defines correctly, so that the user will not require any code changes to switch library types.
- The build system was based on early Conan 1.x commands that are pretty much all deprecated.
  - Switched  build system workflow to Commands that are mostly inter-exchangeable between Conan 1.x and Conan 2.x.
  - Added test coverage for the different supported linkage types.

The changes to the Conan recipe are also mirrored in https://github.com/conan-io/conan-center-index/pull/29492 - so the 3rd party Conan package which your average user will actually first be in contact with is - in terms of usage - aligned with the in-development source tree the user can find in this repository.

The formerly external CMake build wrapper has been ported over and integrated.

</td><td>

</td></tr></table>

Fixes: foxglove/mcap#1489 https://github.com/foxglove/mcap/discussions/1457


cpp/examples/protobuf/BuildFileDescriptorSet.cpp
```diff
@@ -19,8 +19,7 @@ google::protobuf::FileDescriptorSet BuildFileDescriptorSet(
     next->CopyTo(fdSet.add_file());
     for (int i = 0; i < next->dependency_count(); ++i) {
       const auto& dep = next->dependency(i);
-      if (seenDependencies.find(dep->name()) == seenDependencies.end()) {
-        seenDependencies.insert(dep->name());
+      if (seenDependencies.emplace(dep->name()).second) {
```
Ext3h, line 22: `dep->name()` became `std::string_view` in some Protobuf update, so the previous code could only work in C++23 mode, but not C++17.

cpp/examples/protobuf/CMakeLists.txt
```diff
@@ -1,57 +1,40 @@
...
+    PROTOC_OUT_DIR ${CMAKE_CURRENT_BINARY_DIR}/autogenerated)
+target_link_libraries(foxglove_proto_lib PUBLIC protobuf::libprotobuf)
 target_include_directories(foxglove_proto_lib PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/autogenerated)
-# Unfortunately the protobuf library headers include some unused parameters, which have to be
-# downgraded back to warnings for all targets that depends on them.
+# Unfortunately the protobuf library headers include some unused parameters, shadowing of members and missed constexpr,
+# which have to be downgraded back to warnings for all targets that depends on them.
 if (MSVC)
-    target_compile_options(foxglove_proto_lib PUBLIC /wd4100 /wd4267 /wd4244)
+    target_compile_options(foxglove_proto_lib PUBLIC /wd4100 /wd4267 /wd4244 /wd4458 /wd4127)
```
Ext3h, line 26: These are not new warnings introduced by dependency updates, but rather by newer compiler versions.

cpp/examples/conanfile.py
```diff
@@ -1,17 +1,33 @@
...
-        "nlohmann_json/3.10.5",
-        "catch2/2.13.8",
-    ]
+    test_type = "explicit"
+
+    def requirements(self):
+        self.requires(self.tested_reference_str)
+        self.requires("protobuf/[>=3.21.1]")
+        self.requires("nlohmann_json/[>=3.10.5]")
+        self.requires("catch2/[>=2.13.8 <3.0]")
```
Ext3h, line 15: Catch2 in >= 3.0 had a break in source code compatbility.

cpp/mcap/include/mcap/reader.inl
```diff
@@ -11,20 +11,20 @@
 
 namespace mcap {
 
-bool CompareChunkIndexes(const ChunkIndex& a, const ChunkIndex& b) {
+MCAP_PUBLIC_INLINE bool CompareChunkIndexes(const ChunkIndex& a, const ChunkIndex& b) {
```
Ext3h: Same change to every symbol the `.inl` files.

Alternative is to conditionally inject an anonymous namespace around `namespace mcap` - but that would break type compatibility between translation units for real header-only mode.

cpp/mcap/conanfile.py
```diff
@@ -1,28 +1,137 @@
...
-    version = "2.1.2"
-    url = "https://github.com/foxglove/mcap"
-    homepage = "https://github.com/foxglove/mcap"
-    description = "A C++ implementation of the MCAP file format"
+    description = (
+        "MCAP is a modular, performant, and serialization-agnostic container file format for pub/sub messages, "
+        "primarily intended for use in robotics applications."
+    )
     license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
```
Ext3h: Whoops, wrong URL.

cpp/mcap/conanfile.py
```diff
@@ -1,28 +1,137 @@
...
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import copy, get, rmdir
+from conan.tools.scm import Version
+import os
 
+required_conan_version = ">=1.52.0"
 
 class McapConan(ConanFile):
     name = "mcap"
-    version = "2.1.2"
```
Ext3h, line 6: The version shouldn't be hardcoded. The version is something you specify when you export the recipe (including its sources) for a specific version.

cpp/build.sh
```diff
@@ -3,18 +3,54 @@
 set -e
 
 conan config init
+# Keep it out of the reserved _/_ user/channel!
+export VERSION=mcap/2.1.2@local/testing
+#TODO For Conan 1.x only. Explicit -pr:b and -pr:h with default can be removed for Conan 2.x.
```
Ext3h, line 9: Feel free to step up to Conan 2.x now.

cpp/README.md
```diff
@@ -91,6 +91,8 @@ If your project does not need `lz4` or `zstd` support, you can optionally disabl
 To simplify installation of dependencies, the [Conan](https://conan.io/) package
```
Ext3h, line 91: The Conan recipe in this repository has only relevance for development on this library.

The standard user will usually use `mcap/*@_/_` - which is maintained by Conan center and comes with the benefit of pre-built binaries.

cpp/README.md
```diff
@@ -125,6 +127,9 @@ before including the library.
 #include <mcap/writer.hpp>
 ```
 
+Full inlining in multiple translation units is supported by defining `MCAP_INLINE_IMPLEMENTATION`.
```
Ext3h: Probably needs better explanation.

cpp/examples/conanfile.py
```diff
@@ -1,17 +1,33 @@
...
+        deps.generate()
 
     def build(self):
         cmake = CMake(self)
         cmake.configure()
         cmake.build()
+
+    def test(self):
+        if can_run(self):
+            bin_path = os.path.join("protobuf", self.cpp.build.bindirs[0], "example_protobuf_unit_tests")
```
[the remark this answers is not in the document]
Ext3h: The content of `self.cpp.build.bindirs[0]` is controlled by `cmake_layout`. They never contain an absolute path, but only a path fragment. Double-check against https://docs.conan.io/2/reference/tools/cmake/cmake_layout.html

cpp/mcap/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+    target_compile_definitions(mcap PUBLIC MCAP_PUBLIC=)
+endif()
+
+
+install(TARGETS mcap
+    EXPORT mcap
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION bin
+    ARCHIVE DESTINATION lib
+    FILE_SET public_headers DESTINATION include
```
Ext3h: Compatibility issue with CMake 3.16, it did not support `FILE_SET` yet.

cpp/run_examples.sh
```diff
@@ -4,6 +4,6 @@ set -e
 
 proto_mcap_filepath=/tmp/example_out.mcap
 
-examples/build/Release/bin/example_protobuf_writer $proto_mcap_filepath
-examples/build/Release/bin/example_protobuf_static_reader $proto_mcap_filepath
-examples/build/Release/bin/example_protobuf_dynamic_reader $proto_mcap_filepath
+examples/build/protobuf/Release/example_protobuf_writer $proto_mcap_filepath
+examples/build/protobuf/Release/example_protobuf_static_reader $proto_mcap_filepath
+examples/build/protobuf/Release/example_protobuf_dynamic_reader $proto_mcap_filepath
```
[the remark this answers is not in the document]
Ext3h: The analysis is wrong. `cmake_layout` puts  the sub directory in front of the build type.

.github/workflows/ci.yml
```diff
@@ -211,7 +211,7 @@ jobs:
           python-version: 3.8
       - run: pip install conan~=1.0
       - run: bash build.sh --build-tests-only
-      - run: ./test/build/Debug/bin/unit-tests
+      - run: ./test/build/Debug/unit-tests
```
Ext3h, line 214: Technically already called as part of the build. Little use in calling it again.

.github/workflows/ci.yml
Ext3h, line 192: All the host tests (`test-host` and `test-examples-host`) were now already run as part of the `conan test` calls performed in the build step.

cpp/run_examples.sh
```diff
@@ -4,6 +4,6 @@ set -e
 
 proto_mcap_filepath=/tmp/example_out.mcap
 
-examples/build/Release/bin/example_protobuf_writer $proto_mcap_filepath
-examples/build/Release/bin/example_protobuf_static_reader $proto_mcap_filepath
-examples/build/Release/bin/example_protobuf_dynamic_reader $proto_mcap_filepath
+examples/build/protobuf/Release/example_protobuf_writer $proto_mcap_filepath
```
Ext3h: This is far too fragile. The specific path varies depending on the used CMake generator. Works just fine ***within*** `conan test`.

Need to debug where this file ended up in the CI.

Ext3h: Looks like a multi-configuration generator results in  `examples/build/protobuf/Release/` while single configuration results in `examples/build/Release/protobuf/`.

Calling via `conan test` doesn't appear to be a sensible strategy for running inside container (that's not part of CI though?...) so hard-cording those paths in single configuration directory convention. Local development with conan tooling is working correctly in either case.

cpp/build.sh
```diff
@@ -3,18 +3,63 @@
...
 
 if [ "$1" != "--build-tests-only" ]; then
-  conan install bench --install-folder bench/build/Release \
-    -s compiler.cppstd=17 -s build_type=Release --build missing
-  conan install examples --install-folder examples/build/Release \
-    -s compiler.cppstd=17 -s build_type=Release --build missing
-  conan build examples --build-folder examples/build/Release
-  conan build bench --build-folder bench/build/Release
+  # Build and run benchmark.
+  rm -rf ./bench/build
```
Ext3h, line 47: Benchmark is not actually supposed to be run as part of CI workflow - is it? Packaging it as a conan test like this is clean though.

james-rms (commented):
This seems like way more diff than I'd expect for a PR that _only_ updates to Conan 2.

cpp/mcap/include/mcap/types.inl
```diff
@@ -2,40 +2,41 @@
 
 namespace mcap {
 
-constexpr std::string_view OpCodeString(OpCode opcode) {
+std::string_view OpCodeString(OpCode opcode) {
```
Ext3h, line 5: Couldn't  be `constexpr` when using multiple translation units. `constrexpr` is always `inline` by definition, and Clang wouldn't keep the symbol.

amacneil:
Fixed in https://github.com/foxglove/mcap/pull/1706

PR #1526, opened 2026-02-12
Fix a problem with schemas without messages.

clalancette:
### Changelog

Channels with no messages will now always have their schemas written out to an mcap file.  This is to ensure that consumers of the mcap know that the channel was subscribed to, even if no messages were received.

### Description

This problem originated in upstream ROS 2 rosbag2, where an attempt to upgrade the vendored version
of mcap from 1.4.0 to 2.1.1 ended up failing an
upstream rosbag2 test.

The failing upstream rosbag2 test registers some
channels with mcap, and then writes messages to
some of them.  It then expects that all of the
schemas are in the resulting mcap, which currently fails.

After investigation of that problem, it turns out
that cf083175e720eea1ff9b7a95fea3b384acc9a807 changed this behavior to only write out schemas that have
had data written to them.

The behavior that rosbag2 expects seems entirely
reasonable; it may have subscribed to topics that
never have data published, but that is good
information to know.  Without having the schema,
it looks like the capture just never subscribed
to those topics.

The fix here is a partial revert of
cf083175e720eea1ff9b7a95fea3b384acc9a807, where we write out all schemas at the end.  This also required fixing the test that was added at the same time, since it was expecting only one or another of the schemas to be written out, which this is now changing.  Finally, I added a new test which ensures that we always get all schemas registered.  This is a bit redundant, but I think a specific test for this case is warranted.

With this in place, the mcap unit tests all pass.  The upstream rosbag2 unit tests now also pass with this in place.  @MichaelOrlov FYI.

Fixes: FIRE-334

### DIFF
--- a/cpp/mcap/include/mcap/writer.inl
+++ b/cpp/mcap/include/mcap/writer.inl
@@ -500,6 +500,8 @@ void McapWriter::terminate() {
   statistics_.channelMessageCounts.clear();
   statistics_.metadataCount = 0;
   statistics_.attachmentCount = 0;
+  statistics_.messageStartTime = 0;
+  statistics_.messageEndTime = 0;
   writtenSchemas_.clear();
   currentMessageIndex_.clear();
   currentChunkStart_ = MaxTime;


cpp/mcap/include/mcap/writer.inl
```diff
@@ -498,7 +495,13 @@ void McapWriter::terminate() {
   attachmentIndex_.clear();
   metadataIndex_.clear();
   chunkIndex_.clear();
-  statistics_ = {};
+  statistics_.messageCount = 0;
```
eloff, line 498: Seems like this could be a clear() method on the stats type

cpp/mcap/include/mcap/writer.inl
```diff
@@ -515,11 +518,13 @@ void McapWriter::terminate() {
 void McapWriter::addSchema(Schema& schema) {
   schema.id = uint16_t(schemas_.size() + 1);
   schemas_.push_back(schema);
+  ++statistics_.schemaCount;
```
eloff, line 521: Nice, better to have it here

eloff (approved):

clalancette:
> I don't see where this changes code to write out an added channel/schema to the chunk - even if there is no message using it. Am I correct to assume the current behavior does write it out - so this PR only needed to change the behavior with the summary section?

This is a very good question.  There are two different places in the code and in the file where schemas/channels are written out:

* In a Data section when messages are written: https://github.com/foxglove/mcap/blob/84e00dd3f1c851d0df85637ea86dd78389837ab9/cpp/mcap/include/mcap/writer.inl#L533-L565
* In a DataEnd/Summary when the mcap is closed: https://github.com/foxglove/mcap/blob/84e00dd3f1c851d0df85637ea86dd78389837ab9/cpp/mcap/include/mcap/writer.inl#L395-L412

This change ensures that the schemas/channels are written to the DataEnd/Summary section.  Schemas/Channels with no messages will *not* be written to the Data section.

This brings up an interesting question.  If the mcap is opened with `options.noSummary`, then we'll run into the same problem as before, where channels with no messages won't be written.  Should we change it so that these channels are always written to the data section?

PR #1536, opened 2026-02-20
Update CI to use Python 3.10

jgblight:
### Changelog
None

### Docs
None

### Description
Python 3.8 has been EOL for over a year and is no longer available via some package managers. This makes it difficult to set up pipenv to develop locally so this PR updates the Pipfile and CI to use 3.10.

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -50,7 +50,7 @@ jobs:
           retention-days: 1
 
   docs-python:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
 


python/mcap/mcap/_message_queue.py
```diff
@@ -47,24 +47,30 @@ def position(self) -> Tuple[int, Optional[int]]:
 
 class _ChunkIndexWrapper(_Orderable):
     def log_time(self) -> int:
-        self.item: ChunkIndex
+        item = cast(ChunkIndex, self.item)
```
jtbandes: wonder if these casts could be avoided by making _Orderable generic?

python/Pipfile
```diff
@@ -36,4 +36,4 @@ types-protobuf = "*"
 isort = "*"
 
 [requires]
-python_version = "3.8"
+python_version = "3.10"
```
jtbandes, line 39: next time we release this package, does this mean it should be a major version bump? ...or maybe this doesn't matter so much if mostly the incompatibilities are in type annotations? (I don't know enough about python packaging...)

.github/workflows/website.yml
```diff
@@ -50,13 +50,13 @@ jobs:
           retention-days: 1
 
   docs-python:
-    runs-on: ubuntu-22.04 # required for python 3.8
+    runs-on: ubuntu-22.04
```
jtbandes: can this now be latest?

python/Pipfile
jgblight: I don't think it's _required_ since we didn't update `setup.cfg` but we may want to update it anyway and bump the version since CI will no longer catch backwards compatibility issues in pythons 3.7-3.9

jtbandes: Oh yeah, my brain thought this was happening inside setup.cfg but I see that is wrong

PR #1598, opened 2026-03-04
feat(merge): add --workers flag for parallel chunk compression/decompression

bobir01:
- Add Workers int to mcap.WriterOptions; wire to zstd.WithEncoderConcurrency(N) for parallel output chunk compression on multi-core machines.

- Add --workers N flag to mcap merge CLI command:
    - workers=1 (default): single-threaded, identical behaviour to before
    - workers=N: spawn N-buffered per-input prefetch goroutines so all input chunk decompressors run concurrently while the main goroutine drives the k-way timestamp-ordered merge via the existing min-heap.
    - workers=0: resolve to runtime.NumCPU() at merge time.

- Add TestMergeParallelWorkers covering workers 1, 2, 4, and 0 to assert correct timestamp ordering is preserved regardless of concurrency.

The k-way merge step itself (priority queue) remains sequential by design; parallelism is in the I/O+decompression pipeline feeding it and in the zstd encoder serving the output.

Example:
    mcap merge --workers 8 -o out.mcap a.mcap b.mcap c.mcap

### Description

</td><td>

</td></tr></table>

### DIFF
--- a/go/cli/mcap/cmd/merge.go
+++ b/go/cli/mcap/cmd/merge.go
@@ -314,6 +314,7 @@ func (m *mcapMerger) mergeInputs(w io.Writer, inputs []namedReader) error {
 	var prefetchCtx context.Context
 	if parallel {
 		prefetchCtx, cancelPrefetch = context.WithCancel(context.Background())
+		defer cancelPrefetch() // ensure cancel on all exit paths (idempotent)
 	}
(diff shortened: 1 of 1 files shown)


bobir01:
@james-rms @gasmith can you check guys, new io buffering here really helped me

PR #1599, opened 2026-03-11
Add a rename channel command to the mcap cli

dakota-foxglove:
### Changelog
Add `--rename-from` / `--rename-to` flags to `mcap filter` for renaming MCAP channel topics.

### Docs
Updated in this PR: `website/docs/guides/cli.md`

### Description
This adds channel topic renaming as flags on the existing `mcap filter` command rather than introducing a standalone `mcap rename` subcommand. Since renaming requires rewriting the entire file, it's a natural fit for `filter`, which already handles that — along with compression options, time-range filtering, and topic inclusion/exclusion.

**New flags:**
- `--rename-from <topic>` — the existing topic name to rename
- `--rename-to <topic>` — the new topic name

**Behavior:**
- Both flags must be specified together
- Fails if `--rename-from` and `--rename-to` are the same
- Fails if the target topic already exists in the file (collision detection)
- Works in both the seekable (indexed) and streaming (lexer) filter paths
- Can be combined with existing filter flags (`--include-topic-regex`, `--start`, `--end`, `--output-compression`, etc.)
- Topic inclusion/exclusion filters operate on original topic names, so `--include-topic-regex` matches the pre-rename name

**Example:**
```bash
# Rename /tf to /tf_renamed
mcap filter demo.mcap -o demo-out.mcap --rename-from /tf --rename-to /tf_renamed

# Verify
mcap list channels demo-out.mcap
#   (2) /tf_renamed  774 msgs (99.48 Hz)   : tf2_msgs/TFMessage [ros1msg]

# Combine with topic filtering
mcap filter demo.mcap -o filtered.mcap --include-topic-regex "/tf" --rename-from /tf --rename-to /tf_renamed
```

Fixes: FG-14035


amacneil:
Don't merge this yet, I'm planning to add it to the new rust cli.

PR #1606, opened 2026-03-20
[WIP][cpp] Add writer option to flush data to disk every N bytes

MichaelOrlov:
### Changelog

Add a new `sizeForFsynch` MCAP writer option to periodically flush file-backed writes to persistent storage after every configured number of bytes.

### Docs

None

### Description

This change adds a new writer option, `sizeForFsynch`, to allow explicit flushing of file-backed MCAP output to disk after every configured number of written bytes.

Today, once data leaves the MCAP library and is written to a file, it can still remain buffered in OS-managed DRAM before it reaches persistent storage. For recording workloads, that creates a durability gap: if the system crashes or loses power, recently written data may still be lost even though it has already passed through the MCAP writer.

This PR adds an opt-in API to reduce that risk. When `sizeForFsynch` is greater than `0`, the file-backed writer tracks how many bytes have been written since the last sync and calls `fsync(fd)` once the configured threshold is met or exceeded. A value of `0` preserves the current behavior and leaves flush timing entirely to the operating system.

The intent is to give users an explicit durability/performance tradeoff. Applications that prioritize throughput can keep the default behavior. Applications that prioritize minimizing data loss during recording can choose a threshold appropriate for their workload and storage characteristics.

This is especially relevant for long-running or safety-critical recordings where losing the last buffered portion of the output is undesirable. Background reference: [Ensuring data reaches disk](https://lwn.net/Articles/457667/)

Unit tests were added to verify:

- no explicit sync happens when the threshold is `0`
- sync is triggered when accumulated bytes reach the threshold
- `McapWriter` correctly forwards the option to the file-backed writer path

Manual testing:

- compiled the C++ unit test binary with a direct local `g++` build
- ran the sync-focused tests and confirmed they passed
- ran the full unit test binary and confirmed all tests passed in the local verification build

MCAP file writes relied entirely on OS flush behavior. Data written by the library could still remain in kernel buffers for an unspecified amount of time before reaching persistent storage.

</td><td>

MCAP exposes `sizeForFsynch` so callers can request explicit periodic `fsync(fd)` calls after every configured number of bytes written, reducing the window of potential data loss during recording.

</td></tr></table>

### DIFF
--- a/cpp/mcap/include/mcap/writer.hpp
+++ b/cpp/mcap/include/mcap/writer.hpp
@@ -65,6 +65,15 @@ struct MCAP_PUBLIC McapWriterOptions {
    */
   uint64_t chunkSize = DefaultChunkSize;
   /**
+   * @brief Trigger data flush to disk once data written to the file from
+   * the last file synch meets or exceeds this size. The default value equal 0
+   * means no explicit calls for fsync(fd). i.e. OS will manage by its own how
+   * often to dump internal buffers to the disk.
+   * @note fsync(fd) is a blocking system call and could affect total
+   * disk write throughput.
+   */
+  uint64_t sizeForFsynch = 0;
+  /**
    * @brief Compression algorithm to use when writing Chunks. This option is
    * ignored if `noChunking=true`.
    */
@@ -167,7 +176,7 @@ class MCAP_PUBLIC FileWriter final : public IWritable {
 public:
   ~FileWriter() override;
 
-  Status open(std::string_view filename);
+  Status open(std::string_view filename, uint64_t size_for_fsync);
 
   void handleWrite(const std::byte* data, uint64_t size) override;
   void end() override;
@@ -177,6 +186,8 @@ public:
 private:
   std::FILE* file_ = nullptr;
   uint64_t size_ = 0;
+  uint64_t bytes_since_fsync_ = 0;
+  uint64_t size_for_fsync_ = 0;
 };
 
 /**
--- a/cpp/test/unit_tests.cpp
+++ b/cpp/test/unit_tests.cpp
@@ -1,3 +1,9 @@
+#if !defined(_WIN32) && !defined(__CYGWIN__)
+#  define fsync mcap_test_fsync
+#else
+#  define _commit mcap_test_commit
+#endif
+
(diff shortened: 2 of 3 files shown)


cpp/test/unit_tests.cpp
```diff
@@ -1,19 +1,86 @@
...
+    path = tempFile;
+#endif
+  }
+
+  ~TempFilePath() {
+    if (!path.empty()) {
+      std::remove(path.c_str());
+    }
+  }
+};
```
clalancette, line 80: I think we can mostly replace this whole thing with something like:

```
struct TempFilePath {
  std::string path;

  TempFilePath() {
    const auto unique = std::to_string(
      std::chrono::steady_clock::now().time_since_epoch().count());
    path = (std::filesystem::temp_directory_path() /
            ("mcap-writer-test-" + unique + ".mcap")).string();
  }

  ~TempFilePath() {
    std::error_code ec;
    std::filesystem::remove(path, ec);  // no-throw overload; fine in a destructor
  }
};
```

(it is not as free from races as mkstemp, but that doesn't really matter for a test)

cpp/mcap/include/mcap/writer.hpp
```diff
@@ -64,6 +64,15 @@ struct MCAP_PUBLIC McapWriterOptions {
...
   uint64_t chunkSize = DefaultChunkSize;
+  /**
+   * @brief Trigger data flush to disk once data written to the file from
+   * the last file synch meets or exceeds this size. The default value equal 0
+   * means no explicit calls for fsync(fd). i.e. OS will manage by its own how
+   * often to dump internal buffers to the disk.
+   * @note fsync(fd) is a blocking system call and could affect total
+   * disk write throughput.
+   */
+  uint64_t sizeForFsynch = 0;
```
clalancette, line 75:
```suggestion
  uint64_t sizeForFsync = 0;
```

clalancette (changes requested):
I think the idea is fine; let me know if you want to continue this PR.

Besides the changes I've left inline, this PR also needs to be rebased.  And if we are going to add this functionality to C++, I'd rather add it to the other languages as well.

PR #1611, opened 2026-03-26
Add cross-language MCAP performance benchmarks.

clalancette:
### Changelog

None.

### Docs

None.

### Description

I've recently been looking at/thinking about MCAP performance. However, it turns out that we don't currently have a good way to talk about performance since we can't measure it.

Fix that by adding a benchmarking/ directory with read and write benchmarks for C++, Rust, Go, Python, and TypeScript. Each language has its own *_bench/ subdirectory. TypeScript supports unchunked/chunked/zstd (LZ4 compression is unavailable in wasm-lz4).

Benchmark modes:
- Fixed-payload: 1M messages x 100 bytes, single channel
- Mixed-payload: 10-second simulated robot recording with 5 channels (/imu 96B@200Hz, /odom 296B@50Hz, /tf 80-1600B@100Hz, /lidar 230KB@10Hz, /camera 512KB@15Hz) — 3750 messages, ~102 MB
- Filtered reads: topic filter (/imu), time range (seconds 3-5), and combined topic+time (/lidar seconds 4-6) using mixed files

All modes run across 4 compression types (unchunked, chunked, zstd, lz4) with configurable payload fill: uniform (0x42, best-case compression) or varied (deterministic pattern for realistic ratios).

`make bench` runs the full matrix: both fill patterns, all compression modes, fixed + mixed payloads, and filtered reads. Includes peak memory tracking and summary tables via run_bench.sh.


benchmarking/rust_bench/src/bin/bench_read.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+        Some(filter_mode) => {
+            // The indexed reader operates on byte slices, so the filtered
+            // path buffers the whole file. Filtered results do not feed
+            // the memory table.
+            let buf = std::fs::read(filename).expect("Failed to read file");
+            let summary = mcap::Summary::read(&buf)
+                .expect("Failed to read summary")
+                .expect("No summary found in file");
```
bennetthardwick: afaik there is also a sans io summary reader that could be used here instead of reading the whole file into memory.

bennetthardwick (approved):
I haven't thoroughly reviewed the code but I've run it locally and it all works. Very interesting results.

I think it could be cool to:
- Make it possible to run a single language at a time
- Make it possible to attach perf / hotspot / whatever to debug perf
- Continuously run this against main and store the results somewhere

Not required in this change of course.

benchmarking/rust_bench/src/bin/bench_read.rs
clalancette: Oh, very good call.  Switching to that makes it much more similar to the other languages.  I've done that in [cdf948d](https://github.com/foxglove/mcap/pull/1611/commits/cdf948d4dfc2cfcf6fdd54500ef2f682564b4894) now, thanks.

clalancette:
> * Make it possible to run a single language at a time
> 
> * Make it possible to attach perf / hotspot / whatever to debug perf
> 
> * Continuously run this against main and store the results somewhere

All of these are good ideas.  I've implemented the first two in [63c71d7](https://github.com/foxglove/mcap/pull/1611/commits/63c71d7bc7ed84435d3f8b814017262d82a2afc7).  I didn't do the 3rd one, because I don't really know how we'd ensure that we use the "same" kind of runner every time for GitHub actions, but I like the idea in principle.

benchmarking/Makefile
```diff
@@ -0,0 +1,52 @@
+CXX ?= g++
+# -g and -fno-omit-frame-pointer keep perf/hotspot profiles readable at
+# negligible runtime cost.
+CXXFLAGS ?= -O2 -g -fno-omit-frame-pointer -Wall -Wextra -Wpedantic -Wconversion \
```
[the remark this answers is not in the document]
clalancette: I measured, and in C++, `-fno-omit-frame-pointer` had approximately no effect (it was within -1% to 1% of having it off).  I then tried to change Rust to make it similar, but there it had a 8% effect, so I left it as-is.  It is asymmetric, but I think that should be fine in this case.  The latest push updates the README to show how to use this for all languages.

clalancette:
No further comments here, so I'm going to go ahead and merge this one in.  Thanks @bennetthardwick for the review!

PR #1619, opened 2026-03-30
python: accept array-like types for array field serialization

### LINKED ISSUE #1469: ```mcap-ros2-support``` Fails to Serialize ```float64[]``` Array Fields (NumPy Conflict)
**Description**

When attempting to serialize a ROS 2 message (like `sensor_msgs/msg/Imu`) where a fixed-size array field (e.g., `orientation_covariance`: `float64[9]`) is initialized as a NumPy `ndarray`, the mcap-ros2-support library throws a `ValueError`. The serialization checker strictly expects a standard Python sequence (`list` or `tuple`) for array types, even though the array is numeric.

* Version: <br>`mcap-ros2-support` (PyPI)<br>`mcap` 1.3.0 (PyPI)
* Platform:<br>ROS2 Jazzy<br>Python 3.12

**Steps To Reproduce**

1. Set up a minimal ROS 2 package environment with the recorder running.
2. Publish a `sensor_msgs/msg/Imu` message using a tool (e.g., `ros2 topic pub`) that initializes the fixed-size array fields (like `orientation_covariance`) as a NumPy ndarray object.

Note: The issue persists even when initializing with a Python list structure, indicating the conversion to `ndarray` happens upstream of the writer.

**Expected Behavior**<br>The message should be successfully serialized and recorded, as the fixed-size array conversion should be handled by the installed `mcap-ros2-support` version.

\*\* Actual Behavior\*\*

```
ValueError: Field "orientation_covariance" is not an array (<class 'numpy.ndarray'>) but has array type "float64[]"
...
File "/home/user/.local/lib/python3.12/site-packages/mcap_ros2/_dynamic.py", line 431, in _write_complex_type
    raise ValueError( 
```

**Proposed Fix Location** <br>The fix likely needs to ensure that `numpy.ndarray` objects for numeric fixed-size arrays (`float64[]`) are explicitly handled within the dynamic encoding logic:

[https://github.com/foxglove/mcap/blob/8810067675c0f6eb00d6d9dd45a80231da447ada/python/mcap-ros2-support/mcap_ros2/\_dynamic.py#L427](<https://github.com/foxglove/mcap/blob/8810067675c0f6eb00d6d9dd45a80231da447ada/python/mcap-ros2-support/mcap_ros2/_dynamic.py#L427>)

remete618:
Fixes #1469

Accepts any object with `__len__` and `__getitem__` (e.g. numpy arrays, custom Sequence subclasses) when serializing array fields, instead of only `list`/`tuple`/`bytes`/`array.array`.

Uses duck-typing as suggested in #1469 -- no numpy runtime dependency added.

### What changed

A small `_is_array_like` helper gates the check. Four locations in `_write_complex_type` updated to try `list(array)` on array-like objects before raising `ValueError`. Existing types (`list`, `tuple`, `bytes`, `array.array`) still hit their fast paths via `isinstance` first.

`str`, `bytes`, and `dict` are explicitly excluded from the duck-type check.

### Tests

- Roundtrip with a custom `Sequence` subclass (dynamic and fixed-size arrays)
- Byte-for-byte parity: custom Sequence output matches list output
- Negative tests: `str`, `int`, `dict` correctly rejected
- numpy roundtrip tests (dynamic and fixed-size `float64[9]`) using `pytest.importorskip`

numpy tests use `importorskip` so they run where numpy is available and skip otherwise. Happy to add numpy to dev deps if you'd prefer them to run in CI.

### DIFF
--- a/python/mcap-ros2-support/mcap_ros2/_dynamic.py
+++ b/python/mcap-ros2-support/mcap_ros2/_dynamic.py
@@ -23,6 +23,15 @@ PrimitiveValue = Union[bool, int, float, str]
 DefaultValue = Union[PrimitiveValue, List[PrimitiveValue]]
 
 
+def _is_array_like(obj: Any) -> bool:
+    """Check if obj is array-like (has __len__ and __getitem__) but not str/bytes/dict."""
+    return (
+        hasattr(obj, "__len__")
+        and hasattr(obj, "__getitem__")
+        and not isinstance(obj, (str, bytes, dict))
+    )
+
+
 def _parseWstring(reader: CdrReader) -> str:
     raise NotImplementedError("wstring parsing is not implemented")
 
@@ -368,10 +377,13 @@ def _write_complex_type(
                 if array is None:
                     array = []
                 if not isinstance(array, list):
-                    raise ValueError(
-                        f'Field "{field.name}" is not an array but has array type '
-                        f'"{ftype.type}[]"'
-                    )
+                    if _is_array_like(array):
+                        array = list(array)
+                    else:
+                        raise ValueError(
+                            f'Field "{field.name}" is not an array but has array type '
+                            f'"{ftype.type}[]"'
+                        )
 
                 if ftype.is_fixed_size_array() and ftype.array_size is not None:
                     # Fixed length array, ensure the input array is the correct length
@@ -424,16 +436,15 @@ def _write_complex_type(
                 )
                 if array is None:
                     array = []
-                if (
-                    not isinstance(array, list)
-                    and not isinstance(array, tuple)
-                    and not isinstance(array, bytes)
-                    and not isinstance(array, py_array.array)
-                ):
-                    raise ValueError(
-                        f'Field "{field.name}" is not an array ({type(array)}) but has array type '
-                        f'"{ftype.type}[]"'
-                    )
+                if not isinstance(array, (list, tuple, bytes, py_array.array)):
+                    if _is_array_like(array):
+                        array = list(array)
+                    else:
+                        raise ValueError(
+                            f'Field "{field.name}" is not an array '
+                            f"({type(array)}) but has array type "
+                            f'"{ftype.type}[]"'
+                        )
 
                 # Special handling for bytes
                 if isinstance(array, bytes) or (
@@ -476,11 +487,9 @@ def _write_complex_type(
                         )
 
                     if ftype.is_fixed_size_array() and ftype.array_size is not None:
-                        # Convert tuples to lists
+                        # Convert non-list sequences to lists
                         list_array = (
-                            list(array)
-                            if isinstance(array, (tuple, py_array.array))
-                            else array
+                            list(array) if not isinstance(array, list) else array
                         )
                         # Fixed length array, ensure the input array is the correct length
                         while len(list_array) < ftype.array_size:
@@ -501,9 +510,7 @@ def _write_complex_type(
                         ):
                             array = array[: ftype.array_size]
 
-                        array = (
-                            list(array) if isinstance(array, py_array.array) else array
-                        )
+                        array = list(array) if not isinstance(array, list) else array
                         array = _coerce_values(array, ftype.type, field.default_value)
 
                         # Dynamic length array, write a uint32 prefix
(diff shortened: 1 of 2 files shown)


remete618:
Bumping this one. It implements the duck-typing approach discussed in #1469, CI is green and it's still mergeable against main. Happy to rebase or adjust if anything's changed since March.

python/mcap-ros2-support/mcap_ros2/_dynamic.py
```diff
@@ -23,6 +23,15 @@
 DefaultValue = Union[PrimitiveValue, List[PrimitiveValue]]
 
 
+def _is_array_like(obj: Any) -> bool:
+    """Check if obj is array-like (has __len__ and __getitem__) but not str/bytes/dict."""
+    return (
+        hasattr(obj, "__len__")
+        and hasattr(obj, "__getitem__")
+        and not isinstance(obj, (str, bytes, dict))
```
clalancette, line 31: Excluding only `dict` here means that other dict-like things (`OrderedDict`, `defaultdict`, etc) could slip through and give us bogus data.  If we instead use `collections.abc.Mapping` (the superclass of all dict-like types), that should avoid most of this:


```suggestion
        and not isinstance(obj, (str, bytes, Mapping))
```

python/mcap-ros2-support/mcap_ros2/_dynamic.py
```diff
@@ -424,16 +436,15 @@ def _write_complex_type(
...
-                    not isinstance(array, list)
-                    and not isinstance(array, tuple)
-                    and not isinstance(array, bytes)
-                    and not isinstance(array, py_array.array)
-                ):
-                    raise ValueError(
-                        f'Field "{field.name}" is not an array ({type(array)}) but has array type '
-                        f'"{ftype.type}[]"'
-                    )
+                if not isinstance(array, (list, tuple, bytes, py_array.array)):
```
clalancette, line 439: I think this check prevents the "Special handling for bytes" below from triggering, meaning that we aren't doing the fastest thing possible.  What I'd propose here instead is a function like this:

```
def _as_uint8_bytes(array: Any) -> Optional[bytes]:
    if isinstance(array, bytes):
        return array
    try:
        view = memoryview(array)
    except TypeError:
        return None
    if view.ndim != 1 or view.format != "B":
        return None
    return view.tobytes()
```

Then calling it here.  If it returns non-None, just use it, otherwise convert to a list.

python/mcap-ros2-support/mcap_ros2/_dynamic.py
```diff
@@ -476,11 +487,9 @@ def _write_complex_type(
                         )
 
                     if ftype.is_fixed_size_array() and ftype.array_size is not None:
-                        # Convert tuples to lists
+                        # Convert non-list sequences to lists
                         list_array = (
-                            list(array)
-                            if isinstance(array, (tuple, py_array.array))
-                            else array
+                            list(array) if not isinstance(array, list) else array
```
clalancette, line 492: This is a pre-existing bug, but since we are here we may as well fix it.  The semantics here are weird; if the array is a list, then we mutate the *existing* data, and in all other cases, we mutate the newly-created list.  The latter is fine, the former is a problem.  We should probably rewrite this whole block to be something like:

```
                        # Fixed length array, pad or truncate to the correct length
                        # without mutating the caller's array
                        if len(array) < ftype.array_size:
                            array = array + [None] * (ftype.array_size - len(array))
                        elif len(array) > ftype.array_size:
                            array = array[: ftype.array_size]

                        array = _coerce_values(array, ftype.type, field.default_value)
                        array_writer_fn(writer, array)
```

python/mcap-ros2-support/mcap_ros2/_dynamic.py
```diff
@@ -501,9 +510,7 @@ def _write_complex_type(
                         ):
                             array = array[: ftype.array_size]
 
-                        array = (
-                            list(array) if isinstance(array, py_array.array) else array
-                        )
+                        array = list(array) if not isinstance(array, list) else array
```
clalancette, line 513: I think this line can just be deleted; there is no reason to make this copy since the next line will copy again anyway.

python/mcap-ros2-support/tests/test_ros2_writer.py
clalancette, line 1: I know that #1469 said we didn't want to take a numpy dependency, but that is for runtime.  I'm fine with adding one for devel/test time.  So let's import it here and write some tests using it to make sure it passes.

clalancette (changes requested):
Thank you for the PR, and sorry for the very long delay in looking at it.  I've now reviewed it and left some things to think about and change.  Please take a look when you have time.

PR #1640, opened 2026-04-23
[typescript] Add message index caching option

snosenzo:
### Changelog

- [typescript] `McapReader`: specify a cache size that would specifically be used for message indexes. 0 by default

### Docs

None - documented in JSDoc

### Description

This change adds the option for the typescript `McapIndexedReader` to cache message indexes so that over subsequent reads there are fewer read calls, resulting in faster subsequent message reads. This is especially effective when there is a latency for each read due to I/O restrictions. However, when reading from an in memory buffer there would be little-to-no gains.

Results from local benchmark (not averaged):
For the biggest case (200k messages × 1024 B @ 5 ms latency):

| Number of full mcap reads | % Faster with caching enabled |
| -------- | --------- |
| 1        | −2.2%     |
| 2        | 23.5%     |
| 3        | 28.1%     |
| 4        | 32.4%     |
| 5        | 33.7%     |

When the file is read once this is the expected worst case for the cache, since the cache is populated but never reused. Even there the extra work of allocating the cache map is negligible.

ERT-2082

### DIFF
--- a/typescript/core/src/CachedReadable.ts
+++ b/typescript/core/src/CachedReadable.ts
@@ -12,10 +12,27 @@ import type { IReadable } from "./types.ts";
  * starts at a different offset will miss.
  */
 export class CachedReadable implements IReadable {
+  /**
+   * The underlying source of the data to be cached.
+   */
   #readable: IReadable;
+  /**
+   * Cached data. Indexed by offset request and stored as a Uint8Array.
+   * If the requested size is less than the cached data, the cached data is returned as a subarray.
+   * If the requested size is greater than the cached data, the a new request is made to the underlying readable.
+   */
   #cache = new Map<bigint, Uint8Array>();
+  /**
+   * The maximum size of the cache in bytes.
+   */
   #maxCacheSizeBytes: number;
+  /**
+   * The current size of the cache in bytes.
+   */
   #currentCacheSizeBytes = 0;
+  /**
+   * The size of the underlying readable.
+   */
   #size: bigint | undefined;
 
   constructor(readable: IReadable, maxCacheSizeBytes: number) {
@@ -24,7 +41,10 @@ export class CachedReadable implements IReadable {
   }
 
   async size(): Promise<bigint> {
-    return (this.#size ??= await this.#readable.size());
+    if (this.#size == undefined) {
+      this.#size = await this.#readable.size();
+    }
+    return this.#size;
   }
 
   async read(offset: bigint, size: bigint): Promise<Uint8Array> {
--- a/typescript/core/src/CachedReadable.ts
+++ b/typescript/core/src/CachedReadable.ts
@@ -1,0 +1,50 @@
+import type { IReadable } from "./types.ts";
+
+/**
+ * Wraps an {@link IReadable} and caches the bytes returned from `read()` keyed by offset. Reads at
+ * an offset that was previously cached return a subarray of the cached buffer instead of calling
+ * the underlying readable.
+ *
+ * The cache is capped by `maxCacheSizeBytes`; once the cache is full, new reads pass through
+ * without being cached. No eviction is performed.
+ *
+ * Note: reads are cached by *exact* offset. A read that partially overlaps a cached range but
+ * starts at a different offset will miss.
+ */
+export class CachedReadable implements IReadable {
+  #readable: IReadable;
+  #cache = new Map<bigint, Uint8Array>();
+  #maxCacheSizeBytes: number;
+  #currentCacheSizeBytes = 0;
+  #size: bigint | undefined;
+
+  constructor(readable: IReadable, maxCacheSizeBytes: number) {
+    this.#readable = readable;
+    this.#maxCacheSizeBytes = maxCacheSizeBytes;
+  }
+
+  async size(): Promise<bigint> {
+    return (this.#size ??= await this.#readable.size());
+  }
+
+  async read(offset: bigint, size: bigint): Promise<Uint8Array> {
+    const requestedSize = Number(size);
+    const cached = this.#cache.get(offset);
+    if (cached != undefined && cached.byteLength >= requestedSize) {
+      return cached.byteLength === requestedSize ? cached : cached.subarray(0, requestedSize);
+    }
+
+    const data = await this.#readable.read(offset, size);
+
+    // The underlying readable is allowed to reuse its backing buffer across reads, so we must copy
+    // the bytes before storing them in the cache.
+    if (this.#currentCacheSizeBytes + data.byteLength <= this.#maxCacheSizeBytes) {
+      const copy = new Uint8Array(data);
+      this.#cache.set(offset, copy);
+      this.#currentCacheSizeBytes += copy.byteLength;
+      return copy;
+    }
+
+    return data;
+  }
+}
--- a/typescript/core/src/ChunkCursor.ts
+++ b/typescript/core/src/ChunkCursor.ts
@@ -4,20 +4,18 @@ import { sortedIndexBy } from "./sortedIndexBy.ts";
 import { sortedLastIndexBy } from "./sortedLastIndex.ts";
 import type { IReadable, TypedMcapRecords } from "./types.ts";
 
-/**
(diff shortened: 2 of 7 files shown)


typescript/core/src/ChunkCursor.ts
```diff
@@ -111,58 +121,83 @@ export class ChunkCursor {
...
+      }
 
-    if (reader.bytesRemaining() !== 0) {
-      throw new Error(`${reader.bytesRemaining()} bytes remaining in message index section`);
+      if (reader.bytesRemaining() !== 0) {
+        throw new Error(`${reader.bytesRemaining()} bytes remaining in message index section`);
+      }
+
+      if (channelOffsetsToCache) {
+        this.#messageIndexCache?.set(this.chunkIndex.chunkStartOffset, channelOffsetsToCache);
```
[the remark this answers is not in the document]
snosenzo: The linter isn't smart enough to realize this. So I added a comment to hopefully quell the bot

achim-k:
I would like to propose a slighly less intrusive and more low-level solution which uses a cached readable when loading message indexes.

When loading messages indexes, we pass the readable to the chunk cursor which then uses it to read the chunk's message index record.
https://github.com/foxglove/mcap/blob/73c2e41b60ebf0e071a12bef7e13bf91466c0357/typescript/core/src/McapIndexedReader.ts#L391-L392

For this call (and only for this one), we could pass in a "cached readable" that caches reads by offset position, so subsequent reads to the same offset yield much faster. Example:

```ts
await cursor.loadMessageIndexes(this.#cachedMessageIndexReadable ?? this.#readable);
```

with `#cachedMessageIndexReadable` being initialized in the constructor when the `messageIndexCacheSizeBytes` argument is greater than 0:
```
const messageIndexCacheSizeBytes = args.messageIndexCacheSizeBytes ?? 0;
if (messageIndexCacheSizeBytes) {
  this.#cachedMessageIndexReadable = new CachedReadable(this.#readable, messageIndexCacheSizeBytes);
}
```

A simple implementation of the `CachedReadable` class could look like this (no eviction when cache is full):

```ts
export class CachedReadable implements IReadable {
  #readable: IReadable;
  #cache: Map<bigint, Uint8Array>;
  #maxCacheSizeBytes: number;
  #currentCacheSizeBytes: number;
  #size: bigint | undefined;

  constructor(readable: IReadable, maxCacheSizeBytes: number) {
    this.#readable = readable;
    this.#cache = new Map();
    this.#maxCacheSizeBytes = maxCacheSizeBytes;
    this.#currentCacheSizeBytes = 0;
    this.#size = undefined;
  }
  
  async size(): Promise<bigint> {
    return this.#size ??= await this.#readable.size();
  }

  async read(offset: bigint, size: bigint): Promise<Uint8Array> {
    const requestedSize = Number(size);
    const cachedData = this.#cache.get(offset);
    if (cachedData && cachedData.byteLength >= requestedSize) {
      if (cachedData.byteLength > requestedSize) {
        return cachedData.subarray(0, requestedSize);
      }
      return cachedData;
    }

    const data = await this.#readable.read(offset, size);
    if (this.#currentCacheSizeBytes + data.byteLength > this.#maxCacheSizeBytes) {
      // Cache is full, return the data without caching it
      return data;
    }

    // Cache the data
    this.#cache.set(offset, data);
    this.#currentCacheSizeBytes += data.byteLength;
    return data;
  }
}
```
  
Disclaimer: I haven't actually tested if that approach would work

typescript/core/src/ChunkCursor.ts
```diff
@@ -10,6 +10,12 @@ type ChunkCursorParams = {
   startTime: bigint | undefined;
   endTime: bigint | undefined;
   reverse: boolean;
+  /**
+   * When true, `loadMessageIndexes()` reads the full message index region for the chunk (not just
+   * the portion containing `relevantChannels`). This is useful when the underlying readable caches
+   * reads, since it keeps the read offset/size stable across calls with different topic filters.
+   */
+  readFullMessageIndexRange?: boolean;
```
achim-k, line 18: do we actually need this? I guess it's also perfectly fine to only cache portions of relevant channels, maybe that's even preferred?

snosenzo: Ah yeah good callout, yeah it's not really needed

typescript/core/src/CachedReadable.ts
```diff
@@ -0,0 +1,50 @@
...
+    const requestedSize = Number(size);
+    const cached = this.#cache.get(offset);
+    if (cached != undefined && cached.byteLength >= requestedSize) {
+      return cached.byteLength === requestedSize ? cached : cached.subarray(0, requestedSize);
+    }
+
+    const data = await this.#readable.read(offset, size);
+
+    // The underlying readable is allowed to reuse its backing buffer across reads, so we must copy
+    // the bytes before storing them in the cache.
```
achim-k, line 60: Nice catch!

typescript/core/src/CachedReadable.ts
```diff
@@ -0,0 +1,50 @@
...
+  #currentCacheSizeBytes = 0;
+  #size: bigint | undefined;
+
+  constructor(readable: IReadable, maxCacheSizeBytes: number) {
+    this.#readable = readable;
+    this.#maxCacheSizeBytes = maxCacheSizeBytes;
+  }
+
+  async size(): Promise<bigint> {
+    return (this.#size ??= await this.#readable.size());
```
achim-k: Nit: Might want to rewrite this to not use the [??= operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment), which is available from Chrome 85 onwards. Unless we are already using language features that have a higher requirement

typescript/core/src/CachedReadable.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Note: reads are cached by *exact* offset. A read that partially overlaps a cached range but
+ * starts at a different offset will miss.
+ */
+export class CachedReadable implements IReadable {
+  #readable: IReadable;
+  #cache = new Map<bigint, Uint8Array>();
+  #maxCacheSizeBytes: number;
+  #currentCacheSizeBytes = 0;
+  #size: bigint | undefined;
```
achim-k, line 36: would be nice if these members had brief js-doc style comments

achim-k (approved):

typescript/core/src/CachedReadable.ts
snosenzo: Yeah it looks like it does bump our requirement from 84 to 85 so I'll remove it. Good catch.

typescript/core/src/ChunkCursor.ts
snosenzo: Sorry I wasn't quite correct about this. Since I misread how the relevant channels are handled. The current behavior is that it reads the earliest channel's message index offset to the end of the message index. This means that given how our cache works this would cause cache misses because we cache based on offset and we wouldn't know that we already have the indexes in the cache if we fetched a different offset. I think this is better since it reduces the number of requests more generally across different topic reads.

achim-k, line 18: I believe it would be better to undo the change here again. I'd say that it's preferable to only cache message indexes of channels that were actually requested, not all of them

achim-k: oh wait, only now saw your comment on the previous thread

typescript/core/src/CachedReadable.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    const data = await this.#readable.read(offset, size);
+
+    // The underlying readable is allowed to reuse its backing buffer across reads, so we must copy
+    // the bytes before storing them in the cache.
+    if (this.#currentCacheSizeBytes + data.byteLength <= this.#maxCacheSizeBytes) {
+      const copy = new Uint8Array(data);
+      this.#cache.set(offset, copy);
+      this.#currentCacheSizeBytes += copy.byteLength;
```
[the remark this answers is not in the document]
snosenzo: I'll address this in the followup for prefetching. Not an issue as the comment points out in the current architecture.

PR #1641, opened 2026-04-23
[typescript] `McapReader`: add option to prefetch message indexes in parallel

snosenzo:
### Changelog

 -[typescript] Add optional `IReadable`field `supportsConcurrentReads` to advertise capability to `McapReader`s
 - [typescript] Add optional `prefetchMessageIndexes` argument to `McapIndexedReader` which will enable caching and prefetch message indexes if the underlying Readable supports concurrent reads. 

### Docs

None. Documented in JSDoc docs

### Description

This allows message indexes to be loaded in parallel starting when the Reader is initialized which significantly speeds up message index reading. Currently the concurrency is locked to up to 6 concurrent requests, based around browser HTTP:1 limits. This could be specified in the future, but I didn't want to add more api surface area. 

If prefetch is enabled and no cache size is specified it will automatically size the cache to the size of the message indexes in the mcap, with a safe maximum cache size of 256MB.

If one of the prefetches fails then it will bail on prefetching and return the error in the first `readMessages` call so that the error can be surfaced to the owner of the Reader. Subsequent reads will act normally.

| numReads | latencyMs | noCache (ms) | cached (ms) | prefetch (ms) | cached vs noCache | prefetch vs noCache | prefetch vs cached |
| -------- | --------- | ------------ | ----------- | ------------- | ----------------- | ------------------- | ------------------ |
| 1        | 0         | 429.7        | 455.1       | 438.4         | -5.9%             | -2.0%               | 3.7%               |
| 1        | 10        | 4,825.1      | 4,707.0     | 2,988.0       | 2.4%              | **38.1%**           | **36.5%**          |
| 5        | 0         | 2,325.7      | 2,198.6     | 2,529.8       | 5.5%              | -8.8%               | -15.1%             |
| 5        | 10        | 23,871.2     | 15,359.5    | 13,487.8      | **35.7%**         | **43.5%**           | 12.2%              |

The losses are expected for the no/lowest-latency, single read cases, since this benchmark uses an in memory mcap buffer and the overhead of the requests wins out. Enabling this option is not advised in those cases. This option is meant to be used primarily for remote file data sources which have a latencies of >20ms. In that range the savings are very apparent.

### DIFF
--- a/typescript/core/src/CachedReadable.test.ts
+++ b/typescript/core/src/CachedReadable.test.ts
@@ -84,6 +84,57 @@ describe("CachedReadable", () => {
     expect(readable.reads).toHaveLength(3);
   });
 
+  it("concurrent reads for the same offset issue only one underlying read", async () => {
+    const data = new Uint8Array([0, 1, 2, 3, 4, 5, 6, 7]);
+    const readable = makeReadable(data);
+    const cached = new CachedReadable(readable, 1024);
+
+    const [a, b] = await Promise.all([cached.read(0n, 4n), cached.read(0n, 4n)]);
+    expect(Array.from(a)).toEqual([0, 1, 2, 3]);
+    expect(Array.from(b)).toEqual([0, 1, 2, 3]);
+    expect(readable.reads).toHaveLength(1);
+  });
+
+  it("concurrent reads for the same offset do not double-count cache size", async () => {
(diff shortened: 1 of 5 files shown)


jtbandes:
@snosenzo do we still want this?

PR #1643, opened 2026-05-15
cpp: BufferWriter: Reserve chunkSize in constructor

sebastianlipponer:
### Changelog

cpp: BufferWriter: Reserve chunkSize in constructor

### Docs

None

### Description

The Buffer Writer at a 10M payload scale benefits the most, achieving a 2.22% reduction in median execution time. This is followed by the File Writer at 10M and 1M payload scales, which show performance gains of 0.88% and 0.73% respectively. Additionally, the Stream Writer at the 10M scale shows a measurable improvement of 0.41%.

### DIFF
--- a/cpp/mcap/include/mcap/writer.hpp
+++ b/cpp/mcap/include/mcap/writer.hpp
@@ -252,6 +252,8 @@ protected:
  */
 class MCAP_PUBLIC BufferWriter final : public IChunkWriter {
 public:
+  explicit BufferWriter(uint64_t chunkSize = 0);
+
   void handleWrite(const std::byte* data, uint64_t size) override;
   void end() override;
   uint64_t size() const override;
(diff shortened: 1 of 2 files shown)


cpp/mcap/include/mcap/writer.hpp
```diff
@@ -252,6 +252,8 @@ class MCAP_PUBLIC IChunkWriter : public IWritable {
  */
 class MCAP_PUBLIC BufferWriter final : public IChunkWriter {
 public:
+  BufferWriter(uint64_t chunkSize = 0);
```
clalancette: Can you please add `explicit` here?


```suggestion
  explicit BufferWriter(uint64_t chunkSize = 0);
```

clalancette (changes requested):
One minor thing to change, otherwise this looks good.

Using my own benchmarks in #1611 for comparison, on chunked, uncompressed writes, this sped things up by about 2%.

cpp/mcap/include/mcap/writer.hpp
sebastianlipponer: done

clalancette (approved):
Looks good to me, thanks for iterating.

PR #1651, opened 2026-05-25
Add a parallel (multithreaded) MCAP reader to the C++ library

facontidavide:
## Summary

Adds **`mcap::ParallelReader`** to the C++ library: it decompresses chunks on a thread pool ahead of a k-way merge and emits messages in **exactly the same order as the serial reader**, so time-ordered reads of zstd/lz4 files become bound by cores/IO instead of a single decompression thread.

```cpp
mcap::ParallelReader pr;
if (!pr.open("file.mcap").ok()) { /* handle */ }
mcap::ParallelReadOptions opts;
opts.read.readOrder = mcap::ReadMessageOptions::ReadOrder::LogTimeOrder;
opts.threadCount = 4;
opts.maxBytesInFlight = 1ull << 30;   // soft memory cap
opts.lookaheadBytes   = 1ull << 30;
for (const auto& mv : pr.readMessages(onProblem, opts)) {
  decode(mv.message, mv.channel, mv.schema);
}
```

## Motivation / results

On a real 17 GB zstd recording (562,580 messages), output is **byte-identical** to `McapReader::readMessages` (order + payload hash) while scaling:

| threads | wall time | speedup |
|--:|--:|--:|
| serial | 23.4 s | 1.0× |
| 2 | 11.6 s | 2.0× |
| 4 | 6.0 s | 3.9× |
| 8 | 3.3 s | 7.1× |

Peak **resident decompressed** memory stays at a few MiB (bounded by the byte budget), not the full uncompressed size.

## Design

- **`ParallelReader`** composes a `McapReader` (used only through its public API — `readSummary`/`channels`/`schemas`/`chunkIndexes`/static `Parse*`) and owns the concurrent-safe source. `ParallelMessageView` is the ordered, range-for view. It reuses `RecordOffset`'s comparison operators so tie-breaks match the serial reader exactly.
- **`MmapReader`** — a concurrent-safe `IReadable` (POSIX `mmap` + Windows `CreateFileMapping`/`MapViewOfFile`). `ParallelReader::open(IReadable&)` also accepts a caller-provided concurrent source.
- Memory is capped by a byte-budget semaphore; decompressed chunks are RAII-released as the merge drains them, so prefetch never exceeds the budget (no deadlock: required chunks force-acquire).
- zstd decompresses into an uninitialized buffer via a thread-local `ZSTD_DCtx` (avoids per-chunk context alloc + a redundant zero-fill).

## Footprint on existing code

Intentionally minimal. The **only** change to an existing type is a defaulted virtual:

```cpp
// IReadable
virtual bool supportsConcurrentRead() const { return false; }
```

`MmapReader` overrides it `true`; `FileReader`/`FileStreamReader` inherit `false` (they have a single shared cursor and can't back concurrent reads). Everything else is new headers, all gated by **`MCAP_NO_PARALLEL`** (define it to compile the feature out entirely, e.g. for toolchains without threads).

New headers under `cpp/mcap/include/mcap/`: `parallel_reader.hpp`, `parallel_budget.hpp`, `byte_semaphore.hpp`, `thread_pool.hpp`, `mmap_reader.hpp`.

## Tests

`cpp/test/parallel_reader_test.cpp` (Catch2, ~1.66M assertions): parity vs the serial reader across compression {none, zstd, lz4} × overlap/disjoint × duplicate-timestamp ties × orders {forward, reverse} × threads {1,2,4,8}, plus time-range/topic filters, determinism under repeated runs, non-concurrent-source rejection, early-teardown, and the residency-profile/semaphore/budget unit tests. Wired into `test/CMakeLists.txt` with `-nocompress` and a `unit-tests-noparallel` variant (proves the serial API builds with the feature compiled out), and added to `make test-host`.

## Notes for reviewers

- The new headers are currently **header-only inline** (consistent with `intervaltree.hpp` / `read_job_queue.hpp` / `crc32.hpp`). If you'd prefer the `.hpp`/`.inl` split used by `reader`/`writer`/`types`, I'm happy to split `parallel_reader.hpp`.
- The `IReadable` virtual is source- and behaviour-compatible; it adds one vtable slot (no ABI guarantee is made by the library today, and callers compile from source).
- Windows mmap uses `CreateFileMapping`/`MapViewOfFile`; CI (`cpp-windows`) builds it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -51,6 +51,9 @@ test: dev-image
 test-host:
 	./test/build/Debug/bin/unit-tests
 	./test/build/Debug/bin/unit-tests-nocompress
+	./test/build/Debug/bin/unit-tests-noparallel
+	./test/build/Debug/bin/parallel-reader-test
+	./test/build/Debug/bin/parallel-reader-test-nocompress
 
 .PHONY: hdoc-build
 hdoc-build:
--- a/cpp/mcap/include/mcap/byte_semaphore.hpp
+++ b/cpp/mcap/include/mcap/byte_semaphore.hpp
@@ -1,0 +1,109 @@
+#pragma once
+//
+// byte_semaphore.hpp
+//
+// A counting semaphore measured in BYTES, used to cap the total resident
+// (decompressed) memory of the parallel reader. A worker acquires
+// `chunk.uncompressedSize` credits before decompressing a chunk and releases
+// them once the chunk is fully drained and unpinned. The total outstanding
+// credits can never exceed the configured capacity, EXCEPT for the deliberate
+// oversized-chunk case described below.
+//
+// The cap is SOFT. There are two ways to take credits:
+//
+//   * acquire()      - blocking; respects the cap. Used for SPECULATIVE look-ahead
+//                      chunks. This is the back-pressure path: when the consumer is
+//                      slow or memory is full, prefetch workers park here.
+//   * forceAcquire() - non-blocking; ALWAYS succeeds, even if it drives the pool
+//                      over capacity (the counter goes negative). Used for chunks
+//                      the merge MUST decompress to emit the next in-order message,
+//                      where blocking would deadlock. Exceeding the cap to avoid the
+//                      lock is the intended behavior.
+//
+// The overshoot is bounded: the set of chunks force-acquired at once is the
+// temporal overlap depth at the current emit frontier, and it self-corrects,
+// because while available_ is negative every blocking acquire() parks, so no new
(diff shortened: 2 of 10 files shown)


amacneil:
@facontidavide do you still want someone to review this? or no longer needed?

facontidavide:
I will reopen it soon. I am doing first some additional tests to be sure that everything works 100%.

Spoiler alert: it is amazing. My loading time is dramatically lower.

PR #1658, opened 2026-05-31
Chunk-size benchmark and study

amacneil:
### Background

We had no data behind MCAP's default chunk size, and the library defaults are inconsistent across languages (C++/Rust 768 KiB; Go/Python/TS 1 MiB; Swift 10 MiB — the 768 KiB value is inherited from ROS 1 `rosbag`). This PR adds a focused, C++-only benchmark (C++ is the `rosbag2` MCAP writer) that sweeps chunk size and quantifies the real tradeoffs, plus a written report with figures, per-language defaults + history, storage-overhead analysis, and a cross-language read/write correlation check.

The harness varies **chunk size** (256 KiB → 32 MiB), **message-size class** (small telemetry / incompressible JPEG / point cloud / mixed), **compression** (zstd + lz4), and **read access pattern** (full / point / range / streaming / topic), and models read cost for local vs. regional vs. high-latency remote storage under both a naive and a coalescing reader.

### Key findings

- **Compression ratio is flat from 256 KiB to 32 MiB** for every content type; write throughput is flat too. Bigger chunks buy nothing on size.
- **Bigger chunks have monotonic costs:** reader peak RSS grows ~10× (≈7 MiB → ≈65 MiB), and point/random reads get linearly more expensive (single-message read fetches ~14 MiB at 32 MiB chunks vs ~0.4 MiB at 1 MiB).
- **Bigger chunks do not robustly help remote reads.** Full scans are chunk-size-independent; random access is strictly worse with big chunks; and the streaming "benefit" only exists for a naive reader that fetches one chunk per GET — a coalescing reader is flat-to-faster with *small* chunks.
- **Storage overhead is driven by message count, not chunk size** (~16 B/message message index, uncompressed): ~22% for tiny high-rate topics, <0.03% for point clouds.
- **Recommended default: 1 MiB** — every chunk-size-dependent metric is flat or favors smaller chunks. Matches the existing Go/Python/TS defaults; the only outlier worth fixing is Swift's 10 MiB.

### Report

See [REPORT.md](https://github.com/foxglove/mcap/blob/7f7ea94ee3a2f1f6fa4cb5a264a7915d5c21b6d7/benchmarking/chunk_size_study/REPORT.md) for the full write-up.


amacneil:
Migrated all languages to 1 MiB chunks in https://github.com/foxglove/mcap/pull/1659

PR #1703, opened 2026-06-09
[cpp] Multi-threaded ParallelReader

facontidavide:
## Summary

Adds a standalone, multithreaded `mcap::ParallelReader` for log-time-ordered reads.
It decompresses chunks on a thread pool ahead of a k-way merge frontier and emits
messages in exactly the same order as the serial `IndexedMessageReader` (verified
by parity tests). **Purely additive**: adds `mcap::ParallelReader` (+ a concurrent
file reader) without modifying `McapReader`; the only touch to an existing type is
`IReadable::supportsConcurrentRead()` (defaults `false`). Define `MCAP_NO_PARALLEL`
to omit it.

## What's included
- **`parallel_reader.hpp`** — concurrent chunk decompression + ordered
  single-consumer k-way merge.
- **`concurrent_file_reader.hpp`** — the `open(path)` source: positioned reads
  (`pread` on POSIX, `ReadFile`+`OVERLAPPED` offset on Windows). The offset is
  passed per call, so many workers read concurrently with no shared cursor, through
  the page cache.
- **`thread_pool.hpp`, `byte_semaphore.hpp`, `parallel_budget.hpp`** — worker pool
  + memory back-pressure.
- **`message_byte_store.hpp`** — deferred per-message byte access.
- **`cpp/test/parallel_reader_test.cpp`** — parity + behavior tests.

## I/O: positioned reads through the page cache
The source uses positioned reads (`pread` / `ReadFile` with an `OVERLAPPED` offset).
The offset is passed per call, so workers read concurrently with no shared cursor,
and reads stay resident in the page cache: a full read does **~0 major page faults**,
keeps RSS low with no extra RSS-bounding machinery, and is **faster than the serial
reader even single-threaded** on a ~17 GB file.

## Memory back-pressure + thread default
`ParallelReadOptions::memoryCap`: `ByteBudget` (default, precise resident-byte
ceiling) or `ChunkCount` (opt-in, cap live chunks). Default worker threads is **4
(capped at 8)** instead of `hardware_concurrency()` — benchmarks showed the latter
was consistently slower (reads are usually consumer-bound and decompression
saturates memory bandwidth before all cores).

> Reviewer note: the cap currently also clamps an explicit `threadCount`; happy to
> honor explicit values and cap only the automatic (0) default if preferred.

## Per-message read-path optimizations (profiled with `perf`)
1. Resolve channel/schema by **const-ref** from one-time snapshots instead of
   `McapReader::channel()`/`schema()` returning `shared_ptr` by value (removes
   per-message atomic refcount churn).
2. Order each chunk's entries by **k-way merging the already-sorted per-channel
   `MessageIndex` runs** instead of a full `std::sort` (fallback to `std::sort` for
   reverse / non-monotonic / single-run). Emit order unchanged (parity-verified).


cpp/mcap/include/mcap/message_byte_store.hpp
clalancette, line 1: Is this file needed in this PR?  I don't see it being used anywhere in the implementation or the tests.  It also does things like log directly to stderr, and references things like "PJ4", the "ObjectStore", etc. which aren't in the mcap library.  Can you explain more about this file?

cpp/mcap/include/mcap/parallel_budget.hpp
```diff
@@ -0,0 +1,202 @@
...
+  }
+  return p;
+}
+
+// Policy for what to do when the user's cap is below the order's intrinsic floor.
+enum class MemoryCapPolicy {
+  Adapt,                 // raise the effective budget up to the floor (never deadlocks;
+                         // may exceed the user cap — flagged in BudgetDecision)
+  Strict,                // do not exceed the user cap; report infeasible (caller decides)
+  EvictAndReDecompress,  // honor a sub-floor cap by evicting + re-decompressing chunks
```
clalancette: I might be wrong, but it doesn't seem like the `EvictAndReDecompress` mode is implemented.  `resolveBudget` computes for it, but nothing ever looks at `requiresEviction` or `estReDecompressionFactor`.  Can you explain more about it?

cpp/mcap/include/mcap/parallel_reader.hpp
```diff
@@ -0,0 +1,960 @@
...
+#ifndef MCAP_COMPRESSION_NO_ZSTD
+#  include <zstd.h>
+#endif
+
+namespace mcap {
+
+// Back-pressure policy for the parallel reader's resident decompressed memory.
+enum class MemoryCapMode {
+  ByteBudget,  // default: cap resident decompressed BYTES (precise memory ceiling)
+  ChunkCount,  // opt-in: cap the NUMBER of concurrently-live chunks (coarser bound)
```
clalancette: It's not really clear to me whether we should bother implementing the `ChunkCount` right now.  It's not a huge additional amount of code, but the benefits are unclear.  Can you make an argument for whether we should keep it?

cpp/mcap/include/mcap/byte_semaphore.hpp
```diff
@@ -0,0 +1,109 @@
...
+  // driving available_ negative so all other acquirers wait.
+  void acquire(uint64_t n) {
+    std::unique_lock<std::mutex> lk(m_);
+    const int64_t need = int64_t(n);
+    const int64_t threshold = std::min<int64_t>(need, int64_t(capacity_));
+    cv_.wait(lk, [&] {
+      return available_ >= threshold;
+    });
+    available_ -= need;
+  }
```
clalancette: Other than in tests, is `acquire` actually used anywhere?

cpp/mcap/include/mcap/parallel_reader.hpp
```diff
@@ -0,0 +1,960 @@
...
+      bool gotChunk = false;
+      // Each MessageIndex contributes one contiguous, per-channel run of entries.
+      // We record the run boundaries so orderEntries() can k-way MERGE the already-
+      // sorted runs (O(N log C)) instead of a full O(N log N) std::sort.
+      std::vector<std::pair<size_t, size_t>> runs;
+      for (auto rec = rr.next(); rec.has_value(); rec = rr.next()) {
+        if (!rr.status().ok()) {
+          rc->status = rr.status();
+          break;
+        }
```
clalancette: I *think* this condition is unnecessary; in reader.inl, if a record is failed to be read, the loop would have exited.  Checking the status again here isn't bad, but probably unnecessary.

cpp/mcap/include/mcap/parallel_reader.hpp
```diff
@@ -0,0 +1,960 @@
...
+          }
+          if (rc->entries.size() > runStart) {
+            runs.emplace_back(runStart, rc->entries.size());
+          }
+        }
+      }
+      if (rc->status.ok() && !gotChunk) {
+        rc->status = Status{StatusCode::InvalidChunkOffset, "no chunk record at planned offset"};
+      }
+      if (rc->status.ok()) {
```
clalancette: Combined with the last comment, I think this can lead to some data loss.  If we successfully decompress a chunk, in in the next iteration we get a ReadRecord failure (rr->next returns nullopt), then we ignore that error.  I think we need an additional check on rr->status() here to make sure that doesn't happen.

clalancette (commented):
Thanks for the improvements, I've taken a look.  Overall, I like the idea.  Besides the things I've added inline, I have a few overarching things to think about:

1.  All of the other parts of the C++ library use the .hpp/.inl split to split between declarations and implementation.  This new implementation does not.  I think we should follow the existing convention.
2.  From what I can tell, the per-chunk work is handed out to the pool, while the per-message count is still done on the caller's thread.  That suggests to me that this will have the most benefit for large messages with lots of chunks, rather than lots of small messages.  Have you measured the performance of this vs. the default reader with lots of small messages?
3.  I'm somewhat reluctant to commit to maintaining another reader implementation.  What do you think it would take to consider making this the one and only reader implementation?

cpp/mcap/include/mcap/byte_semaphore.hpp
facontidavide: it will be in a follow up PR

cpp/mcap/include/mcap/parallel_budget.hpp
facontidavide: fixed

cpp/mcap/include/mcap/parallel_reader.hpp
facontidavide: fixed

cpp/mcap/include/mcap/parallel_reader.hpp
facontidavide: removed

facontidavide:
I think the two readers don't really overlap, since some platform may not support what the parallel one requires.

But the final decision is up to you.
I addressed the other comments

PR #1709, opened 2026-06-10
cli: Document doctor/recover bugs as fixed in rust-cli

amacneil:
### Changelog

**CLI:**

- mcap doctor: no longer crashes (SIGSEGV) when examining files with malformed summary records. The Go CLI logs the parse error but continues to dereference the nil result, causing a segmentation fault on any file with a truncated or corrupt ChunkIndex, AttachmentIndex, Schema, Channel, Message, Metadata, MetadataIndex, or Header record.
- mcap recover: always produces a valid, readable output file. The Go CLI's writer can produce internally inconsistent ChunkIndex records (declared byte length exceeds actual written bytes) when the channel registration order doesn't match the chunk contents.

### Docs

None

### Description

Adds two `knownDifference` cases to the CLI conformance suite showing that Go doctor crashes (nil-dereference → SIGSEGV) on files with truncated summary records, while Rust doctor reports the error and continues. Both cases use hand-built 88-byte MCAP fixtures with a truncated record in the summary section.

- **`doctor-crash-on-malformed-chunk-index`**: truncated ChunkIndex triggers nil deref at `doctor.go:489`.
- **`doctor-crash-on-malformed-attachment-index`**: truncated AttachmentIndex triggers nil deref at `doctor.go:498`. Same pattern affects most record types in the Go doctor loop.

Fixes FG-14850


amacneil:
> is there a tracking issue to actually fix the Go doctor nil-derefs and the WriteChunkIndex length bug, or is the plan to let them rot now that Rust is the shipping CLI

Plan is to let them rot, Rust CLI will ship imminently

PR #1706, opened 2026-06-10
cpp: Migrate to Conan 2

amacneil:
### Changelog

None

### Docs

Updated `AGENTS.md` and `cpp/README.md` to reference Conan 2.x.

### Description

Migrates the C++ build from Conan 1 to Conan 2, for compatibility with the latest GitHub Actions runners (see https://github.com/foxglove/mcap/pull/1707):

- Rewrote all five `conanfile.py` files for the Conan 2 API (`CMakeDeps`, `CMakeToolchain`, `package_type`, editable `layout()`)
- Replaced `conanbuildinfo.cmake` / `${CONAN_LIBS}` with `find_package()` and explicit CMake targets in all consumer `CMakeLists.txt` files
- Updated `build.sh` / `build-docs.sh` for Conan 2 CLI (`conan profile detect`, `conan install -of`, `conan build --build=editable`)
- CI, Dockerfiles, and Makefile now install Conan 2.x and cache `~/.conan2/p`
- Updated `protobuf/3.21.1` → `3.21.12` because Conan Center 2 has no recipe for 3.21.1. The newer headers also require relaxing `-Werror=shadow` and `-Werror=pedantic` for generated protobuf code in the examples.
- `cpp-windows` CI jobs are updated to `windows-latest` (including the VS 2026 image).

PR #1738, opened 2026-06-16
doc: Fix C++ docs

gasmith:
### Changelog
None

### Docs
None

### Description
In #1706, we migrated to Conan 2, and broke the publication of C++ docs
in the process. The problem is that hdoc only documents symbols declared
in non-system headers, and that commit changed how the mcap headers are
treated.

This change updates cmake to stop treating imported headers as system
headers. It does that for all imported targets (including lz4 and zstd),
but that's OK, because hdoc filters symbols by directory (/mcap/cpp).

### Links
Fixes: FG-14880


gasmith:
Link to C++ docs in the preview deployment: https://03e6eb7c.mcap.pages.dev/docs/cpp/

james-rms (approved):

PR #1739, opened 2026-06-18
python: read chunks without message indexes

### LINKED ISSUE #1735: python: Reader returns no messages with chunk index and without message index
Using the python reader, I get no messages form the iter_messages function, when the MCAP file has a chunk index but no message index.

(We want to do this, to safe some storage with many small messages for slightly slower read access, but still want some kind of stream reader, to not load the multi GB file into storage)

Reading with the rust client and mcap-cli cat seems to work.

- Version: python mcap 1.3.1
- Platform: MacOS

**Steps To Reproduce**
First run this to create a file without message index.
```python
import json
from time import time_ns

from mcap.writer import Writer, IndexType

with open("no-msg-idx.mcap", "wb") as stream:
    writer = Writer(stream, index_types = IndexType.CHUNK)
    # writer = Writer(stream) # <- this works

    writer.start()

    channel_id = writer.register_channel(
        schema_id=0,
        topic="sample_topic",
        message_encoding="json",
    )

    writer.add_message(
        channel_id=channel_id,
        log_time=time_ns(),
        data=json.dumps({"sample": "test"}).encode("utf-8"),
        publish_time=time_ns(),
    )

    writer.finish()

``` 
Then run this to see the message output. The output will be empty.
```python

with open("no-msg-idx.mcap", "rb") as f:
    reader = make_reader(f)
    for schema, channel, message in reader.iter_messages(topics=["sample_topic"]):
        print(f"{channel.topic} ({schema.name}): {message.data}")

```

**Expected Behavior**
I would expect the reader to still know which chunk to look up, but then scan the full chunk for new messages instead of just returning no messages.

amacneil:
### Changelog

Python readers now return messages from chunk-indexed MCAP files that omit message indexes.

The Python `mcap` package version is bumped to `1.4.0`.

### Docs

None

### Description

Fixes #1735.

Fixes the Python seeking reader's chunk prefilter so a chunk with no message index offsets is still scanned after passing time-range checks. This preserves indexed time narrowing while allowing the existing chunk scan to filter topics and timestamps.


jtbandes:
I feel like the conformance tests should have caught this? Looks like they may be disabled for certain test variants for the python indexed reader - could you make sure that the appropriate ones are enabled and verify that the conformance tests would catch this issue if you hadn't made a fix?

jtbandes:
What's the reasoning for this being a minor release rather than a patch release?

amacneil:
> I feel like the conformance tests should have caught this?

Yes. Apparently they are not enabled, "the Python indexed reader runner explicitly skips every chunk-index-only variant". I don't want to fix that here though, python mcap library is not long for this world.

> What's the reasoning for this being a minor release rather than a patch release?

There is another unreleased commit to python library (change to header library string) that warrants a minor bump.

PR #1742, opened 2026-06-22
perf(cpp): skip LZ4 frame checksums (XXH32) for a measurable throughp…

facontidavide:
…ut gain

Profiling a point-cloud-heavy MCAP showed LZ4_XXH32_update consuming more CPU than the decompression itself (~12% vs ~9% of total application CPU on a GUI thread). Skipping the frame-level block and content checksum via LZ4F_decompressOptions_t.skipChecksums eliminates that cost — roughly a 41% reduction in total LZ4 decode CPU.

Structural corruption is still caught by the size/consumption checks already present in decompressAll(): wrong output size (dstSize != uncompressedSize), partial consumption (srcSize != compressedSize), and LZ4F_isError on the return code. The XXH32 integrity hash is therefore redundant for callers that trust the source (local files, verified network transports).

### Description

</td><td>

</td></tr></table>


james-rms (commented):
I think this makes sense to control via an option on the MCAP library, but might be unexpected as a default behavior.

PR #1746, opened 2026-06-26
cli: reduce compress chunk default

fgDevDarwin:
### Changelog
`mcap compress` now defaults to a 768 KiB target chunk size.

### Docs

None

### Description

`mcap compress` previously inherited the shared writer default chunk size of 1 MiB. For index-in-place use cases, smaller chunks avoid requiring readers to fetch and decompress larger ranges than needed. This changes only the `compress` command default to 768 KiB while preserving explicit `--chunk-size` behavior and other writer command defaults.

Validated with `cargo test -p mcap-cli` and `cargo fmt --all -- --check`. `cargo clippy -p mcap-cli --all-targets -- --no-deps -D warnings` could not run because this environment does not have the `clippy` cargo command installed and does not include `rustup` to add it.

`mcap compress` defaulted `--chunk-size` to 1048576 bytes.

</td><td>

`mcap compress` defaults `--chunk-size` to 786432 bytes.

</td></tr></table>

### DIFF
--- a/rust/cli/src/cli.rs
+++ b/rust/cli/src/cli.rs
@@ -31,6 +31,8 @@ pub(crate) static LIBRARY_IDENTIFIER: LazyLock<String> = LazyLock::new(|| {
     )
 });
 
+pub(crate) const DEFAULT_COMPRESS_CHUNK_SIZE: u64 = 768 * 1024;
+
 #[derive(Parser, Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


amacneil:
We only just standardized on 1 MiB across the board:

https://github.com/foxglove/mcap/pull/1659

I don’t see why we would want to special case the cli compress command.

PR #1754, opened 2026-07-03
cli: add topic mode to rewrite --order

amacneil:
## Changelog

- Add a `topic` mode to the `--order` option (available on `mcap sort`, `filter`, `compress`, and `decompress`). It groups each channel's messages together — channels ordered by topic name, then channel ID, and each channel's messages by ascending log time — and places every channel in its own chunk(s). Reading a single topic then touches one contiguous region of the file instead of scanning the whole file.
- Defaults are unchanged: `sort` still orders by `log_time`; `filter`/`compress`/`decompress` still `preserve` the input order.

## What changed

Adds `topic` to the `MessageOrder` enum backing the shared rewrite `--order` flag, and implements it in the rewrite engine's single-input pipeline:

- **Indexed inputs:** `preserve` and `log_time` stream straight to the writer (one chunk resident at a time). `topic` is not an order the index can produce, so its messages are read in log-time order, buffered, and re-sorted by `(topic, channel_id, log_time)` before writing.
- **Linear (summaryless) inputs:** `topic` uses the same buffer-and-sort path already used for `log_time`.
- **Chunk isolation:** for `topic`, the writer's chunk is flushed at every `channel_id` boundary, so no chunk ever mixes two channels. This is *in addition to* the normal size-based roll-over — chunks still break when they reach the `--chunk-size` target (default 1 MiB uncompressed), so a large channel still spans multiple chunks. Standardized record placement (metadata first, attachments last) is unchanged.
- Sorts are stable, so messages with equal keys keep their reader/file order, matching the indexed reader's "earlier record wins" tie-break.

## Why

In a time-ordered file, a single topic's messages are scattered across every chunk, so reading one topic requires downloading the whole file or issuing many range requests. Grouping each channel into its own contiguous chunk(s) lets a reader fetch one topic in a single contiguous byte range, which is a significant win for remote/range-request reads and improves per-topic compression locality.

## Not included

- **Memory-bounded `topic` ordering.** `topic` holds the selected message set in memory while sorting (as the summaryless `log_time` path already does). A streaming/external-sort implementation is intentionally out of scope for this change and left as a follow-up; the in-memory approach keeps this change small and matches the existing ordered-path behavior.
- **`merge --order`.** The `--order` flag (including `topic`) is only on the single-input commands. `merge` is unchanged and continues to output in log-time order.

## Walkthrough

`mcap sort --order topic` on a 600-message, 3-topic file. Before, the 7 chunks each mix all three topics; after, there are 3 chunks — one topic each — every chunk spanning the full time range, so `/camera` is a single contiguous region.

[sort_order_topic_e2e.log](https://cursor.com/agents/bc-f828ef94-25ca-4cbf-8cc2-08dbeb3870ad/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fsort_order_topic_e2e.log)

```
BEFORE (log-time ordered):
  chunk 0..6: each channels=3 topics=["/camera","/gps","/imu"]

AFTER (--order topic):
  chunk 0: offset=56   time=[0..199] channels=1 topics=["/camera"]
  chunk 1: offset=4797 time=[0..199] channels=1 topics=["/gps"]
  chunk 2: offset=9425 time=[0..199] channels=1 topics=["/imu"]
```

## Testing

- ✅ `cargo test -p mcap-cli` — 348 unit + 13 integration tests pass, including topic grouping and within-channel ordering across indexed and both linear input shapes, topic chunk isolation, and `--order` argument parsing.
- ✅ `cargo clippy -p mcap-cli --all-targets -- --no-deps -D warnings`
- ✅ `cargo fmt --all -- --check`
- ✅ Manual end-to-end run of the built `mcap` binary (see walkthrough): 7 mixed chunks become 3 single-topic chunks, each spanning the full time range, with all 600 messages preserved.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-a3f3c26b-9438-41dd-9753-a204e702325e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a3f3c26b-9438-41dd-9753-a204e702325e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


amacneil:
> Many tiny chunks: topic flushes a chunk per channel (rust/cli/src/rewrite/single.rs:299), so a file with many low-volume topics turns into as many chunks — chunk-index/summary bloat and worse per-chunk compression

This is an accepted trade-off, since its the whole point of the flag.

> My two open inline threads still stand (topic buffering / bounded-memory), and ae5243347 removed the "known follow-up" note that used to flag it inline, so the RAM cost is now less visible in the code, not fixed.

Implementing in-memory approach in this PR, moving to an external sort is future work. Prefer to keep this in the PR comments rather than code comments.

PR #1757, opened 2026-07-04
cli: reimplement sort as a rewrite-engine preset

amacneil:
### Changelog

- `mcap sort` gains an `--order <preserve|log_time>` flag (default `log_time`; `preserve` keeps the input's stored message order)
- `mcap sort` output flag is now `-o/--output` to match the other rewrite commands, and omitting the output writes to stdout. Across all rewrite commands (`filter`/`compress`/`decompress`/`sort`), the old `--output-file` name is now accepted as a deprecated alias for `--output` (it warns).
- Sorted output now uses a standardized record layout: metadata at the start of the file (immediately after the header), and attachments at the end of the file (just before the data end record).

### Docs

None. The new `--order` flag is documented in `sort --help`.

### Description

`sort` was ~600 lines of bespoke read / sort / copy / validate code that duplicated the engine's indexed and linear read paths. It is now a thin adapter — like `compress`/`decompress` — that builds `RewriteOptions` and calls `rewrite::run`. The engine already handles indexed vs. summaryless reading, log-time ordering (indexed via `ReadOrder::LogTime`, summaryless via a buffered sort), and the standardized record placement.

`SortCommand` flattens the shared `CommonRewriteArgs` (adding only `--compression`, `--no-chunks`, and its own `--order`), so all four rewrite commands share one definition for the truly common flags. Two knock-on points:

- **`--order` is per-command, by design.** It needs a different default (`log_time` for `sort`, `preserve` for the copy commands), and a single flattened field can't show a per-command default in `--help`. Rather than trade away that clarity, each command declares its own `--order` with the correct default; the flag name, values, and help text are identical across commands, so the user surface stays aligned. `--order` stays overridable everywhere, so future modes (e.g. `publish_time`) apply to `sort` too. `sort` intentionally does not take `filter`'s input-narrowing flags (topic/time selection, `--exclude-*`): `filter` is the superset for reducing a file, while `sort` only reorders it.
- **Output flag.** Every other rewrite command plus `recover`/`get attachment` use `-o/--output`; only `merge` and `sort` used `--output-file`, a historical inconsistency. `sort` now uses `-o/--output`, and the deprecated `--output-file` alias lives on `CommonRewriteArgs`, so `filter`/`compress`/`decompress`/`sort` all accept it and warn uniformly (mirroring how `filter` deprecated `--output-compression`). The short `-o` is unchanged. (`merge` is a separate, multi-input command and is left as-is.)

Behavior notes:
- `sort` now emits the standardized layout (metadata first → messages → attachments last) instead of passing records through in place, so anyone byte-diffing `sort` output or relying on physical record order will see the change.
- `sort`'s output is no longer required: like the other rewrite commands it now streams to stdout when `--output` is omitted (and can read from stdin when the input path is omitted).
- The summaryless ordered path uses the engine's buffered linear sort. Making that path memory-bounded (via a synthesized chunk index) is intentionally left to a separate PR.

### Testing

- [x] `mcap sort <out-of-order.mcap> -o sorted.mcap` then `mcap doctor --strict-message-order sorted.mcap` exits `0`, while the same check on the input exits `1`.
- [x] `mcap sort --order preserve <out-of-order.mcap> -o preserved.mcap` leaves the messages in their original (unsorted) order.
- [x] `mcap sort <in.mcap> > out.mcap` (no `-o`) streams a valid sorted file to stdout.
- [x] `mcap <filter|compress|decompress> --help` shows `--order [default: preserve]` and `mcap sort --help` shows `--order [default: log_time]`.
- [x] `mcap <filter|compress|decompress|sort> <in.mcap> --output-file out.mcap` still works but prints a deprecation warning; `mcap sort --help` shows `-o/--output` and not `--output-file`.
- [x] A sorted file keeps its metadata and attachments, with metadata placed before the message chunks and attachments after them.

<div><a href="https://cursor.com/agents/bc-22387126-660e-4ae3-b765-eef465e4fd50"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-22387126-660e-4ae3-b765-eef465e4fd50"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/rust/cli/src/cli.rs
+++ b/rust/cli/src/cli.rs
@@ -134,12 +134,6 @@ pub struct CommonRewriteArgs {
     /// Disable all output CRC fields
     #[arg(long = "no-crc", default_value_t = false)]
     pub no_crc: bool,
-
-    /// Message order in the output: preserve (keep the input order) or log_time.
-    ///
-    /// Defaults to preserve, except `sort`, which defaults to log_time.
-    #[arg(long = "order", value_enum)]
-    pub order: Option<MessageOrder>,
 }
 
 impl CommonRewriteArgs {
@@ -166,12 +160,20 @@ pub struct CompressCommand {
     /// Compression algorithm for output file: zstd, lz4, or none
     #[arg(long = "compression", default_value = "zstd")]
     pub compression: String,
+
+    /// Message order in the output: preserve (keep the input order) or log_time
+    #[arg(long = "order", value_enum, default_value = "preserve")]
+    pub order: MessageOrder,
 }
 
 #[derive(clap::Args, Debug, PartialEq, Eq)]
 pub struct DecompressCommand {
     #[command(flatten)]
     pub common: CommonRewriteArgs,
+
+    /// Message order in the output: preserve (keep the input order) or log_time
+    #[arg(long = "order", value_enum, default_value = "preserve")]
+    pub order: MessageOrder,
 }
 
 #[derive(clap::Args, Debug, PartialEq, Eq)]
@@ -535,6 +537,10 @@ pub struct FilterCommand {
     /// Write records outside of chunks
     #[arg(long = "no-chunks", default_value_t = false)]
     pub no_chunks: bool,
+
+    /// Message order in the output: preserve (keep the input order) or log_time
+    #[arg(long = "order", value_enum, default_value = "preserve")]
+    pub order: MessageOrder,
 }
 
 #[derive(clap::Args, Debug, PartialEq, Eq)]
@@ -580,6 +586,13 @@ pub struct SortCommand {
     /// Write records outside of chunks
     #[arg(long = "no-chunks", default_value_t = false)]
     pub no_chunks: bool,
+
+    /// Message order in the output: preserve (keep the input order) or log_time.
+    ///
+    /// `sort` defaults to log_time; it accepts the same flag as the other rewrite commands so it
+    /// can be overridden (and future modes such as publish_time will apply here too).
+    #[arg(long = "order", value_enum, default_value = "log_time")]
+    pub order: MessageOrder,
 }
 
 pub type InfoCommand = FileCommand;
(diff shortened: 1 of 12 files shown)


rust/cli/src/cli.rs
```diff
@@ -403,15 +423,10 @@ pub struct DuCommand {
...
 #[derive(clap::Args, Debug, PartialEq, Eq)]
-pub struct FilterCommand {
-    /// Input MCAP file path. If omitted, reads from stdin.
-    pub file: Option<PathBuf>,
-
-    /// Output file path. If omitted, writes to stdout.
-    #[arg(short = 'o', long = "output")]
-    pub output: Option<PathBuf>,
-
+pub struct TranscodeArgs {
```
amacneil: maybe this should be rewriteargs to match the rewrite module name?

rust/cli/src/cli.rs
```diff
@@ -310,6 +310,26 @@ pub enum CompressionFormat {
...
+        match self {
+            CompressionFormat::Zstd => "zstd",
+            CompressionFormat::Lz4 => "lz4",
+            CompressionFormat::None => "none",
+        }
+    }
+}
+
+#[derive(clap::ValueEnum, Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum MessageOrder {
```
amacneil, line 359: wdyt about calling this OrderBy or something to match the option name

PR #1762, opened 2026-07-05
cli: add --order preserve|log_time to filter/compress/decompress

amacneil:
### Changelog

- `mcap filter`, `mcap compress`, and `mcap decompress` gain `--order <preserve|log_time>` (default `preserve`): keep the input's stored message order, or sort messages by log time. (`log-time` is accepted as an alias for `log_time`.)
- `mcap compress` and `mcap decompress` no longer silently re-sort an out-of-order **indexed** file to log time — they now preserve the input's message order by default (already-ordered files are unaffected). Pass `--order log_time` to sort. Unindexed files were never automatically sorted.
- `--order log_time` can now sort **summaryless** (unindexed) files, which these commands previously always copied in stored order.

### Docs

None. The new flag is documented in each command's `--help`.

### Description

The shared rewrite engine (used by `filter`, `compress`, and `decompress`) routes each input down one of two paths: an **indexed** path (input has usable chunk indexes) or a **linear** path (summaryless input — chunked or unchunked). Previously the indexed path always read with `ReadOrder::LogTime` (re-sorting), while the linear path always streamed in stored order. So message ordering was an implicit, path-dependent side effect. This PR makes ordering an explicit `--order` knob and changes the indexed path's default to preserve.

**Behavior by input type (before → after):**

| Input type | Before (on `main`) | After — `--order preserve` (default) | After — `--order log_time` |
| --- | --- | --- | --- |
| Indexed (chunk indexes present) | re-sorted to log time | **preserves stored order** (change) | sorts to log time (same as old default) |
| Chunked, summaryless | preserved (stored order) | preserved (unchanged) | **sorts to log time** (newly possible) |
| Unchunked | preserved (stored order) | preserved (unchanged) | **sorts to log time** (newly possible) |

So the only behavior change at the default is for indexed inputs (they were silently re-sorted, now preserved); summaryless inputs were already preserved and stay that way. `--order log_time` reproduces the old default for indexed inputs and newly enables sorting summaryless inputs (which the rewrite commands could not sort before).

Implementation:
- **Indexed path:** `preserve` reads via `ReadOrder::File`; `log_time` via `ReadOrder::LogTime`. Time-range and topic filtering are applied per-message and are unaffected by read order.
- **Linear (summaryless) path:** `preserve` streams messages in stored order (unchanged); `log_time` buffers the selected messages and stably sorts them by log time (equal log times keep file order). Buffering the whole selected set is a documented memory caveat that matches `sort`'s existing linear path (`copy_linear_messages_in_log_time_order`) — making the summaryless ordered path memory-bounded is a known follow-up that should land once in a shared place.

Ordering is carried end-to-end as the `MessageOrder` enum (`RewriteOptions`/`ResolvedOptions` hold the enum, not a bool), and the engine maps it with exhaustive `match`es, so future order modes extend one enum and the compiler flags every site that must handle a new variant.

For conformant (already log-time-ordered) files the output is unchanged; only genuinely out-of-order indexed inputs differ at the default. The `sort` command is intentionally left untouched.

Design notes:
- The canonical flag value is `log_time` (matching the MCAP spec's serialized field name, and the Python/Rust field name); `log-time` is accepted as an alias.
- `compress`/`decompress` get `--order` but no selection flags — `filter` is the place for selecting + transcoding.
- The `--last-per-channel-topic-regex` seed messages remain a log-time-sorted preamble ahead of the window regardless of `--order` (documented in-code): they are a log-time-window state snapshot relocated from before the window, so there is no stored order to preserve for them.

Follow-ups (separate PRs): completing `filter` as a transcode superset (`--no-crc`/`--no-chunks`); and memory-bounded summaryless ordered rewrite (synthesized chunk index for chunked inputs, `(log_time, offset)` pointer list for unchunked), landed alongside porting `sort` onto the engine.

### Testing

- [x] `cargo test -p mcap-cli` — engine tests cover indexed + linear (chunked/unchunked) `preserve` vs `log_time`, stable tie-break for equal log times, and the shared-engine default (`RewriteOptions::new`, used by compress/decompress) preserving order on an out-of-order indexed input; plus CLI parse tests for `--order` (both spellings) on filter/compress/decompress and invalid-value rejection.
- [x] On an out-of-order file: `mcap filter in.mcap -o out.mcap` (default) keeps input order; `mcap filter --order log_time` / `mcap compress --order log_time` / `mcap decompress --order log_time` produce output that passes `mcap doctor --strict-message-order`.
- [x] Confirm `mcap compress in.mcap -o out.mcap` no longer re-sorts an out-of-order indexed input.

<div><a href="https://cursor.com/agents/bc-3ea18c6d-204e-47ad-875c-3700e26cf87f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3ea18c6d-204e-47ad-875c-3700e26cf87f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/rust/cli/src/cli.rs
+++ b/rust/cli/src/cli.rs
@@ -415,7 +415,7 @@ pub enum MessageOrder {
     #[value(name = "preserve")]
     Preserve,
     /// Sort messages by log time.
-    #[value(name = "log_time", alias = "log-time")]
+    #[value(name = "log-time", alias = "log_time")]
     LogTime,
 }
 
@@ -498,7 +498,7 @@ pub struct FilterCommand {
     #[arg(long = "chunk-size", default_value_t = mcap::WriteOptions::DEFAULT_CHUNK_SIZE)]
     pub chunk_size: u64,
 
-    /// Message order in the output: preserve (keep the input's stored order) or log_time (sort by
+    /// Message order in the output: preserve (keep the input's stored order) or log-time (sort by
     /// log time). See `mcap sort` for sorting by log time as a standalone operation.
     #[arg(long = "order", value_enum, default_value = "preserve")]
     pub order: MessageOrder,
--- a/rust/cli/src/cli.rs
+++ b/rust/cli/src/cli.rs
@@ -310,6 +310,18 @@ pub enum CompressionFormat {
     None,
 }
 
+/// Message ordering applied by the file-rewriting commands.
+#[derive(clap::ValueEnum, Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum MessageOrder {
+    /// Keep the input's stored message order.
+    #[default]
+    #[value(name = "preserve")]
+    Preserve,
+    /// Sort messages by log time.
+    #[value(name = "log_time", alias = "log-time")]
+    LogTime,
+}
+
 #[derive(clap::Args, Debug, PartialEq, Eq)]
 pub struct ConvertCommand {
     /// Local path to the input file
@@ -481,6 +493,10 @@ pub struct FilterCommand {
     /// Target uncompressed chunk size for output
     #[arg(long = "chunk-size", default_value_t = mcap::WriteOptions::DEFAULT_CHUNK_SIZE)]
     pub chunk_size: u64,
+
+    /// Message order in the output: preserve (keep the input's stored order) or log_time (sort by log time)
+    #[arg(long = "order", value_enum, default_value = "preserve")]
+    pub order: MessageOrder,
 }
 
 #[derive(clap::Args, Debug, PartialEq, Eq)]
--- a/rust/cli/src/cli.rs
+++ b/rust/cli/src/cli.rs
@@ -127,7 +127,7 @@ pub struct CompressCommand {
     #[arg(long = "compression", default_value = "zstd")]
     pub compression: String,
 
-    /// Message order in the output: preserve (keep the input's stored order) or log_time
+    /// Message order in the output: preserve (keep the input order) or log_time
     #[arg(long = "order", value_enum, default_value = "preserve")]
     pub order: MessageOrder,
 }
@@ -145,7 +145,7 @@ pub struct DecompressCommand {
     #[arg(long = "chunk-size", default_value_t = mcap::WriteOptions::DEFAULT_CHUNK_SIZE)]
     pub chunk_size: u64,
 
-    /// Message order in the output: preserve (keep the input's stored order) or log_time
+    /// Message order in the output: preserve (keep the input order) or log_time
     #[arg(long = "order", value_enum, default_value = "preserve")]
     pub order: MessageOrder,
 }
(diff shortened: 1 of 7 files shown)


amacneil:
> non-blocking and opt-in only: the summaryless --order log-time path buffers the whole selected message set in memory

Will address this in a follow up PR

rust/cli/src/rewrite/options.rs
```diff
@@ -27,6 +28,9 @@ pub(crate) struct RewriteOptions {
     pub(crate) output_compression: String,
     pub(crate) chunk_size: u64,
     pub(crate) use_chunks: bool,
+    /// Message order applied to the output (e.g. preserve the input's stored order, or sort by log
+    /// time).
```
amacneil: weird that this is the only option with a comment

rust/cli/src/rewrite/options.rs
```diff
@@ -115,6 +126,9 @@ pub(crate) struct ResolvedOptions {
     pub(crate) compression: Option<mcap::Compression>,
     pub(crate) chunk_size: u64,
     pub(crate) use_chunks: bool,
+    /// Message order applied to the output (e.g. preserve the input's stored order, or sort by log
+    /// time).
```
amacneil: weird that this is the only option with a comment. either we should document all or none of them

rust/cli/src/cli.rs
```diff
@@ -75,13 +75,19 @@ pub enum Command {
     #[command(verbatim_doc_comment)]
     Completion(CompletionCommand),
     /// Create a compressed copy of an MCAP file
+    ///
+    /// Messages are copied in the input's stored order by default; pass `--order log-time` (or use
+    /// `mcap filter`) to reorder.
```
amacneil: Is this necessary to call out explicitly? We don't mention any other defaults here

rust/cli/src/cli.rs
```diff
@@ -75,13 +75,19 @@ pub enum Command {
...
     Compress(CompressCommand),
     /// Convert supported input files to MCAP
     #[command(
         long_about = "Convert supported input files to MCAP.\n\nSupported inputs:\n  .bag  ROS 1 bag\n  .db3  ROS 2 SQLite db3"
     )]
     Convert(ConvertCommand),
     /// Create an uncompressed copy of an MCAP file
+    ///
+    /// Messages are copied in the input's stored order by default; pass `--order log-time` (or use
+    /// `mcap filter`) to reorder.
```
amacneil: Is this necessary to call out explicitly? We don't mention any other defaults here

Issue #1642, opened 2026-05-07
mcap cli doesn't output default field values of protobuf attributes


matawed:
**Description**

When using mcap cat --json, protobuf3 fields that have their default value (e.g., 0 for enums/integers, "" for strings  see [doc](https://protobuf.dev/programming-guides/proto3/#default) for reference) are not shown in output of mcap cli. This is inconsistent with how other protobuf libraries (e.g., Python) handle default values, where the actual default is returned.

- Version:  mcap cli version 0.0.51
- Platform: Ubuntu 24.04

**Steps To Reproduce**

mcap cat --topics /topic --json | jq ".attribute"
Result: null

output without jq looks like the following

```
{"topic":"/example/road","sequence":0,"log_time":1.000000000,"publish_time":1.000000000,"data":{"description":"field at default value"}}
{"topic":"/example/road","sequence":0,"log_time":2.000000000,"publish_time":2.000000000,"data":{"eventDataQualifier":"VALID", "description":"field at non-default value"}}
```

Reading the same message with the Python mcap-protobuf-support library returns the expected default enum value (e.g., 0 or the enum's first variant name).

**Expected Behavior**

Fields with their proto3 default value should be emitted with their actual default (e.g., 0, "", false) rather than null.

james-rms:
Do you have a concrete use-case for this functionality? The MCAP CLI behavior matches how the Go protobuf library behaves. I'm not sure if the python library is more authoritative.

matawed:
I mean, if I postprocess the output of mcap cli with e.g. jq to gain insights on the data, it looks strange that default value attributs are the reported as "non existing".

E.g.  for the mcap content mentioned in description of issue:

mcap cat --topics /topic --json | jq ".data.description"

I'd see null values, which might confuse users as they might think there is no corresponding data in mcap, but indeed it is, but it's not output by mcap cli.

At least, this output confused me and I'd say it's rather unexpected. As you mentioned it might be because of used protobuf lib, but it doesn't reflect the actual value of the attribute.

matawed:
Still I think this is a bug, as if you're not used to this handling of default value data with mcap cli, you might end up with the wrong conclusion. 

E.g. one colleague of mine used mcap cli to extract the value of an structured attribute. The attribute was a structure with only booleans in it. 

Then he used mcap cli to extract that attribute and it just reported `{}` for something similar like `mcap cat --topics /topic --json | jq ".data.description"` (assuming description is a structure with only booleans in it)

[...]

PR #1764, opened 2026-07-06
cli: print protobuf default values instead of omitting them in output

matawed:
### Description

</td><td>

</td></tr></table>

issue: https://github.com/foxglove/mcap/issues/1642


amacneil:
thanks!

Issue #1697, opened 2026-06-06
cli: json output


amacneil:
It would be nice if the CLI had a global option for json output, making it easier to parse from scripts

Issue #1673, opened 2026-06-04
cli: Standardize Output Path and Overwrite Behavior


amacneil:
_Context: uncovered while porting the MCAP CLI from Go to Rust. The Rust CLI currently mirrors the Go CLI behavior for parity; this tracks whether to diverge before 1.0._

Current Go-compatible behavior overwrites existing output files for commands such as `convert`, `filter`, `merge`, `recover`, and `sort`. Current output path syntax is inconsistent: `convert` uses a positional output path, while other output-producing commands generally use `-o` / `--output` or `--output-file`.

Before Rust CLI 1.0:

- Decide whether output-producing commands should fail when the output path exists unless the user passes `--force` / `-f`.
- Decide whether Rust CLI output paths should standardize on positional arguments or `-o` / `--output` flags, then apply that consistently.
- Decide whether stdin/stdout should use explicit `-` path arguments instead of Go-compatible implicit stdin/stdout behavior (for example, omitted output path plus redirected stdout).
- Write output-producing commands through a temporary sibling file and rename it into place only after successful completion, so failures do not leave a partial output at the requested path.
- Apply the policy consistently so users do not have to remember separate overwrite behavior per command.

Issue #1672, opened 2026-06-04
cli: Decide Topic-Filter Argument Shape and Apply It Consistently


amacneil:
_Context: uncovered while porting the MCAP CLI from Go to Rust. The Rust CLI currently mirrors the Go CLI behavior for parity; this tracks whether to diverge before 1.0._

Current Go-compatible `cat --topics` accepts one comma-separated string.

Before Rust CLI 1.0:

- Decide whether topic filtering should use the comma list, repeatable flags such as `--topic foo --topic bar`, or support both.
- Apply the chosen shape consistently across commands that filter by topic so users do not have to learn separate `cat` and `filter` conventions.

Issue #1671, opened 2026-06-04
cli: Unify --start/--end Time Arguments and Parse RFC3339/Decimal/Unit-Suffixed Values


amacneil:
_Context: uncovered while porting the MCAP CLI from Go to Rust. The Rust CLI currently mirrors the Go CLI behavior for parity; this tracks whether to diverge before 1.0._

Current Go-compatible behavior treats bare numeric `--start` / `--end` values as integer nanoseconds, while `--start-secs` / `--end-secs` select seconds and `--start-nsecs` / `--end-nsecs` select nanoseconds.

This matches MCAP's internal timestamp unit and preserves copy/paste from existing `mcap cat` output, but it is surprising for CLI users.

For Rust CLI 1.0:

- Remove the split `--*-secs` / `--*-nsecs` variants and standardize all commands on `--start` / `--end` where applicable.
- Parse `--start` / `--end` values as RFC3339 timestamps, exact decimal seconds strings (for example `1.23456789`), or explicit unit-suffixed durations/timestamps such as `1.5s`, `250ms`, and `1709146829659264519ns`.
- Apply the same timestamp input syntax consistently across all commands that accept times, including `cat`, `filter`, `add attachment`, and any future time-filtered commands.

PR #1774, opened 2026-07-13
cli: clarify and correct --help text

amacneil:
### Changelog

Clarified and corrected the `mcap` CLI `--help` output.

### Docs

None — the changes are inline `--help` text; no separate documentation updates are needed.

### Description

The `mcap` CLI `--help` output is the primary (and often only) reference a first-time user — including an AI agent driving the tool — has for figuring out how to use it. Reading that surface as such a user turned up descriptions that were ambiguous, inconsistent between sibling commands, or in a few cases outright inaccurate. This PR reworks the clap doc comments / `about` text in `rust/cli/src/cli.rs` to fix those. There are no behavior changes.

Highlights:

- **Corrected misleading/inaccurate text.** `get attachment` was described as "by name or offset" even though `--name` is required and `--offset` only disambiguates duplicate names; `cat --json` listed the wrong encoding identifiers; and the `add` in-place-rewrite corruption warning never actually rendered (the doc comment sat on the args struct, which clap ignores). All corrected.
- **Documented remote inputs.** Every read command accepts remote URLs (`http(s)://` and object-store), yet file arguments said "Local path" and `--allow-remote-scan` was unexplained. File args now read "Path or URL" (except `add`, which is genuinely local-only), and the flag documents its accepted schemes and scope.
- **Named the clock and clarified time filters.** Time options now say they filter on log time, state units (nanoseconds since the Unix epoch, or RFC3339), and explain the end-time sentinel (`0`/omitted = no upper bound).
- **Filled gaps and improved consistency.** Added `merge`/`recover`/`doctor` semantics and exit-code notes, topic-regex matching rules and include/exclude mutual exclusivity, chunk-size units, and de-duplicated the `--coalesce-channels` / `sort --order` rendering.
- **Hid deprecated flags from help.** `filter --start-nsecs` / `--end-nsecs` are still accepted but no longer shown, matching the other hidden deprecated flags.

Every wording change was verified against the implementation, and a separate code-grounded review pass caught and corrected several would-be inaccuracies before they landed.

```
$ mcap get attachment --help
Get an attachment by name or offset
...
      --offset <OFFSET>    Offset of attachment to extract
  -o, --output <OUTPUT>    Location to write attachment bytes
```

</td><td>

```
$ mcap get attachment --help
Get an attachment by name
...
      --offset <OFFSET>    Byte offset of the attachment record (from
                           `mcap list attachments`), used to disambiguate
                           multiple attachments that share the same name
  -o, --output <OUTPUT>    Location to write attachment bytes. If omitted,
                           writes to stdout (refuses if stdout is a terminal)
```

</td></tr></table>

The only user-visible effect beyond wording is that the already-deprecated `--start-nsecs` / `--end-nsecs` no longer appear in `filter --help` (still accepted).

Testing:

- `cargo build -p mcap-cli`, `cargo clippy -p mcap-cli --all-targets -- --no-deps -D warnings`, `cargo test -p mcap-cli` (362 pass)
- Manually inspected `--help` for the top-level command and every subcommand / nested subcommand.

Larger design questions surfaced during the audit (unifying time/topic argument shapes, JSON output for inspection commands, output-overwrite policy, and `get` retrieval-by-offset / duplicate-name handling) are intentionally out of scope here and tracked separately.

Related: foxglove/mcap#1671, foxglove/mcap#1672, foxglove/mcap#1673, foxglove/mcap#1697, foxglove/mcap#1775

<div><a href="https://cursor.com/agents/bc-6a899557-8ed1-4574-9487-5103cceec830"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6a899557-8ed1-4574-9487-5103cceec830"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/rust/cli/src/cli.rs
+++ b/rust/cli/src/cli.rs
@@ -64,55 +64,55 @@ pub struct Args {
 
 #[derive(Subcommand, Debug, PartialEq, Eq)]
 pub enum Command {
-    /// Add records to an existing MCAP file
+    /// Add records to an existing MCAP file.
     Add(AddCommand),
-    /// Concatenate the messages in one or more MCAP files to stdout
+    /// Concatenate the messages in one or more MCAP files to stdout.
     Cat(CatCommand),
-    /// Generate shell completion scripts
+    /// Generate shell completion scripts.
     ///
     /// To load completions in the current shell session:
     ///   bash:       source <(mcap completion bash)
     ///   zsh:        source <(mcap completion zsh)
     ///   fish:       mcap completion fish | source
     ///   powershell: mcap completion powershell | Out-String | Invoke-Expression
-    #[command(verbatim_doc_comment)]
+    #[command(verbatim_doc_comment, about = "Generate shell completion scripts")]
     Completion(CompletionCommand),
-    /// Create a compressed copy of an MCAP file
+    /// Create a compressed copy of an MCAP file.
     ///
     /// Equivalent to running `mcap filter --compression=zstd` (pick another
     /// algorithm with `--compression`).
     Compress(CompressCommand),
-    /// Convert supported input files to MCAP
+    /// Convert supported input files to MCAP.
     #[command(
         long_about = "Convert supported input files to MCAP.\n\nSupported inputs:\n  .bag  ROS 1 bag\n  .db3  ROS 2 SQLite db3"
     )]
     Convert(ConvertCommand),
-    /// Create an uncompressed copy of an MCAP file
+    /// Create an uncompressed copy of an MCAP file.
     ///
     /// Equivalent to running `mcap filter --compression=none`.
     Decompress(DecompressCommand),
-    /// Check an MCAP file structure
+    /// Check an MCAP file structure.
     Doctor(DoctorCommand),
-    /// Compute byte usage statistics for MCAP records
+    /// Compute byte usage statistics for MCAP records.
     Du(DuCommand),
-    /// Copy filtered MCAP data to a new file
+    /// Copy filtered MCAP data to a new file.
     ///
     /// The general-purpose rewrite command: copies selected records to a new
     /// file, optionally selecting by topic and time range and changing
     /// compression, chunking, and message order. `compress`, `decompress`, and
     /// `sort` are presets over this command.
     Filter(FilterCommand),
-    /// Get a record from an MCAP file
+    /// Get a record from an MCAP file.
     Get(GetCommand),
-    /// Report statistics about an MCAP file
+    /// Report statistics about an MCAP file.
     Info(InfoCommand),
-    /// List records of an MCAP file
+    /// List records of an MCAP file.
     List(ListCommand),
-    /// Merge MCAP files
+    /// Merge MCAP files.
     Merge(MergeCommand),
-    /// Recover data from a potentially corrupt MCAP file
+    /// Recover data from a potentially corrupt MCAP file.
     Recover(RecoverCommand),
-    /// Read an MCAP file and write it back with messages reordered (log_time, preserve, or topic)
+    /// Read an MCAP file and write it back with messages reordered (log_time, preserve, or topic).
     ///
     /// Equivalent to running `mcap filter --order=log_time`.
     Sort(SortCommand),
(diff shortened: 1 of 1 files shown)


rust/cli/src/cli.rs
```diff
@@ -64,45 +64,57 @@ pub struct Args {
...
-    /// Check an MCAP file structure
+    /// Check an MCAP file structure.
     Doctor(DoctorCommand),
-    /// Compute byte usage statistics for MCAP records
+    /// Compute byte usage statistics for MCAP records.
     Du(DuCommand),
-    /// Copy filtered MCAP data to a new file
+    /// Copy filtered MCAP data to a new file.
+    ///
+    /// The general-purpose rewrite command: copies selected records to a new
```
james-rms: This reads LLMish to me, can be simpler
```suggestion
    /// Copies selected records to a new
```

rust/cli/src/cli.rs
```diff
@@ -64,45 +64,57 @@ pub struct Args {
...
     ///   fish:       mcap completion fish | source
     ///   powershell: mcap completion powershell | Out-String | Invoke-Expression
-    #[command(verbatim_doc_comment)]
+    #[command(verbatim_doc_comment, about = "Generate shell completion scripts")]
     Completion(CompletionCommand),
-    /// Create a compressed copy of an MCAP file
+    /// Create a compressed copy of an MCAP file.
+    ///
+    /// Equivalent to running `mcap filter --compression=zstd` (pick another
+    /// algorithm with `--compression`).
```
james-rms: I think the parenthetical is a bit confusing (is it talking about adding --compression to `mcap filter` or `mcap compress`?) and really should go in the subcommand `--help` output.
```suggestion
    /// Equivalent to running `mcap filter --compression=zstd`
```

rust/cli/src/cli.rs
```diff
@@ -64,45 +64,57 @@ pub struct Args {
...
     Compress(CompressCommand),
-    /// Convert supported input files to MCAP
+    /// Convert supported input files to MCAP.
     #[command(
         long_about = "Convert supported input files to MCAP.\n\nSupported inputs:\n  .bag  ROS 1 bag\n  .db3  ROS 2 SQLite db3"
```
james-rms: what's the difference between `long_about` here and what goes in the docstring?

rust/cli/src/cli.rs
amacneil: Mmm yeah good call, specifically it feels very Claudish

rust/cli/src/cli.rs
amacneil: Agreed, removing parenthetical

rust/cli/src/cli.rs
amacneil: `long_about` seems to text the command-specific help text (versus the short help text on root `mcap --help`).

However, it appears to be unnecessary, without it clap will take the first line of the docstring, which is what we do for every other command, so will make this consistent.

rust/cli/src/cli.rs
```diff
@@ -477,23 +500,23 @@ pub struct FilterCommand {
     #[command(flatten)]
     pub common: CommonRewriteArgs,
 
-    /// Include topics matching this regex (repeatable)
+    /// Include only topics matching this regex, matched against the full topic name (repeatable; cannot be combined with --exclude-topic-regex)
```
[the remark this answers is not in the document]
amacneil: _(Posted by the Cursor agent on Adrian's behalf.)_

I don't think this is unanchored — the patterns are anchored before matching, so the current wording is accurate. `compile_matchers` wraps each user pattern as `^(?:{pattern})$` before compiling it (`rust/cli/src/rewrite/options.rs:202`). So although `include_topic` calls `Regex::is_match` at line 213, the compiled regex already requires a full-string match: `--include-topic-regex camera` compiles to `^(?:camera)$` and would **not** match `/sensors/camera/image`.

So "matched against the full topic name" is correct. If it's clearer, I can make the anchoring explicit, e.g. "matched against the full topic name (anchored; use `.*` for substring matches)".

Issue #438, opened 2022-06-23
CLI: mcap cat --csv --topics /tf


wkalt:
It would be nice if we could spit out a CSV file with column headers. If a user wanted to look at the history of a single topic in R or python this would make that very easy to do that.

I think it would probably require that the --topics flag be specified with only one topic. Then we could define a JSON->CSV translation and pipe the JSON output to it. Seems easier to work off the newline-delimited JSON output than write new ros/protobuf converters for CSV.

sisaha9:
Has there been any work started on this?

jtbandes:
Currently no work has been started. Could you describe your use case and desired functionality, to aid the discussion?

sisaha9:
We had a similar tool developed for SQL based ROS2 bags. We could pass a ROS2 bag, a topic and an output file name and we would get a CSV file with all the information on that topic. I like the CLI method mentioned here. Would love to contribute on it but not too familiar with Go. If there was a way to do this via the Python API I would like to take a look at it

jtbandes:
You could absolutely achieve this with a Python script. Here are some Python examples to get you started: https://mcap.dev/guides/python/ros2.html

sisaha9:
Do you have any suggestions on how to get the list of attributes for a message (including nested). A DFS could then be used to construct the CSV file

sisaha9:
Or if it's possible to convert to JSON via a Python method

PR #1778, opened 2026-07-14
cli: add CSV output for single-topic cat

tw-ilson:
## Summary

Adds CSV output to `mcap cat` for exporting a single decoded topic with:

```sh
mcap cat demo.mcap --csv --topic /chatter
```

CSV rows are prefixed with `log_time`, `publish_time`, and `sequence`, followed by flattened message fields using dot notation for nested objects and index suffixes for arrays. The CSV output reuses the existing JSON transcoding support, so it supports the same decoded message encodings as `--json`: `jsonschema`, schemaless `json`, `protobuf`, and `ros1msg`.

`--csv` requires `--topic`, `--json` conflicts with `--csv`, and `--topic` is only valid with `--csv` so it is not silently ignored in regular `cat` output.

The CSV header is derived from the first matching message to keep streaming behavior bounded-memory. Later fields not present in that first header are dropped with a warning outcome instead of requiring a full pre-scan. This covers dynamic shapes such as schemaless JSON, permissive JSON schema, maps, and variable-length arrays where the complete CSV column set cannot always be known from the channel schema.

Related: #438

### Future improvements

Nice-to-have follow-ups include multi-topic CSV export, explicit user-provided column lists, and more schema-derived headers for fixed protobuf, ROS1, or strict JSON-schema fields.

## Testing

- `cargo +nightly test -p mcap-cli`
- `cargo +nightly clippy -p mcap-cli --all-targets -- --no-deps -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`


amacneil:
thanks! I've been wanting to better standardize output format, just landed https://github.com/foxglove/mcap/pull/1779 so will pull your changes in next

PR #1779, opened 2026-07-15
cli: add --format=ndjson for cat output

amacneil:
### Changelog

- `mcap cat` now supports `--format=text|ndjson`; use `--format=ndjson` to print one JSON object per message.
- The `--json` flag is deprecated: it still works but prints a warning — use `--format=ndjson` instead.

### Docs

Updated the CLI guide example to use `--format=ndjson`.

### Description

Prerequisite cleanup for [#1697](https://github.com/foxglove/mcap/issues/1697) and [#1778](https://github.com/foxglove/mcap/pull/1778).

Only `mcap cat` already supported `--json`. This adds a `--format` selector and names the existing newline-delimited output `ndjson`, leaving room for a future `--format=json` that emits a single JSON document/array (per the plan in #1697). `--json` is deprecated the same way as other legacy flags (`hide = true`, stderr warning, still accepted) and is now an alias for `--format=ndjson`; passing both `--format` and `--json` conflicts.

The JSON transcoder now dispatches on the channel message encoding (`ros1`, `protobuf`, `json`) rather than the schema encoding, so `--help` and the unsupported-encoding error use the same vocabulary.

[`clispec.dev`](https://clispec.dev/) recommends `--output`/`-o` for format selection, or `--format` when `-o`/`--output` is already used for destination paths — which is the case throughout this CLI — so `--format` is the right spelling here. No other commands gained structured output support.

<div><a href="https://cursor.com/agents/bc-a28d7118-3dc4-434a-9bc1-78ed3b95779a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a28d7118-3dc4-434a-9bc1-78ed3b95779a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #1782, opened 2026-07-15
cli: add global --time-format flag for consistent timestamp rendering

amacneil:
### Changelog

- Added a global `--time-format` flag (`auto`, `rfc3339`, `seconds`, `nanoseconds`) controlling how `mcap cat`, `info`, `list chunks`, and `list attachments` render timestamps. Hidden abbreviation aliases are also accepted: `iso8601` (for `rfc3339`), `s`/`sec`/`secs` (for `seconds`), and `ns`/`nano`/`nanos`/`nsec`/`nsecs` (for `nanoseconds`).
- Timestamps now render consistently across those commands (previously each used a different format).
- **Breaking:** `cat` (text), `list chunks`, and `list attachments` now render timestamps as human-readable `auto` dates instead of raw nanoseconds, and `info` no longer appends the decimal seconds in parentheses. Pass `--time-format=nanoseconds` for the previous integer output.
- **Breaking:** `cat --format=ndjson` emits `log_time`/`publish_time` as quoted strings (RFC3339 by default) instead of bare decimal-seconds numbers — and they are **always** quoted, even under `--time-format=seconds`/`nanoseconds`, so full-precision values are read as text/`BigInt` rather than lossy JSON numbers. There is no bare-number option, so scripts doing arithmetic on `log_time` must parse the string.

### Docs

- New "Timestamp formatting" section in `website/docs/guides/cli.md` documenting the flag, its values/aliases, the human-vs-machine `auto` behavior (including the `1970`-relative ndjson example), that the flag is display-only (rewrite commands ignore it), and how to get numeric output.
- Refreshed the guide's `mcap --help` sample plus the `info`, `cat --format=ndjson`, and `list chunks` outputs to match current behavior, and switched all of the guide's code blocks from 4-space indentation to plain fenced blocks.
- Noted in `rust/cli/AGENTS.md` that changing a top-level command or global flag requires regenerating the `--help` sample in the guide.

### Description

Timestamps were rendered four different ways: `cat` text used raw nanoseconds, `cat --format=ndjson` used decimal seconds, `info` used an `RFC3339 (decimal)` combo, and `list` used raw nanoseconds. This adds one flag and one consistent policy.

Formats for a nanosecond timestamp:

- `auto` (default) — behavior depends on the audience (see below).
- `rfc3339` — RFC3339 UTC, honored literally even below the cutoff (a pre-2000 time renders its true UTC instant, not a fallback). Always uses a fixed 9-digit fractional part, so the field is fixed-width and lexical order equals chronological order (`sort` works without `-n`). Hidden alias: `iso8601` (the emitted strings are a valid ISO 8601 profile).
- `seconds` — decimal seconds, 9 fractional digits. Hidden aliases: `s`, `sec`, `secs`.
- `nanoseconds` — integer nanoseconds. Hidden aliases: `ns`, `nano`, `nanos`, `nsec`, `nsecs`.

**Why `auto` differs for human vs machine output:** human-facing output (`cat` text, `info`, `list`) uses RFC3339 at/above a year-2000 cutoff, else decimal seconds. Recordings with relative/monotonic clocks start near 0, so rendering those as `1970-…` dates would be misleading — decimal seconds is clearer. The RFC3339-vs-decimal choice is latched once per invocation (primed from the recording's start time where known), so a file whose clock crosses the cutoff mid-recording (e.g. GPS time acquired partway through) still renders uniformly instead of flipping row to row. Machine-facing output (`cat --format=ndjson`) instead always uses RFC3339 — see below.

**Why RFC3339 by default for ndjson output:** a machine-readable field should have one predictable shape, so the machine path does not apply the human cutoff (which would emit RFC3339 for some files and decimal seconds for others). Among the fixed options we compared how the common consumers actually parse each one:

- **RFC3339** auto-types into a real datetime column in the tools this output feeds — pandas (`datetime64[ns]`), Apache Arrow, DuckDB (`TIMESTAMP_NS`) — never corrupts the value, and (being fixed-width) sorts lexicographically. Its only downside is that relative recordings render `1970`-relative dates, which still round-trip exactly.
- **nanoseconds** (bare integer) is silently mangled by float-coercing tools (Excel, Google Sheets, MATLAB, R defaults) because a 19-digit integer exceeds their ~15 significant digits.
- **decimal seconds** loses sub-microsecond precision on a naive numeric parse and auto-types as a plain float, not a timestamp.

RFC3339 has the mildest, most-recoverable failure mode of the three, so it is the default; anyone needing a numeric column can opt in with `--time-format=seconds`/`nanoseconds`. This same default should apply to the upcoming `cat --format=csv` output (#1778/#1781), so ndjson and CSV stay consistent — CSV's consumers (spreadsheets, pandas, DuckDB) make the case for RFC3339 even stronger.

**Why ndjson timestamps are quoted strings:** JSON numbers can't safely hold these values — a float loses precision and an integer > 2^53 is truncated by many parsers — so `log_time`/`publish_time` are emitted as quoted strings for every format, letting consumers parse them as text/`BigInt`.

**Why the flag is global:** it mirrors `--color`/`--allow-remote-scan`; commands that don't render timestamps accept but ignore it, and explicit `--time-format` values are honored as-is everywhere.

Related: #1779 added the `cat --format` selector this builds on; #1778/#1781 add `cat --format=csv`, which should adopt the same machine-facing RFC3339 default when it lands (review on #1781 flagged the very cross-format timestamp inconsistency this unifies).

Out of scope (follow-up): aligning `cat`'s time-input flags (`--start-secs`/`--start-nsecs`) with `filter`'s `--start`/`--end` (nanoseconds or RFC3339); timezone/local rendering; showing `publish_time` in `cat` text.

### Testing

- ✅ `cargo test -p mcap-cli`
- ✅ `cargo clippy -p mcap-cli --all-targets -- --no-deps -D warnings`
- ✅ `cargo fmt --all -- --check`
- ✅ Manual end-to-end on generated fixtures (a post-2000 file and a mixed pre/post-cutoff file): text `auto` honors the cutoff + latch; ndjson `auto` always emits RFC3339 (fixed-width, uniform shape, even pre-cutoff); explicit values and every alias (`iso8601`, `s`/`sec`/`secs`, `ns`/`nano`/`nanos`/`nsec`/`nsecs`) are honored across all four commands; invalid values are rejected.

<div><a href="https://cursor.com/agents/bc-019f6610-ad51-7c40-a397-cef9483de935"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-019f6610-ad51-7c40-a397-cef9483de935"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


amacneil:
> The one thing still open before this leaves draft is coordination, not code: this flips the default output of five surfaces at once

Yes it does change the default. `cat --format=ndjson` is really the only one that I think would be likely to consumed by scripts, and it's one extra flag (`--time-format=seconds`) to restore the old behavior.

We're still in 0.x land for the cli, and now is the time to make breaking changes in the name of consistency, before we hit 1.0.

amacneil:
Claude you already asked that and I already answered it.

PR #1787, opened 2026-07-17
cpp: McapWriter: Add scatter-gather write(Message/Attachment, ByteSpanArray)

sebastianlipponer:
### Changelog

cpp: McapWriter: Add scatter-gather `write(Message, ByteSpanArray)` and `write(Attachment, ByteSpanArray)` APIs

### Docs

None

### Description

Add `McapWriter::write(const Message&, const ByteSpanArray&)` and `McapWriter::write(Attachment&, const ByteSpanArray&)` overloads that write a message's or attachment's payload from a list of spans instead of a single contiguous buffer, so callers with non-contiguous payload chunks don't need to copy them into one buffer first. The existing `write(const Message&)` and `write(Attachment&)` overloads are unchanged and now implemented as thin wrappers around the new ones.

### DIFF
--- a/cpp/mcap/include/mcap/internal.hpp
+++ b/cpp/mcap/include/mcap/internal.hpp
@@ -56,6 +56,14 @@ inline uint32_t KeyValueMapSize(const KeyValueMap& map) {
   return (uint32_t)(size);
 }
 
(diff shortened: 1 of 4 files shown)


sebastianlipponer:
Intentionally leaving out tests for now — wanted to get the API reviewed first before building on top of it.

PR #1788, opened 2026-07-17
spec: add capnproto to the format registry

amacneil:
### Changelog

Documented Cap'n Proto as a well-known message encoding and schema encoding in the MCAP format registry.

### Docs

Adds new `capnproto` entries to `website/docs/spec/registry.md` under both "Message encodings" and "Schema encodings".

Also syncs `python/mcap/mcap/well_known.py` — the only language library that mirrors the registry as constants (the others just link to it) — so its enums match: adds the previously-missing `capnproto`, `msgpack`, and `omgidl` values, and points its docstring at the current registry URL.

### Description

MCAP schema registry did not define a well-known encoding for Cap'n Proto messages, so there was no interoperable, writer-agnostic contract. This adds that contract by registering `capnproto` for both the message encoding and the schema encoding, modeled closely on the existing `protobuf` and `flatbuffer` entries.

The registry entry is deliberately just a writer/reader contract over opaque bytes — MCAP remains a container and does not constrain which Cap'n Proto features a payload uses. A few tradeoffs were considered:

- **Schema representation: binary `CodeGeneratorRequest` vs. raw `.capnp` text.** We chose the binary `CodeGeneratorRequest` (defined in `capnp/schema.capnp`, produced by `capnp compile -o/bin/cat`). This mirrors how `protobuf` uses a `FileDescriptorSet` and `flatbuffer` uses a `reflection.Schema`: it is self-contained (all imports are flattened in), it is the canonical artifact every Cap'n Proto code-generator plugin already consumes, and its layout is guaranteed forward/backward compatible upstream. Raw `.capnp` text was rejected because it would force every reader to embed a full schema parser and `import` resolver.

- **Message framing: standard stream framing (unpacked) vs. packed.** We chose the recommended stream framing (segment table + segments), unpacked. It is self-delimiting and mmap/zero-copy friendly, which aligns with MCAP's bounded-memory reading model, and it is what every Cap'n Proto runtime's default writer/reader produces and consumes. Packing was left out because MCAP already provides chunk-level compression (lz4/zstd), so packing would be largely redundant while defeating zero-copy reads. A separate encoding could be added later if there is demand, rather than overloading one name.

- **Encoding name: `capnproto` vs. `capnp`.** We chose `capnproto` to match the registry's convention of spelling out the project name (`protobuf`, `flatbuffer`), rather than the tool/file-extension name.

### Root type identification

The one design point that needs care is how `Schema.name` selects the root type inside `data`. The options:

- **Node `displayName`** (chosen), e.g. `my_schema.capnp:Foo.Bar` — the path passed to `capnp compile` followed by the type's scoped name.
- **The 64-bit node `id`** — Cap'n Proto's stable, canonical type identifier.
- **Scoped name only** (e.g. `Foo.Bar`) — not globally unique across files in a request.

We chose `displayName` for consistency with the `protobuf` and `flatbuffer` schema entries (both use a fully-qualified name string), and because it is human-readable and trivial for writers to produce:

- It is deterministic from inputs the writer already controls — it is literally the compile path plus the scoped type name — so no byte-level inspection of `data` is required.
- Every Cap'n Proto runtime also exposes it directly from the generated type (e.g. `Schema::from<Foo>().getProto().getDisplayName()` in C++, `Foo.schema.node.displayName` in Python), and writers already have such a runtime loaded to build messages.

The known objection to `displayName` (raised previously in #860) is that Cap'n Proto documents it as human-facing and names the 64-bit `id` as the stable key, so it is not guaranteed unique or rename-resistant. Two things make this a non-issue here rather than a reason to switch identifiers:

- **`name` and `data` always travel together in the same Schema record**, produced by the same `capnp compile` invocation, so the lookup is self-contained — cross-toolchain display-name drift cannot desynchronize them.
- **The stable 64-bit `id` is never lost**: every node's `id` is retained inside the `CodeGeneratorRequest`, so a reader that wants to verify it has the correct schema (or detect schema evolution) can recompile its local `.capnp`, read the deterministic ids, and compare — independent of `displayName`.

The only residual risk is a `displayName` collision within a single request, which requires a pathological layout (two files sharing a display name, e.g. via `--src-prefix`). We close that in the spec by requiring `name` to match a single struct node exactly, making it a well-defined error rather than a silent mis-resolution — enforced reader/validator-side so writers are not burdened.

One case none of the three options can express is a generic type. Cap'n Proto erases generics at runtime, and an instantiation is not a node in the request: compiling `struct Wrapper(T)` alongside `struct Concrete { w @0 :Wrapper(Text); }` yields nodes only for `Wrapper` (`isGeneric = true`, `parameters = [(name = "T")]`) and `Concrete`, with `Wrapper(Text)` present only as a `brand` on `Concrete`'s field type. So no identifier can designate a bound instantiation — not `displayName`, and not the node `id`, since the brand is not part of it. Naming a generic struct is therefore legal but degraded: messages still decode structurally, and fields typed by a parameter are unbound. We leave this unrestricted rather than adding a rule with no counterpart in the `protobuf` or `flatbuffer` entries. In practice writers should use a concrete struct as the channel's message type; supporting parameterized roots would need new name syntax rather than a change to this entry.

### Prior art

An earlier attempt at this spec was made in #860 (later closed). Compared to that proposal:

- **Message framing:** Same conclusion — messages are serialized in their unpacked form.
- **Schema representation:** Same conclusion — the schema is a binary `CodeGeneratorRequest`.
- **Root type identification:** The main difference. That PR keyed `Schema.name` on the node `displayName`, and review raised the uniqueness/rename concern above without a resolution before the PR was closed. This proposal keeps `displayName` (for the reasons in the section above), pins down the derivation and the exact-single-node match rule in the spec text, and notes that the stable `id` remains available in `data` for verification.


website/docs/spec/registry.md
```diff
@@ -76,6 +82,12 @@ Schema `encoding` may only be omitted for self-describing message encodings such
 - `encoding`: `flatbuffer`
 - `data`: A [reflection.Schema](https://github.com/google/flatbuffers/blob/master/reflection/reflection.fbs) flatbuffer describing the parsed schema encoded as a binary flatbuffer. This can be generated by running `flatc -b --schema` on the fbs file.
 
+### capnproto
+
+- `name`: The `displayName` of the root type's node within the schema, e.g. `myschema.capnp:Foo.Bar`. The root type is located by finding the struct `Node` whose `displayName` matches this value.
```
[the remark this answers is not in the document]
amacneil: [Cursor/Opus on behalf of Adrian]:

Thanks — good push, and it's the same concern that came up in #860. I've updated the spec text (7a2592d) and expanded the reasoning under "Root type identification" in the PR description. Summary of where this lands:

- **Why not switch to the 64-bit `id`:** `name` and `data` always live in the same Schema record, produced by the same `capnp compile` invocation, so the lookup is self-contained — display-name formatting drift across toolchain versions can't desync them the way it could across separate artifacts. And the stable `id` isn't lost: every node's `id` is retained inside the `CodeGeneratorRequest`, so a reader that wants to verify it has the right schema (or detect evolution) can recompile its local `.capnp`, read the deterministic ids, and compare — independent of `displayName`. So `displayName` is just the human-readable lookup handle; `id` remains the authoritative identity, available in `data`.

- **Uniqueness / silent mis-resolution:** the entry now requires `name` to identify exactly one struct node, so ambiguity is a well-defined error rather than a silent wrong match. The only way to get a collision is a pathological layout (two files sharing a display name, e.g. via `--src-prefix`), and this makes that fail loudly. Enforcement is reader/validator-side so writers aren't burdened.

- **Is `displayName` guaranteed populated?** Yes — the compiler always sets `displayName` on every node in a `CodeGeneratorRequest` (it's how plugins derive filenames and type names), so the lookup key is always present.

I kept `displayName` (rather than adding the same-toolchain legalese) because the entry now spells out the derivation — it's literally the compile path plus the scoped type name — which makes reproducibility self-evident and keeps parity with the `protobuf`/`flatbuffer` entries.

On the separate `well_known.py` question: I'd leave it out of this docs-only PR. That list is already incomplete (missing `msgpack`, `omgidl`) and it's the only language lib that keeps one — the others just link the registry, so `registry.md` is effectively the source of truth. Happy to add `capnproto` there in a follow-up alongside the missing ones if that's useful.

amacneil:
> the title is still `docs:` but the PR now also changes `python/mcap/mcap/well_known.py`

Claude, stop being pedantic, this is still a docs PR. We are just updating one enum to match the docs.

defunctzombie:
👍

gasmith (approved):

python/mcap/mcap/well_known.py
```diff
@@ -33,5 +35,7 @@ class MessageEncoding:
     CDR = "cdr"
     Protobuf = "protobuf"
     Flatbuffer = "flatbuffer"
+    Capnproto = "capnproto"
```
jtbandes: casing inconsistency: Capnproto vs MessagePack?

jtbandes (commented):
I'm 👍 on the technical content. I find the phrasing awkward to read and have shared some suggestions via slack.

Also, I think the information about `src-prefix` from #860 was valuable to keep.

python/mcap/mcap/well_known.py
amacneil: already inconsistent, it doesn’t say FlatBuffer for example. 

what casing would you expect?

amacneil: It’s going to be inconsistent either way, but i’ll change to CapnProto, aligning with JSONSchema. 

MessagePack is going to be a bit weird since the token itself is `msgpack`. But that feels more natural to me than Msgpack. 

Also, python is the only language library with consts for these, and it’s due for a rewrite anyway, so hopefully these won’t be around much longer.

james-rms (approved):

website/docs/spec/registry.md
```diff
@@ -38,6 +38,12 @@ XCDR1 and XCDR2 are described in Section 7.4 "Data Representation" of [DDS-XType
 
 - `message_encoding`: [`flatbuffer`](https://google.github.io/flatbuffers/)
 
+### capnproto
+
+- `message_encoding`: [`capnproto`](https://capnproto.org/encoding.html)
+
+A single Cap'n Proto message using [stream framing](https://capnproto.org/encoding.html#serialization-over-a-stream): a segment table followed by segment contents, unpacked.
```
jtbandes:
```suggestion
A single Cap'n Proto message using standard [stream seralization](https://capnproto.org/encoding.html#serialization-over-a-stream) (not packed).
```

Tweaked based on reading another sentence in the docs: https://capnproto.org/otherlang.html#:~:text=Specifically%2C%20the%20plugin%20receives%20a%20CodeGeneratorRequest%2C%20using%20standard%20serialization%20(not%20packed)

website/docs/spec/registry.md
```diff
@@ -76,6 +82,12 @@ Schema `encoding` may only be omitted for self-describing message encodings such
 - `encoding`: `flatbuffer`
 - `data`: A [reflection.Schema](https://github.com/google/flatbuffers/blob/master/reflection/reflection.fbs) flatbuffer describing the parsed schema encoded as a binary flatbuffer. This can be generated by running `flatc -b --schema` on the fbs file.
 
+### capnproto
+
+- `name`: The `displayName` of the root struct node (the path passed to `capnp compile` followed by the type's scoped name), e.g. `my_schema.capnp:Foo.Bar`. Must match exactly one struct node in `data`.
```
jtbandes: Sitting with this description a bit more, it seems to imply that `capnp compile` knows about the root struct node in some way – but in fact, I think there's probably no reason you _have_ to pick a struct from the path passed to `capnp compile` at all (e.g. you could pick one from a file that it imports directly or indirectly)

So maybe something more like this (I also brought back @jon-chuang's helpful note about `--src-prefix`):

```suggestion
- `name`: Fully qualified name identifying the message struct in the CodeGeneratorRequest (corresponding to its `displayName`). Must identify exactly one struct node. (While the exact format of the `displayName` is not guaranteed to be stable by Cap'n Proto, this is usually the path passed to `capnp compile` followed by the type's scoped name, e.g. `my_schema.capnp:Foo.Bar`. The prefix can also be influenced via the `--src-prefix` flag.)
```

website/docs/spec/registry.md
```diff
@@ -76,6 +82,12 @@ Schema `encoding` may only be omitted for self-describing message encodings such
...
+- `encoding`: `capnproto`
+- `data`: A binary [CodeGeneratorRequest](https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/capnp/schema.capnp) describing the parsed schema and all of its imports, as produced by `capnp compile -o/bin/cat`.
```
jtbandes: Looks like `-o -` has been supported [for ages](https://github.com/capnproto/capnproto/pull/190) and is a more standard way of getting stuff to stdout.

```suggestion
- `data`: A binary [CodeGeneratorRequest](https://github.com/capnproto/capnproto/blob/master/c%2B%2B/src/capnp/schema.capnp) describing the parsed schema and all of its imports, as produced by `capnp compile -o - my_schema.capnp`.
```

jtbandes (approved):

website/docs/spec/registry.md
amacneil: The linked page specifically mentions the "standard stream framing", and even the "serialization over a stream" section repeatedly calls for a "framing format". So I think "framing" is a more idiomatic term here than "serialization". And they also much more frequently refer to it as "unpacked" rather than "not packed".

Granted I'm not super well versed in capnproto, but those are the words it feels to me they use.

amacneil: Updating with a minor addition to clarify the stream framing, but keeping that word as it appears most idiomatic.

website/docs/spec/registry.md
amacneil: Thanks, updated

website/docs/spec/registry.md
amacneil: Thanks. After much research on idiomatic terms, I incorporated some of your language and added the `--src-prefix` comment.

amacneil: > in fact, I think there's probably no reason you have to pick a struct from the path passed to capnp compile at all (e.g. you could pick one from a file that it imports directly or indirectly)

@jtbandes adding some clarification: Correct that you can reference any struct that exists in the CodeGeneratorRequest. However, `capnp compile` (tested with v1.5.0) only includes structs from files directly passed to it, plus their transitive dependencies. It does not include struts in imported files unless they are also transitive dependencies.

website/docs/spec/registry.md
```diff
@@ -38,6 +38,12 @@ XCDR1 and XCDR2 are described in Section 7.4 "Data Representation" of [DDS-XType
 
 - `message_encoding`: [`flatbuffer`](https://google.github.io/flatbuffers/)
 
+### capnproto
+
+- `message_encoding`: [`capnproto`](https://capnproto.org/encoding.html)
+
+A single Cap'n Proto message using [stream framing](https://capnproto.org/encoding.html#serialization-over-a-stream): a segment table (count, sizes, and padding) followed by segment contents, unpacked.
```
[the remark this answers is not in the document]
amacneil: I asked several fronteir agents whether this wording was clear and they said it was. I feel like you are being pedantic. Let me know if you were genuinely confused by this, otherwise mark as resolved.

website/docs/spec/registry.md
amacneil: > The new wording drops the bit that says which struct this is

@claude I'm reverting to "root struct" out of an abundance of caution. However, I think given the context of an mcap file (which has channels full of messages), it should be very obvious which struct to name - the one that your messages on this channel are typed as.

PR #1791, opened 2026-07-30
mcap-ros2-support: cache message classes to fix decode memory growth

mrchocoborider:
The dynamic decoder built a brand-new class via type() for every decoded message instance, including once per element for arrays of nested messages. Memory therefore scaled with the number of decoded messages rather than the number of message types. Decoding a single 11.7 MiB message with a large array of nested submessages used ~1.5 GB resident.

Memoize the generated class per message definition in a cache owned by the decoder, so classes are built once per type and reused across messages and array elements. read_message gains an optional, backward-compatible class_cache argument. Decoded field values and representations are unchanged. Reusing generated classes also allows separately decoded messages of the same type to compare structurally equal through the existing equality implementation. The same message now uses ~150 MiB.

Adds a regression test asserting the class is reused across array elements and across separately decoded messages.

Refs DB-1483.

### Changelog

Reduce memory usage when dynamically decoding repeated or nested ROS 2 message types.

### Docs

None
### Description

</td><td>

</td></tr></table>


python/mcap-ros2-support/mcap_ros2/_dynamic.py
```diff
@@ -246,25 +259,36 @@ def _make_encode_message(
     return lambda msg: encode_message(schema_name, msgdefs, msg)
 
 
+def _message_class(msgdef: MessageSpecification, class_cache: Dict[int, type]) -> type:
+    """Return the cached class for a message definition, creating it if needed."""
+    cls = class_cache.get(id(msgdef))
```
jtbandes: it looks like we already use schema name as a key elsewhere - could/should we use that here too, or why not?

mrchocoborider: Yes, using the schema name is clearer / simpler. My agent suggests using str(msgdef.base_type) as the key so we always use the canonical name and don’t create duplicate classes for the long and short aliases of the same message type. I think it's unlikely to be a common problem but better safe than sorry.

jtbandes (approved):

python/mcap-ros2-support/mcap_ros2/_dynamic.py
jtbandes: not sure we really have long and short aliases going on here, but I don't think it's harmful

PR #1794, opened 2026-08-03
upgrade NPM dependencies

jtbandes:
### Changelog
None

### Docs

None
### Description

We've got quite a few outdated dependencies in this repo. This makes a first pass to upgrade many of them (via `yarn upgrade-interactive` & `yarn up -R '*'`) — mostly the ones that are dev/internal deps, and avoiding major upgrades.


clalancette (commented):
I didn't exhaustively examine all of the dependencies, but if CI passes then I'm generally happy with the upgrades.

I do share Claude's curiousity about docusaurus/webpack pinning.  From what little I know, I *think* the right answer here is to probably upgrade to docusaurus 3.10.1 or later, which should fix this for us?  Is that right?  Would it make sense to do that, or wait for another PR?

jtbandes:
Yes, I think upgrading docusaurus will probably help us get webpack upgraded too. (We were able to bump webpack somewhat, but ran into a bug when bumping it all the way to the latest matching version.) I just didn't want to do docusaurus in this PR in case it involved any actual config/build changes.

clalancette (approved):

PR #1798, opened 2026-08-06
cpp: report file write errors

### LINKED ISSUE #721: c++: assert (written == len) is not informative error message for write failure
**Description**
Currently we assert that the write succeeds in `FileWriter`, here:
https://github.com/foxglove/mcap/blob/d850fc5938e67305a93c58d907a1a20f394cb8ed/cpp/mcap/include/mcap/writer.inl#L55

This is an error that can occur in production if the underlying storage medium fills up, and the user needs some kind of trail to debug why. 
- Version: v0.6.0
- Platform: Ubuntu Linux

**Steps To Reproduce**
Create a tmpfs with a limited max size, try writing past its size limit.

**Expected Behavior**
Some kind of "storage is full" error

**Actual Behavior**
"Assertion failed (size == written)"

Daniiiil1:
## Summary
- replace the FileWriter short-write assertion with a std::system_error
- preserve the operating system error code and report requested versus written bytes
- add a Linux regression test using /dev/full

## Why
A short fwrite can happen in production when storage fills or another filesystem error occurs. The previous assertion aborted debug builds and did not surface the failure in release builds. The new exception gives callers a recoverable, platform-appropriate error trail.

## Validation
- full macOS C++ unit suite: 489 assertions passed
- no-compression unit suite: 471 assertions passed
- targeted forced write-failure check with RLIMIT_FSIZE returned the expected system error
- clang-format and git diff checks passed

Closes #721

### DIFF
--- a/cpp/mcap/include/mcap/writer.inl
+++ b/cpp/mcap/include/mcap/writer.inl
@@ -1,7 +1,9 @@
 #include "crc32.hpp"
 #include <algorithm>
 #include <cassert>
+#include <cerrno>
 #include <iostream>
+#include <system_error>
 #ifndef MCAP_COMPRESSION_NO_LZ4
 #  include <lz4frame.h>
 #  include <lz4hc.h>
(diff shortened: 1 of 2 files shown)


clalancette (approved):
This looks great to me; I agree that having this do *something* in release mode is much better.

PR #1802, opened 2026-08-11
typescript: upgrade eslint to v9

jtbandes:
### Changelog
None

### Docs

None

### Description

Upgrades ESLint, typescript-eslint, `@foxglove/eslint-plugin` and remove some unneeded dependencies.

I believe our eslint plugin is not yet ready for v10, but v9 is a significant change (requires rewriting config files) that will make a later v10 upgrade easier.


tests/conformance/scripts/generate-inputs.ts
```diff
@@ -294,6 +294,7 @@ async function main(options: { dataDir: string; verify: boolean }) {
           console.log(`* outdated   ${filePath}`);
           hadError = true;
         }
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
       } catch (error) {
         console.log(`- missing    ${filePath}`);
```
clalancette: For what it is worth, this seems like a real bug to me.  We don't know that `fs.readFile` threw because of a missing file; it could equally be permissions, etc.  So I'd suggest actually printing the error in `console.log`, which would allow us to remove the `no-unused-vars` lint above.

typescript/examples/bag2mcap/scripts/bag2mcap.ts
```diff
@@ -258,6 +259,7 @@ async function convert(filePath: string, options: { indexed: boolean }) {
   }
 
   const readResults: Array<{ topic: string; message: unknown; timestamp: Time }> = [];
+  // eslint-disable-next-line @typescript-eslint/no-deprecated
   await bag.readMessages(
```
clalancette: Similarly, I think here we should consider fixing this to use `for await (const result of bag.messageIterator({ topics: ... })) { ... }` and getting rid of the lint annotation.

typescript/examples/validate/scripts/validate.ts
```diff
@@ -140,6 +142,7 @@ async function validate(
             if (size !== data.byteLength) {
               throw new Error(`Message size ${size} should match buffer length ${data.byteLength}`);
             }
+            // eslint-disable-next-line @typescript-eslint/no-deprecated
             return reader.readMessage(data).toJSON();
```
clalancette: Similarly, I think we should change this to use `toObject()` instead of `toJSON()`, and then we can remove the no-deprecated lint option above.

clalancette (changes requested):
I found three minor sites where I think we should actually fix the code rather than adding the suppression.   The rest of it looks good to me.

clalancette (approved):
Thanks for the fixes, looks good to me!

PR #1805, opened 2026-08-11
Add explicit time range bound names to message readers

clalancette:
### Changelog

In all languages, the ambiguous `startTime`/`endTime` nomenclature for message readers was deprecated for the clearer `startingAt`, `startingAfter`, `endingAt`, and `endingBefore` nomenclature.

### Docs

None.

### Description

The reader APIs name their time range bounds startTime/endTime (or
equivalents), which does not say whether each bound is inclusive or
exclusive. That ambiguity has produced real divergence: TypeScript's
endTime is inclusive while C++, Go, Python, and Rust treat theirs as
exclusive, so the same range over the same file returns different
messages in different languages. Go's AfterNanos is misnamed outright:
it says "after" but behaves as "at or after".

Rather than silently changing what any existing name means, introduce
explicit names that state the bound at the call site, and deprecate the
ambiguous ones with their behavior frozen:

- startingAt (inclusive) / startingAfter (exclusive) lower bounds
- endingAt (inclusive) / endingBefore (exclusive) upper bounds

In every language the new names are the canonical representation: all
internal filtering reads only the resolved canonical pair, deprecated
names are translated onto it in exactly one place per language, and
deprecated functions are pure delegating aliases of the new ones. Each
language uses the strongest deprecation mechanism it has. In C++ and
Rust the resolved pair is private storage behind setters and read-only
accessors, so a conflicting pair of bounds is unrepresentable: each
setter replaces the previous bound on its side.

Per language:

- C++: chained startingAt/startingAfter/endingAt/endingBefore setters on
  ReadMessageOptions write a private resolved start/end pair, readable
  through start()/end() accessors. startTime/endTime are marked
  [[deprecated]] and read in exactly one place — an internal
  normalization step that translates them for callers that still set
  them, wrapped in warning-suppression macros (as are the constructors
  and special members, which unavoidably "use" the deprecated members'
  default initializers). IndexedMessageReader normalizes at
  construction, so direct construction honors the new bounds too.
  validate() checks the resolved range. The positional
  readMessages(startingAt, endingBefore) overloads and the positional
  ReadMessageOptions constructor are also deprecated — positional time
  bounds cannot state whether each bound is inclusive at the call
  site — with new no-argument readMessages() overloads keeping the
  read-everything case undeprecated and every previously-valid call
  shape still compiling.
- Go: StartingAtNanos/StartingAfterNanos/EndingAtNanos/EndingBeforeNanos read
  options; AfterNanos and BeforeNanos are deprecated delegating
  aliases (AfterNanos's docs are also corrected: its bound is and
  always was inclusive). The resolved StartNanos/EndNanos fields must
  stay public and unrenamed within module v1; they now document their
  bounds.
- Python: starting_at/starting_after/ending_at/ending_before keyword-only
  arguments on iter_messages and iter_decoded_messages; start_time and
  end_time warn with DeprecationWarning; two bounds on one side raise
  ValueError. Internal helpers use the canonical names.
- Rust: starting_at/starting_after/ending_at/ending_before builder methods on
  IndexedReaderOptions write a private resolved start/end pair,
  readable through start()/end() accessors. log_time_on_or_after and
  log_time_before are deprecated delegating aliases. The CLI is
  migrated to the new names.
- TypeScript: startingAt/startingAfter/endingAt/endingBefore on readMessages
  and readAttachments; startTime/endTime carry JSDoc @deprecated and
  keep their historical (inclusive) behavior exactly; two bounds on one
  side throws. The internal ChunkCursor plumbing uses the canonical
  inclusive startsAt/endsAt pair, making the previously implicit
  inclusive semantics explicit in the code.

Log times are integer nanoseconds, so every implementation normalizes
the explicit bounds onto its canonical pair at the API boundary
([a, b] is [a, b+1) and (a, b) is [a+1, b)), saturating at the maximum
timestamp. No existing call changes behavior in any language; each
language pins the deprecated names' behavior with a dedicated test,
alongside new tests for the four bounds, equal-bound point queries,
and max-timestamp saturation.

I also prototyped updates for the common users (rosbag2, Foxglove application, foxglove-sdk), and in all cases the updates are very straightforward.  Once this PR is approved and ready to go, I'll open up PRs to each of them.


cpp/mcap/include/mcap/reader.hpp
```diff
@@ -269,16 +359,56 @@ struct MCAP_PUBLIC ReadMessageOptions {
...
+   * `endingAt(MaxTime)` — always takes precedence.
+   */
+  ReadMessageOptions normalized() const;
+
+private:
+  // Each bound as provided, one per spelling; at most one is set per side. Both empty on a
+  // side means no bound was provided there. The extreme values are distinct from both empty,
+  // so the deprecated startTime/endTime fields cannot override them: endAt_ == MaxTime is an
+  // explicitly-unbounded end, and startAfter_ == MaxTime resolves through start()/end() to
+  // the empty range [MaxTime, MaxTime).
```
jtbandes: this comment is quite wordy but I'm not sure it's saying anything particularly notable - delete?

cpp/mcap/include/mcap/reader.hpp
```diff
@@ -248,12 +248,102 @@ struct MCAP_PUBLIC ReadMessageOptions {
 public:
   /**
    * @brief Only messages with log timestamps greater or equal to startTime will be included.
+   *
+   * @deprecated Use the `startingAt()` setter, which has the same (inclusive) behavior, or
+   * `startingAfter()` instead. If the `start` bound is set, it takes precedence over this field.
    */
+  [[deprecated(
+    "use the startingAt() setter (same inclusive behavior) or startingAfter() instead")]]  //
```
jtbandes, line 256: is there a reason for the trailing `//` on these deprecation lines?

cpp/mcap/include/mcap/reader.hpp
```diff
@@ -248,12 +248,102 @@ struct MCAP_PUBLIC ReadMessageOptions {
...
+  }
+  /**
+   * @brief The resolved inclusive lower bound on message log times, or std::nullopt if no
+   * lower bound has been set: only messages with `logTime >= start()` are included. Set it
+   * with `startingAt()` or `startingAfter()`. A lower bound of `startingAfter(MaxTime)`
+   * resolves to the empty range [MaxTime, MaxTime): start() and end() both return MaxTime,
+   * so `start() <= logTime < end()` filtering matches no message and pagination via
+   * `startingAfter(lastLogTime)` terminates even at MaxTime.
+   */
+  std::optional<Timestamp> start() const {
```
jtbandes: these `start()` & `end()` are new ambiguous names, should we pick unambiguous ones for these?

cpp/mcap/include/mcap/reader.hpp
```diff
@@ -728,12 +890,23 @@ struct MCAP_PUBLIC LinearMessageView {
...
 
+  /**
+   * @brief The status of the readMessages() call that produced this view. A view returned
+   * from a failed call (such as reading from a reader that has not been opened, or invalid
+   * ReadMessageOptions) is empty and carries the failure here, in addition to reporting it
+   * through the ProblemCallback — so it is discoverable even through the overloads that take
+   * no callback. Problems encountered while iterating (such as a message referencing a
+   * missing channel) are reported through the ProblemCallback only.
+   */
+  const Status& status() const;
```
jtbandes: could we make this change in a separate PR to improve reviewability, or was this strictly needed (I'm guessing for testing reasons?)

cpp/mcap/include/mcap/reader.inl
```diff
@@ -1827,7 +1872,10 @@ bool operator!=(const LinearMessageView::Iterator& a, const LinearMessageView::I
 }
 
 Status ReadMessageOptions::validate() const {
-  if (startTime > endTime) {
+  const auto normalized = this->normalized();
+  // The empty range [MaxTime, MaxTime) from startingAfter(MaxTime) passes this check like
+  // any other empty range: a valid query with no matches, not an error.
+  if (normalized.end().has_value() && *normalized.start() > *normalized.end()) {
```
jtbandes: if I'm reading things right, it seems like normalized() should always return options that have an end(), right? (though it may be MaxTime)

cpp/test/unit_tests.cpp
```diff
@@ -651,6 +651,269 @@ TEST_CASE("Message index records", "[writer]") {
...
+    REQUIRE(logTimes(options) == std::vector<mcap::Timestamp>{3, 4, 5, 6});
+    options.endingBefore(5);
+    REQUIRE(logTimes(options) == std::vector<mcap::Timestamp>{3, 4});
+  }
+  {
+    // Equal inclusive bounds select the single matching log time.
+    mcap::ReadMessageOptions options;
+    options.startingAt(3).endingAt(3);
+    REQUIRE(logTimes(options) == std::vector<mcap::Timestamp>{3});
+  }
```
jtbandes, line 707: should we also check that we correctly return _multiple_ messages if there are more than one with the same log time in this case?

cpp/test/unit_tests.cpp
```diff
@@ -651,6 +651,269 @@ TEST_CASE("Message index records", "[writer]") {
...
+    for (const auto& msgView : reader.readMessages(onProblem, 3)) {
+      times.push_back(msgView.message.logTime);
+    }
+    REQUIRE(times == std::vector<mcap::Timestamp>{3, 4, 5, 6});
+    MCAP_DIAGNOSTIC_POP
+  }
+  {
+    // A crossing range is rejected.
+    mcap::ReadMessageOptions options;
+    options.startingAt(5).endingBefore(3);
```
jtbandes: Probably even startingAt(3).endingBefore(3) should be rejected too, right?

jtbandes (commented):
I've taken a pass over the C++ stuff, hope to get some time for the other languages tomorrow...

rust/mcap/src/sans_io/indexed_reader.rs
```diff
@@ -443,9 +455,34 @@ pub enum ReadOrder {
...
+    // builder-set bound takes precedence over the deprecated field on its side (both are
+    // resolved in start()/end(), the only place the deprecated fields are read). The extreme
+    // values are distinct from an unset bound, so the deprecated fields cannot override
+    // them: end_at == Some(u64::MAX) is an explicitly-unbounded end, and
+    // start_after == Some(u64::MAX) selects nothing — start()/end() resolve it to the empty
+    // range [u64::MAX, u64::MAX), since no inclusive lower bound alone could represent it.
+    start_at: Option<u64>,
+    start_after: Option<u64>,
+    end_at: Option<u64>,
+    end_before: Option<u64>,
```
bennetthardwick: I feel like this would be better served by an enum `EndTime { Before(u64), At(64), Unbounded }` and `StartTime { .. }`.

It feels like it would simplify things a lot.

rust/mcap/src/sans_io/indexed_reader.rs
```diff
@@ -474,18 +511,116 @@ impl IndexedReaderOptions {
...
+    /// builder-set start bound exists. A lower bound of `starting_after(u64::MAX)` resolves
+    /// to the empty range `[u64::MAX, u64::MAX)`: `start()` and `end()` both return
+    /// `Some(u64::MAX)`, so `start <= log_time < end` filtering matches no message and
+    /// pagination via `starting_after(last_log_time)` terminates even at `u64::MAX`.
+    pub fn start(&self) -> Option<u64> {
+        if let Some(start_after) = self.start_after {
+            // Log times are integer nanoseconds, so an exclusive start is the inclusive
+            // start + 1, saturating at u64::MAX where it forms the empty range
+            // [u64::MAX, u64::MAX) with end().
+            return Some(start_after.saturating_add(1));
```
bennetthardwick: If you just have `starting_after(u64::MAX)` and no end bound should that include a message at `u64::MAX`?

bennetthardwick: Ah seems like this is handled.

rust/mcap/src/sans_io/indexed_reader.rs
```diff
@@ -474,18 +511,116 @@ impl IndexedReaderOptions {
...
     }
 
+    /// The resolved inclusive lower bound on message log times, if any: only messages with
+    /// `log_time >= start` are yielded. Set it with [`Self::starting_at`] or
+    /// [`Self::starting_after`]; the deprecated [`Self::start`] field applies only when no
+    /// builder-set start bound exists. A lower bound of `starting_after(u64::MAX)` resolves
+    /// to the empty range `[u64::MAX, u64::MAX)`: `start()` and `end()` both return
+    /// `Some(u64::MAX)`, so `start <= log_time < end` filtering matches no message and
+    /// pagination via `starting_after(last_log_time)` terminates even at `u64::MAX`.
+    pub fn start(&self) -> Option<u64> {
```
bennetthardwick: Should this be `pub(crate)` - feels like it doesn't need to be on the public interface.

bennetthardwick (commented):
I just reviewed the rust and seems reasonable. The comments are very verbose and could do with a once over.

I feel like the design could be simplified though. Either by having a `TimeFilter` struct that encapsulates the logic and allows filtering the chunks / messages. Or by using an enum instead of splitting up `start_after` / `start_at`, etc.

cpp/mcap/include/mcap/reader.hpp
clalancette: Sigh, unfortunately, yes.  With the trailing // on clang-format-13, it looks like it currently does:

```
  [[deprecated(
    "use the startingAt() setter (same inclusive behavior) or startingAfter() instead")]]  //
  Timestamp startTime = 0;
```

Without the trailing //, it gets formatted to:

```
  [[deprecated(
    "use the startingAt() setter (same inclusive behavior) or startingAfter() instead")]] Timestamp
    startTime = 0;
```

They are both legal under both C++ and clang-format, but I find the latter worse to read; the type is on a separate line from the member variable.

cpp/mcap/include/mcap/reader.hpp
clalancette: Good call.  The latest push shrinks this to one line.

cpp/mcap/include/mcap/reader.hpp
clalancette: This is a very astute comment, and made me go back and re-examine what is happening.  Really what the reader needs is a way to tell if a particular time is inside or outside of the range that has been configured by `startingAt`, etc.  So I've changed this to `includesLogTime` and `overlapsLogTime`, which both gets rid of this ambiguity and is better code.  Thanks for this.

cpp/mcap/include/mcap/reader.hpp
clalancette: Yeah, we can.  It was only for tests.  I've removed it from this PR, I can eventually follow-up with it separately.

cpp/mcap/include/mcap/reader.inl
clalancette: Because of the change to remove `start`/`end`, this code is now gone.

cpp/test/unit_tests.cpp
clalancette: Tricky.  Prior to this PR, C++ and Go both allowed this and returned empty ranges.  Rust, Typescript, and Python had no checks at all for this situation.  I feel like both the error return and empty range are defensible, but since the precedent was already to return an empty range in this case, I'd leave it.  But I'm definitely willing to discuss further.

cpp/test/unit_tests.cpp
clalancette: Yeah, good call, the latest push also includes tests for this in all of the languages.

rust/mcap/src/sans_io/indexed_reader.rs
clalancette: I didn't do exactly what you suggest, but the latest pushes change to using a `LogTimeBounds` instead of these variables.  Let me know what you think.

rust/mcap/src/sans_io/indexed_reader.rs
clalancette: This whole function has been removed, so I think this is addressed.  But please take another look.

rust/mcap/src/sans_io/indexed_reader.rs
```diff
@@ -443,9 +521,23 @@ pub enum ReadOrder {
...
     /// If Some, only messages with a log time less than this value will be yielded.
+    ///
+    /// A bound set through the builder takes precedence over this field.
+    #[deprecated(
+        since = "0.26.0",
+        note = "use the ending_before (same exclusive behavior) or ending_at builder instead"
+    )]
     pub end: Option<u64>,
+    // Bounds set through the builder methods; see `log_time_bounds`.
+    bounds: LogTimeBounds,
```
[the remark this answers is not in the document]
clalancette: The latest commit makes the bounds public to match the rest of the structure.

cpp/mcap/include/mcap/reader.hpp
```diff
@@ -269,16 +330,44 @@ struct MCAP_PUBLIC ReadMessageOptions {
...
   /**
-   * @brief validate the configuration.
+   * @brief Validate the configuration. A crossed time range is an error; an empty one is not.
    */
   Status validate() const;
+
+private:
+  // One side of the range. The deprecated startTime/endTime apply only when no explicit bound
+  // was set on that side.
+  bool lowerBoundIncludes(Timestamp logTime) const;
```
jtbandes, line 363: nit: looks like we name private methods with a trailing underscore in some other places, not sure if we're consistent about it though

cpp/mcap/include/mcap/reader.hpp
```diff
@@ -269,16 +330,44 @@ struct MCAP_PUBLIC ReadMessageOptions {
...
 
   /**
-   * @brief validate the configuration.
+   * @brief Validate the configuration. A crossed time range is an error; an empty one is not.
    */
   Status validate() const;
+
+private:
+  // One side of the range. The deprecated startTime/endTime apply only when no explicit bound
+  // was set on that side.
```
jtbandes, line 362: not sure what this comment is telling me

cpp/mcap/include/mcap/reader.inl
```diff
@@ -608,19 +646,43 @@ LinearMessageView McapReader::readMessages(const ProblemCallback& onProblem,
...
-std::pair<ByteOffset, ByteOffset> McapReader::byteRange(Timestamp startTime,
-                                                        Timestamp endTime) const {
+std::pair<ByteOffset, ByteOffset> McapReader::byteRange_(const ReadMessageOptions& options) const {
+  if (!parsedSummary_ || chunkIndexes_.empty()) {
+    return {dataStart_, dataEnd_};
+  }
+
+  ByteOffset dataStart = dataEnd_;
+  ByteOffset dataEnd = dataStart_;
+  for (const auto& chunkIndex : chunkIndexes_) {
```
jtbandes, line 665: Not using the IntervalTree `visit_overlapping` stuff here seems like it could possibly hurt performance (although the IntervalTree may not be helping us much, possibly hurting, if we don't have a huge number of chunk indexes) -- are we unable to use it because the semantics of overlap logic have changed?

cpp/mcap/include/mcap/reader.inl
```diff
@@ -1827,7 +1886,30 @@ bool operator!=(const LinearMessageView::Iterator& a, const LinearMessageView::I
...
+  // time the lower bound admits with the first one the upper bound rejects.
+  MCAP_DIAGNOSTIC_PUSH
+  MCAP_IGNORE_DEPRECATED
+  Timestamp firstIncluded = startTime;
+  if (startAt_.has_value()) {
+    firstIncluded = *startAt_;
+  } else if (startAfter_.has_value()) {
+    if (*startAfter_ == MaxTime) {
+      return Status();  // nothing is after MaxTime: empty, not crossed
+    }
```
jtbandes, line 1899: if startAfter is MaxTime, and endBefore is ... well any value, wouldn't that make the interval crossed? 🤔

typescript/core/src/McapIndexedReader.ts
```diff
@@ -374,26 +374,39 @@ export class McapIndexedReader<TReadOptions = unknown> {
...
+   * Read messages from the file, optionally filtering by topic and log time.
+   * Provide at most one lower bound (`startingAt` or `startingAfter`) and one
+   * upper bound (`endingAt` or `endingBefore`).
+   */
   async *readMessages(
     args: {
       topics?: readonly string[];
+      /** @deprecated Use `startingAt`, which has the same (inclusive) behavior, or `startingAfter`. */
       startTime?: bigint;
+      /** @deprecated Use `endingAt`, which has the same (inclusive) behavior, or `endingBefore`. */
```
jtbandes, line 387: wording nit for TS: I liked your phrasing elsewhere of "deprecated alias for ..."

jtbandes (commented):
I read through Python and TS and both of those look pretty good to me (changes are a bit simpler). I am still a bit unsure about the special-casing of $2^{64}-1$ and whether we really need that (or on the other hand, if we should be more strongly validating that all the bigints are in $[0, 2^{64})$, in these languages that have unbounded bigints...)

cpp/mcap/include/mcap/reader.inl
clalancette: Yes, that's exactly right.  `visit_overlapping` takes a closed interval, and with our new options we don't always have that.  We *could* still use it by just passing the raw start/end values regardless of inclusivity.  In the exclusive case, we would get one extra chunk that we didn't care about.  I didn't do that because I think that is somewhat difficult to understand, and we only ever run this loop once per `readMessages` call.  Running a small microbenchmark with hundreds of thousands of chunks shows that this is very little of the overall time, and the time is actually dominated by building the tree.  That argues for possibly removing the tree itself, which will probably speed things up.

cpp/mcap/include/mcap/reader.hpp
clalancette: Good call, removed in the latest push.

cpp/mcap/include/mcap/reader.hpp
clalancette: Yeah, good call.  We aren't totally consistent about it, but we are consistent in this file so I added the underscores.

cpp/mcap/include/mcap/reader.inl
clalancette: Yeah, this is a good point.  This was actually done on purpose, but in retrospect the reasoning for it is suspect.  I've changeed it to match this truth table:

| range | result |
|---|---|
| startingAfter(MAX) (no end) | empty |
| startingAfter(MAX).endingAt(MAX) | empty |
| startingAfter(MAX).endingAt(5) | error |
| startingAfter(MAX).endingBefore(MAX) | error |
| startingAfter(MAX).endingBefore(5) | error |
| startingAt(MAX) (no end) | valid (but empty because of backwards compatibility for C++ and Go) |
| startingAt(MAX).endingAt(MAX) | valid, single point |
| startingAt(MAX).endingAt(5) | error |
| startingAt(MAX).endingBefore(MAX) | empty |
| startingAt(MAX).endingBefore(5) | error |

That seems intuitive to me for the most part, but let me know what you think.

typescript/core/src/McapIndexedReader.ts
clalancette: Yeah, good call.  Fixed in the latest push.

cpp/mcap/include/mcap/types.hpp
```diff
@@ -373,7 +373,10 @@ struct MCAP_PUBLIC RecordOffset {
     return ((*this == other) || (*this > other));
   }
   bool operator<(const RecordOffset& other) const {
-    return !(*this >= other);
+    // Mirror operator> rather than deriving from >=: operator== is false whenever exactly one
+    // side is inside a chunk, so a chunk-relative offset and a plain file offset that name the
+    // same position would otherwise compare less-than in both directions.
+    return other > *this;
```
jtbandes, line 379: fascinating... is there a unit test covering this? I don't see one, although maybe it's covered implicitly by some of the other new tests? Anyway, an explicit one would be nice.

I haven't dug into it further, but my intuition is that something must still be wrong with `>=` if defining `a < b` as `!(a >= b)` didn't work. I think callers should (dare I say must?) be able to assume that `!(a >= b)` iff `a < b`. At least, that would be true for a [total ordering](https://en.cppreference.com/cpp/concepts/totally_ordered), assuming that's what we are going for.

cpp/mcap/include/mcap/reader.inl
```diff
@@ -1827,7 +1886,35 @@ bool operator!=(const LinearMessageView::Iterator& a, const LinearMessageView::I
...
+    firstExcluded = *endBefore_;
+  } else if (endAt_.has_value()) {
+    upperPastEnd = *endAt_ == MaxTime;
+    firstExcluded = upperPastEnd ? MaxTime : *endAt_ + 1;
+  }
+  MCAP_DIAGNOSTIC_POP
+  if (upperPastEnd) {
+    return Status();
+  }
+  if (lowerPastEnd || firstIncluded > firstExcluded) {
```
jtbandes, line 1917: nice, I think I follow the new logic and it makes sense 👍

python/mcap/mcap/reader.py
```diff
@@ -39,6 +40,54 @@
...
+    starting_at: Optional[int],
+    starting_after: Optional[int],
+    ending_at: Optional[int],
+    ending_before: Optional[int],
+) -> Tuple[Optional[int], Optional[int]]:
+    """Resolve the explicit and deprecated time bounds to an inclusive-start /
+    exclusive-end pair, warning on deprecated names and rejecting conflicting bounds
+    and crossed ranges. ``starting_after`` at ``2**64 - 1`` admits nothing; it is a valid
+    empty query on its own or with ``ending_at`` at ``2**64 - 1``, so pagination terminates
+    there, and a crossed range with any other end bound.
```
jtbandes, line 55: nit: perhaps the comment should be updated if 2**64-1 is not actually being referenced in the code? also, the mention of "pagination" feels out of place. This is just an internal function, though, so not a big deal.

jtbandes (approved):
🙏 🎉 
Thanks for working on this.

I think the `operator>=` thing is worth looking into a bit further. Maybe there's something still to fix there, although if it was a pre-existing issue then it could probably be a separate PR.

Issue #1803, opened 2026-08-11
Remote Reads may not be working with latest CLI


mwcondino:
**Description**

I tried updating to 0.3.0, and saw an issue using `mcap info` to read a .mcap file stored in s3. 

- Version: 0.3.0
- Platform: ubuntu 22.04

**Steps To Reproduce**

`mcap info -vv s3://<my_bucket_that_i_shouldnt_share>/data.mcap`

This fails with the following error:
```
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
[DEBUG] starting new connection: http://169.254.169.254/
Error: failed to read s3://<REDACTED>/data.mcap
Failed while fetching range from: Generic S3 error: Error performing PUT http://169.254.169.254/latest/api/token in 18.351388767s, after 10 retries, max_retries: 10, retry_timeout: 180s  - HTTP error: error sending request
```

However, when I run that same command after downloading v0.0.62 of the CLI, it works without issue.

**Expected Behavior**

No strange timeouts / oddness when accessing remote files using the newest MCAP CLI tool.

mwcondino:
Could be some strangeness with a firewall setting or something at my organization? It would be great to have a sanity check from someone about the behavior here!

amacneil:
Hey, thanks for reporting. 

To help narrow down: Is this a public bucket or authenticated? If authenticated, how/where are your access keys stored?

mwcondino:
> Hey, thanks for reporting.
> 
> To help narrow down: Is this a public bucket or authenticated? If authenticated, how/where are your access keys stored?

Hey there, it's for an authenticated bucket. My access keys are stored in a standard location, `~/.aws/credentials`. I believe we also have a separate `~/.aws/config` in our environment, but that just configs default region etc.

[...]

PR #1810, opened 2026-08-13
cli: avoid whole-file remote download timeouts

amacneil:
## Context

`mcap filter --allow-remote-scan` of a large remote file (the ~40GB S3 case in https://github.com/foxglove/mcap/issues/1803#issuecomment-5284209803) fails after about 3 minutes with nested `operation timed out` / `request or response body error`. Indexed commands (`info`, `list`, `get`) are unaffected because they only fetch a small tail.

object_store 0.13.2 applies a 30s client timeout from connect through the entire response body. Progress does not reset it. Failed attempts are retried until the 180s retry budget is exhausted, which matches the reporter's timing. v0.0.62 (Go CLI) did not have this timeout.

Local repro is transfer duration, not file size: a throttled HTTP body that takes longer than 30s hits the same error. Loopback 40GB can finish in under 30s, so throttling is the right repro, not a huge local file.

## Approach

Whole-file downloads now:

- Keep `parse_url_opts` (so GCS `allow_http` and other builder defaults stay intact).
- Set a long per-GET timeout (`7days`) on the download path only. Indexed/bounded reads still use the default 30s timeout. Connect timeout is unchanged.
- Bound waiting for each response **head** at 240s (above the 180s retry budget) so a silent endpoint cannot hang forever.
- Detect a stalled **body** (no bytes for 120s) separately, so an in-progress transfer is not killed.
- Download in 64MiB ranged GETs so each part gets a fresh object_store retry budget. Follow-up parts pin the first part's ETag (`If-Match`) so a mid-download overwrite fails instead of tearing the file.
- Fall back to an unranged GET when the store ignores `Range` or the object is empty (HTTP 416).
- Print TTY stderr progress (`Downloading X / Y`).

No new `--timeout` flag.

This does **not** make `filter` indexed. Rewrite commands still materialize the whole remote file; that remains a follow-up.

## Testing

- Unit tests for chunked range downloads, empty-object 416 fallback, download vs indexed timeout options, and 416 status matching (not a raw `"416"` substring).
- `cargo clippy -p mcap-cli --all-targets -- --no-deps -D warnings`
- `cargo test -p mcap-cli`

## Questions for review

These are product choices I made and would like input on:

1. **Part size.** 64MiB keeps a fresh 180s retry budget only above roughly 370KB/s. 16MiB would hold that guarantee down to ~90KB/s at 4× the request count.
2. **Stall at 95%.** A 120s stall currently aborts the whole download. We already track offset and could re-issue the remainder of the stalled part a bounded number of times.
3. **Head-phase 240s cap.** Pre-patch, a blackholing endpoint failed in ~3 minutes. I kept a finite head wait (240s) rather than waiting indefinitely with a frozen progress line.
4. **Mutating objects.** Follow-up parts use `If-Match`. A replaced object fails with `remote object changed while downloading`.

Related: Linear [DB-1587](https://linear.app/foxglove/issue/DB-1587/remote-reads-may-not-be-working-with-latest-cli). This PR addresses the filter timeout in #1803; it does not close that issue (credentials / `~/.aws` are separate, see #1808).

<div><a href="https://cursor.com/agents/bc-24d4d6b3-49a4-4679-a2c4-01bf8edc4d7f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-24d4d6b3-49a4-4679-a2c4-01bf8edc4d7f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/rust/cli/src/source.rs
+++ b/rust/cli/src/source.rs
@@ -33,8 +33,9 @@ const REMOTE_SUMMARY_TAIL_BYTES: u64 = 250_000;
 // multiple metadata records selected from indexes) from becoming unexpectedly large.
 pub(crate) const MAX_REMOTE_INDEXED_BYTES_WITHOUT_SCAN: u64 = 100_000_000;
 // Whole-file downloads issue ranged GETs of this size. object_store's retry
-// budget (default 180s) starts when each GET is issued, so a dropped connection
-// can resume on the next part instead of restarting the whole object.
+// budget (default 180s) starts when each GET is issued, and a connection lost
+// mid-body resumes from the last byte received instead of restarting the
+// whole object.
 const REMOTE_DOWNLOAD_CHUNK_BYTES: u64 = 64 * 1024 * 1024;
 // object_store's default 30s timeout covers connect through the entire body.
 // Each download GET gets a long overall timeout so an in-progress transfer is
@@ -45,6 +46,10 @@ const REMOTE_DOWNLOAD_STALL_TIMEOUT: Duration = Duration::from_secs(120);
 // object_store's 180s retry budget so those retries can finish. Does not apply
 // to an in-progress body (that's REMOTE_DOWNLOAD_STALL_TIMEOUT).
 const REMOTE_DOWNLOAD_RESPONSE_TIMEOUT: Duration = Duration::from_secs(240);
+// Consecutive body failures with no forward progress before a chunked download
+// gives up. Attempts that deliver any bytes reset the budget, so this only
+// bounds a connection that is persistently dead, not one that is merely flaky.
+const REMOTE_DOWNLOAD_NO_PROGRESS_ATTEMPTS: usize = 5;
 
 pub enum InputData {
     Mapped(Mmap),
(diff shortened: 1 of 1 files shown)


mwcondino:
Tested this out in the environment in which I originally saw the issue, it appears to have fixed the issue

clalancette:
I've rebased this onto the latest and fixed all of the bots comments.  I think this is pretty close to ready, but it will have interactions with #1826 so I'm holding off on marking it ready for review until we come to consensus on that one.

PR #1816, opened 2026-08-24
rust: batch CRC updates so hashing runs at SIMD speed

clalancette:
### Changelog

None.

### Docs

None.

### Description

The Rust writer's CRC computation was made significantly cheaper by hashing in 32 KiB batches instead of per write call. Default-options writes get 15-33% faster depending on mode; written bytes and CRC values are unchanged.

CountingCrcWriter updated its crc32fast hasher on every write call. Record serialization emits individual fields as 1-8 byte writes, and crc32fast only reaches its PCLMULQDQ fast path on bulk input — tiny updates run the scalar fallback, which is more than an order of magnitude slower per byte (measured on one machine, hashing 128 MiB: 3 ms in 1 MiB updates, 30 ms in 100 B updates, 245 ms in 8 B updates). A profile of an unchunked 1M-message write showed 36% of total time inside crc32fast::baseline::update_fast_16, and the same write with CRCs disabled matched the Go and C++ writers. The Go writer avoids the problem structurally by serializing each record into a buffer and hashing it in one update.

Batch the hashing instead: writes still go to the inner writer immediately (I/O ordering, flush behavior, and position tracking are untouched), but bytes are folded into the hasher only once 32 KiB has accumulated, or directly for writes that are already at least that large. current_checksum() and finalize() account for pending bytes, so every observable CRC is identical to before.

Writing 1M 100-byte messages with default options (all CRCs on), same payload, best of 3:

- unchunked: 277 ms -> 197 ms (-29%)
- chunked:   410 ms -> 274 ms (-33%)
- zstd:      993 ms -> 954 ms (compression-dominated)
- lz4:       626 ms -> 535 ms (-15%)

This closes most of the gap to the Go and C++ writers on the uncompressed modes while keeping the Rust writer's strongest-in-the-repo integrity defaults (chunk, data section, and summary CRCs all on).

### DIFF
--- a/rust/mcap/src/io_utils.rs
+++ b/rust/mcap/src/io_utils.rs
@@ -2,29 +2,34 @@ use std::io::{self, prelude::*};
 
 use crc32fast::Hasher;
 
+/// Bytes are hashed in batches of at least this size (except at drain points), so the
+/// hasher's SIMD fast path gets bulk input even when the writer receives many small
+/// writes. Record serialization emits individual fields as 1-8 byte writes; hashing at
+/// that granularity keeps crc32fast on its scalar fallback, which is more than an order
+/// of magnitude slower than hashing the same bytes in bulk.
+const HASH_BATCH_SIZE: usize = 32 * 1024;
+
 pub struct CountingCrcWriter<W> {
     inner: W,
     hasher: Option<Hasher>,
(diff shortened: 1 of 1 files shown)


clalancette:
CI is going to fail here until https://github.com/foxglove/mcap/pull/1815 merges.

gasmith (approved):

PR #1823, opened 2026-08-25
rust: reject oversized chunk compression strings before allocating

### LINKED ISSUE #1813: Reading a summary allocates gigabytes from a small file
[mcap-summary-overallocation.mcap.zip](https://github.com/user-attachments/files/31374731/mcap-summary-overallocation.mcap.zip)

**Crate:** `mcap` 0.25.0 (crates.io, unmodified) · **Found by:** fuzzing a viewer that opens files supplied by users

## What happens

A **191 KB** file drives peak RSS to **1.69 GB** while its summary is read. The allocation happens before the reader concludes anything is wrong: it eventually returns `Err`, but only after the memory has been taken. On a machine with a 2 GB ceiling the process is killed instead.

Both documented entry points are affected:

| entry point | peak RSS on a 191 KB input |
|---|---|
| `mcap::Summary::read(&bytes)` | 1.69 GB |
| the `sans_io::SummaryReader` loop copied verbatim from its rustdoc example | 1.69 GB |

The second one still reaches 1.69 GB **with `record_length_limit` set** to 64× the file size, which is what makes this worth reporting: a caller who has read the docs and set the available bound is not protected.

## Where

The growth is the reader's own buffer, not a single record's declared length:

```
raw_vec::RawVecInner::finish_grow
raw_vec::RawVecInner::reserve::do_reserve_and_handle
mcap::sans_io::linear_reader::LinearReader::insert
mcap::sans_io::summary_reader::SummaryReader::next_event
```

`compute_record_length_limit` bounds what one record may *declare*; it does not bound how large the buffer becomes across the read requests that fill it. A file that keeps the reader asking grows it without limit.

## Reproducer

`mcap-summary-overallocation.mcap` (attached, 191,486 bytes) — a fuzzer-found input with its strings replaced by same-length filler, so every offset and declared length still means what it did:

```rust
fn main() {
    let bytes = std::fs::read(std::env::args().nth(1).unwrap()).unwrap();
    println!("{:?}", mcap::Summary::read(&bytes));
}
```

```
$ /usr/bin/time -l ./repro mcap-summary-overallocation.mcap
Err(UnexpectedEof)          # "MCAP file ended in the middle of a record"
1687060480  maximum resident set size
```

1.69 GB of peak RSS for a 191 KB file, and the same figure through the sans-io loop with `record_length_limit` set.

## Suggested direction

Bound the reader's buffer by what the file can actually contain — the summary section cannot be larger than the file, and a read request larger than the bytes that remain can never be satisfied. Clamping the request to the remaining bytes is a one-line change at the call site and fixed it for us; doing it inside the reader would protect everyone using the documented example.

official-burak:
### Changelog
- rust: return an error instead of allocating when a chunk header declares a compression string larger than the chunk record or the configured record length limit
- rust: clamp summary-reader `ReadRequest` sizes to the remaining file length when the file size is known

### Docs
None.

### Description

`LinearReader` loads a chunk header in two steps: first the fixed prefix, then `MIN_CHUNK_HEADER_SIZE + compression_len` bytes. `compression_len` is taken from the file and was passed to `consume!` / `insert()` with no check against the chunk record length or `record_length_limit`. `insert(n)` resizes the buffer *before* any bytes are read, so a 191 KiB file that declared a ~1.69 GiB compression string drove peak RSS to about 1.69 GiB and then returned `UnexpectedEof`.

`record_length_limit` did not prevent this. The limit is applied to parsed record bodies, and the chunk path skipped that check until after the header (including the compression string) had been requested.

This PR:
1. Rejects a compression-string length that exceeds the chunk record body (`BadChunkLength`).
2. Applies `record_length_limit` to that header length (`ChunkTooLarge`).
3. Caps `SummaryReader` read requests to the remaining file bytes when `file_size` is known, so the documented sans-io loop cannot allocate more than the file can contain.
4. Caps decompressor `next_read_size()` by `compressed_remaining`.

Well-formed files are unchanged. Regression tests cover the 1.69 GiB compression-string case both with a mismatched record length and with a matching length plus a 64x file-size record limit.

Fixes #1813

### DIFF
--- a/rust/mcap/src/sans_io/linear_reader.rs
+++ b/rust/mcap/src/sans_io/linear_reader.rs
@@ -571,7 +571,20 @@ impl LinearReader {
                     let min_header_buf = load!(MIN_CHUNK_HEADER_SIZE);
                     let compression_len =
                         u32::from_le_bytes(min_header_buf[28..32].try_into().unwrap());
-                    let header_len = MIN_CHUNK_HEADER_SIZE + compression_len as usize;
+                    // The compression string sits between the length prefix and compressed_size.
+                    // A corrupt file can declare a multi-gigabyte string here. Requesting that
+                    // many bytes used to allocate them in insert() before any length check ran.
+                    let header_len =
+                        (MIN_CHUNK_HEADER_SIZE as u64).saturating_add(u64::from(compression_len));
+                    if header_len > len {
+                        return Some(Err(McapError::BadChunkLength {
+                            header: header_len,
+                            available: len,
+                        }));
+                    }
+                    let header_len =
+                        check!(check_len(header_len, self.options.record_length_limit)
+                            .ok_or(McapError::ChunkTooLarge(header_len)));
                     let header_buf = consume!(header_len);
                     let header: ChunkHeader = check!(std::io::Cursor::new(header_buf).read_le());
                     // Re-use or construct a compressor
@@ -843,7 +856,9 @@ fn decompress_inner(
     }
     dest_buf.reserve_exact(additional);
     loop {
-        let need = decompressor.next_read_size();
+        let need = decompressor
+            .next_read_size()
+            .min(clamp_to_usize(*compressed_remaining));
         let have = src_buf.len();
         if need > have {
             return Ok(Some(need - have));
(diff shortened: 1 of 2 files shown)


clalancette (approved):
Thanks, this is a nice improvement.  I've found similar problems in most of the rest of the languages, which I'll fix in a follow-up PR.

Issue #1814, opened 2026-08-24
An uncompressed chunk whose declared sizes disagree panics the reader


Nandite:
[mcap-uncompressed-chunk-panic.mcap.zip](https://github.com/user-attachments/files/31374796/mcap-uncompressed-chunk-panic.mcap.zip)

**Crate:** `mcap` 0.25.0 (crates.io, unmodified) · **Found by:** fuzzing a viewer that opens files supplied by users

## What happens

`sans_io::indexed_reader::insert_chunk_record_data` validates the data it is given against the chunk index's `compressed_size`, and then, for a chunk whose compression is `""`, does this:

```rust
"" => {
    slot.buf.resize(uncompressed_size, 0);
    slot.buf[..].copy_from_slice(compressed_data);
}
```

`uncompressed_size` comes from the chunk index, `compressed_data.len()` has been checked against `compressed_size`, and **nothing checks the two sizes against each other**. For a chunk stored as-is they must be equal, so a file declaring `compressed_size: 10, uncompressed_size: 999` produces

```
panicked at src/sans_io/indexed_reader.rs:333:
source slice length (10) does not match destination slice length (999)
```

A panic rather than an `Err` matters here: a viewer that opens files from outside cannot catch this without wrapping every read in `catch_unwind`, and one malformed file takes down whatever else the process was showing.

## Suggested patch

```rust
"" => {
    if compressed_data.len() != uncompressed_size {
        return Err(McapError::DecompressionError(format!(
            "an uncompressed chunk declares {uncompressed_size} bytes and carries {}",
            compressed_data.len()
        )));
    }
    slot.buf.clear();
    slot.buf.extend_from_slice(compressed_data);
}
```

This is what we are running against locally, and it turns the crash into a refusal without changing behaviour for well-formed files.

## Reproducer

`mcap-uncompressed-chunk-panic.mcap.zip` (attached — GitHub will not take a `.mcap`; it unzips to `mcap-uncompressed-chunk-panic.mcap`, 191,486 bytes), against unmodified `mcap = "0.25.0"`:

```rust
use mcap::sans_io::{IndexedReadEvent, IndexedReader, IndexedReaderOptions};

fn main() {
    let bytes = std::fs::read(std::env::args().nth(1).unwrap()).unwrap();
    let summary = mcap::Summary::read(&bytes).unwrap().unwrap();
    let topics: Vec<String> = summary.channels.values().map(|c| c.topic.clone()).collect();
    for topic in &topics {
[...]

PR #1824, opened 2026-08-25
c++, go: Fix uncompressed chunk size panic

clalancette:
### Changelog

None.

### Docs

None.

### Description

We recently got bug report #1814, with the follow-up fix in #1820.  That bug and fix found that in the Rust client, we weren't properly detecting an uncompressed chunk-size mismatch and panicing.

While investigating that further, I found that both C++ and Go have similar problems.  This PR does 3 things:

1.  Fixes C++ to detect and reject the uncompressed chunk-size mismatch, like Rust.
2.  Ditto for Go.
3.  Does a very small performance improvement for Rust, since we don't need to zero-fill the buffer.

### DIFF
--- a/go/mcap/parse_test.go
+++ b/go/mcap/parse_test.go
@@ -258,3 +258,57 @@ func TestParseSchema(t *testing.T) {
 		})
 	}
 }
+
+func TestParseChunk(t *testing.T) {
+	records := []byte{1, 2, 3}
+	makeChunk := func(uncompressedSize uint64, compression string, recordsLength uint64, records []byte) []byte {
+		return flatten(
+			encodedUint64(0),                // message start time
+			encodedUint64(0),                // message end time
+			encodedUint64(uncompressedSize), // uncompressed size
+			encodedUint32(0),                // uncompressed CRC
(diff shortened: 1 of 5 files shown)


go/mcap/parse_test.go
```diff
@@ -258,3 +258,57 @@ func TestParseSchema(t *testing.T) {
 		})
 	}
 }
+
+func TestParseChunk(t *testing.T) {
```
bennetthardwick, line 262: Could it be worth porting these tests to `tests/conformance` instead so they run against all languages?

bennetthardwick (approved):

go/mcap/parse_test.go
clalancette: That is probably worth doing as a follow-up.  The reason not to do it here is that the conformance tests are currently testing well-formed files only; we'll have to add a new harness and tests for this class of problems.

What I'm going to do is to merge this PR as-is, then open a follow-up issue to add in a new class of tests to the conformance tests.  Thank you for the review!

clalancette: And now filed as #1828

PR #1826, opened 2026-08-26
cli: resolve S3 credentials through the AWS SDK default chain

clalancette:
### Changelog

Restore full AWS credential functionality, including reading credentials from environment variables and the ~/.aws/credentials file.

### Docs

None.

### Description

object_store only reads credentials from env vars and instance metadata, so ~/.aws/credentials, AWS_PROFILE, SSO sessions, and credential_process silently did not work — a regression from the Go CLI, which used the SDK default chain (#1803).

Bridge aws-config's default chain into object_store via a custom CredentialProvider for s3:// inputs. Precedence matches the aws CLI (env vars first), a region in ~/.aws/config is picked up when no env var sets one, and the chain refreshes expiring credentials mid-download. Other stores keep object_store's env-var mechanisms.

The SDK's HTTP-backed providers run on a ring-backed rustls client instead of the default aws-lc one, avoiding a cmake-built C dependency that the zigbuild release cross-builds cannot compile.  Makes the stripped release binary grow from 15MB -> 16.2MB.


rust/cli/src/aws_credentials.rs
```diff
@@ -0,0 +1,313 @@
...
+        if let Some(region) = runtime.block_on(default_chain_region()) {
+            builder = builder.with_region(region);
+        }
+    }
+    for (key, value) in options {
+        if let Ok(key) = key.to_ascii_lowercase().parse::<AmazonS3ConfigKey>() {
+            builder = builder.with_config(key, value);
+        }
+    }
+    builder = builder.with_credentials(Arc::new(SdkChainCredentialProvider::default_chain()));
```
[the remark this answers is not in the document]
clalancette: The chain self-resolves its region from the real-fs/env ProviderConfig, so this is a non-issue.

rust/cli/src/sdk_s3.rs
```diff
@@ -0,0 +1,711 @@
...
+        let size = response.content_length().unwrap_or(0).max(0) as u64;
+        let meta = ObjectMeta {
+            location: location.clone(),
+            last_modified: smithy_datetime_to_chrono(response.last_modified()),
+            size,
+            e_tag: response.e_tag().map(str::to_string),
+            version: None,
+        };
+        Ok(GetResult {
+            payload: GetResultPayload::Stream(futures_util::stream::empty().boxed()),
```
bennetthardwick, line 341: Is this how `object_store` impls a head request too? Interesting.

rust/cli/src/sdk_s3.rs
```diff
@@ -0,0 +1,711 @@
...
+        if let Some(range) = &options.range {
+            request = request.range(http_range_header(range));
+        }
+        if let Some(etag) = &options.if_match {
+            request = request.if_match(etag);
+        }
+        let response = request
+            .send()
+            .await
+            .map_err(|err| map_sdk_error(location, err))?;
```
bennetthardwick, line 225: I would expect you to call `error_for_status_ref` or whatever and handle the status code. Is status code handling done somewhere else?

bennetthardwick: Ah this is not `reqwest` nevermind - seems like the SDK handles this. I am seeing some weird errors though on unauthenticated accesses.

bennetthardwick (approved):
Makes sense to me.

I've tested the using `AWS_PROFILE` to auth with sso and using `AWS_SESSION_TOKEN` from `aws-vault` and it works well.

When you're not authenticated (or don't have access) the error that comes back is really vague. It would be great to improve that if possible:

```txt
Error: failed to read s3://<bucket>/<private file>.map
Failed while fetching range from: Generic S3 error: dispatch failure
```

The `AWS_REGION` one is great and actionable. This one will definitely confuse people.

clalancette:
> When you're not authenticated (or don't have access) the error that comes back is really vague. It would be great to improve that if possible:
> 
> ```
> Error: failed to read s3://<bucket>/<private file>.map
> Failed while fetching range from: Generic S3 error: dispatch failure
> ```
> 
> The `AWS_REGION` one is great and actionable. This one will definitely confuse people.

Yeah, this is a great point.  The latest push should make that a lot more clear and actionable.

PR #1827, opened 2026-08-28
kotoba: add v1 in-language MCAP codec

com-junkawasaki:
Add a first-class Kotoba binding next to cpp/ go/ python/ rust/ swift/ typescript/.

Kotoba cannot do ambient FFI, so this is an in-language codec. v1 covers magic, Header, DataEnd, and Footer only. Schema/Channel/Message/Chunk/indexes are follow-up.

Kotoba CI on the fork is green. Fork PR: https://github.com/kotoba-lang/mcap/pull/1


amacneil:
Thank you for the submission. I believe this would be best maintained as a separate community repo, rather than in the Foxglove-maintained mcap monorepo.

PR #1835, opened 2026-09-07
typescript: forward optional read options through IReadable

achim-k:
### Changelog
typescript: allow passing optional arguments to an IReadable implementation

### Docs
None

### Description
This PR extends `IReadable.read()` with an optional argument. Callers can pass extra context to a custom `IReadable`. One use case is to pass tracing information.

`McapIndexedReader` forwards the same value through `Initialize`, `readMessages`, `readMetadata`, and 

Example:

```ts
type ReadOptions = { tracingId: string };

class MyReadable implements IReadable<ReadOptions> {
  // ...

  async read(offset: bigint, size: bigint, options?: ReadOptions): Promise<Uint8Array> {
    console.log(options);
    // ...
  }
}

const readable = new MyReadable(/* ... */);
const reader = await McapIndexedReader.Initialize({ readable });

for await (const message of reader.readMessages({
  topics: ["foo"],
  readOptions: { tracingId: "abc123" },
})) {
  // ...
}
```

### DIFF
--- a/typescript/core/src/CachedReadable.ts
+++ b/typescript/core/src/CachedReadable.ts
@@ -1,9 +1,10 @@
 import type { IReadable } from "./types.ts";
 
 /**
- * Wraps an {@link IReadable} and caches the bytes returned from `read()` keyed by offset. Reads at
- * an offset that was previously cached return a subarray of the cached buffer instead of calling
- * the underlying readable.
+ * Wraps an {@link IReadable} and caches the bytes returned from `read()` keyed by offset. The
+ * `size` and `options` parameters of `read()` are not part of the cache key. Reads at an offset
+ * that was previously cached return a subarray of the cached buffer instead of calling the
+ * underlying readable.
  *
  * The cache is capped by `maxCacheSizeBytes`; once the cache is full, new reads pass through
  * without being cached. No eviction is performed.
--- a/typescript/core/src/CachedReadable.ts
+++ b/typescript/core/src/CachedReadable.ts
@@ -11,11 +11,11 @@ import type { IReadable } from "./types.ts";
  * Note: reads are cached by *exact* offset. A read that partially overlaps a cached range but
  * starts at a different offset will miss.
  */
-export class CachedReadable implements IReadable {
+export class CachedReadable<TReadOptions = unknown> implements IReadable<TReadOptions> {
   /**
    * The underlying source of the data to be cached.
    */
-  #readable: IReadable;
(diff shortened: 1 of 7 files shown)


typescript/core/src/CachedReadable.ts
```diff
@@ -47,14 +47,14 @@ export class CachedReadable implements IReadable {
     return this.#size;
   }
 
-  async read(offset: bigint, size: bigint): Promise<Uint8Array> {
+  async read(offset: bigint, size: bigint, options?: TReadOptions): Promise<Uint8Array> {
     const requestedSize = Number(size);
     const cached = this.#cache.get(offset);
     if (cached != undefined && cached.byteLength >= requestedSize) {
       return cached.byteLength === requestedSize ? cached : cached.subarray(0, requestedSize);
```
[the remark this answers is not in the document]
achim-k: Expected behavior

typescript/core/src/CachedReadable.ts
```diff
@@ -47,14 +47,14 @@ export class CachedReadable implements IReadable {
     return this.#size;
   }
 
-  async read(offset: bigint, size: bigint): Promise<Uint8Array> {
+  async read(offset: bigint, size: bigint, options?: TReadOptions): Promise<Uint8Array> {
```
snosenzo, line 51: Maybe worth noting here that the `options` are not incorporated into the cache key if they are something in them that could affect reading underneath.

typescript/core/src/McapIndexedReader.ts
```diff
@@ -124,7 +125,11 @@ export class McapIndexedReader {
      * that later queries against different channels can be served from the cache.
      */
     messageIndexCacheSizeBytes?: number;
-  }): Promise<McapIndexedReader> {
+    /**
+     * Opaque options passed to each `readable.read()` call during initialization.
+     */
+    readOptions?: TReadOptions;
```
snosenzo, line 131: should this be read and stored so that later `readMessages/readMetadata/readAttachments` can also use the readOptions passed in initialization? Or do we want it to pass `undefined`?

snosenzo (approved):

typescript/core/src/McapIndexedReader.ts
achim-k: We could, but I believe it's more flexible if a user can pass these to each individual `readMessages/readMetadata/readAttachments` call. For instance, when passing tracing information, those would change with each `readMessages` call.

PR #1837, opened 2026-09-07
rust: restore deterministic summary record ordering

jrbronkar:
### Changelog

- Rust: serialize repeated schema and channel records in deterministic ID order

### Docs

None.

#### Human Authored:

The reason for this change is I was validating a deterministic pipeline replay system, and noticing that my mcaps were not byte-for-byte identical because of non-determinism that I isolated to this library. This made it so that I couldn't just hash the files to diff them, I had to do parsing and normalization here, as the random summary ordering was causing byte-level diffs.

#### AI Assisted for further context:

Before #1364, the Rust writer serialized repeated schema and channel records directly from its ID-keyed `BTreeMap`s. Their iteration order—and therefore the generated MCAP bytes—was stable and sorted by numeric ID.

#1364 changed `Writer::finish()` to return a public `Summary`. As part of that refactor, the writer began copying its ordered schema and channel bookkeeping into the `HashMap`s exposed by `Summary`, then serializing by iterating those maps. Because Rust's default `HashMap` uses a randomly seeded hasher, equivalent writer runs can emit those records in different orders. That also changes record positions, summary offsets, and the summary CRC.

Sort repeated schemas and channels by numeric ID at the serialization boundary. This removes the writer's randomly seeded `HashMap` ordering from generated files. It does not attempt to guarantee byte-identical output across different compression-library versions, writer configurations, platforms, or input record orderings.

A regression test writes schemas and channels with deliberately unordered IDs and verifies that their summary records are ordered by ID.

Developed and validated with assistance from OpenAI Codex.

### Testing

- `cargo test -p mcap --all-features`
- `cargo clippy -p mcap --all-targets -- --no-deps -D warnings`
- `cargo fmt --all -- --check`
- Built a minimal uncompressed writer against the parent and fixed commits and ran each in 20 separate processes. The parent commit produced 20 distinct SHA-256 hashes; the fixed commit produced one hash across all 20 files, and `cmp` confirmed that the first and last fixed outputs were byte-identical.

### DIFF
--- a/rust/mcap/src/write.rs
+++ b/rust/mcap/src/write.rs
@@ -1295,7 +1295,7 @@ fn write_summary_and_footer_magic<W: Write + Seek>(
     summary: &Summary,
     options: &WriteOptions,
 ) -> McapResult<()> {
-    let all_channels: Vec<_> = summary
+    let mut all_channels: Vec<_> = summary
         .channels
         .iter()
         .map(|(&id, channel)| {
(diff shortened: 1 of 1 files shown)


james-rms:
Hi @jrbronkar , I slightly prefer this approach: https://github.com/foxglove/mcap/pull/1838

jrbronkar:
> Hi @jrbronkar , I slightly prefer this approach: #1838

thanks for the quick response! sounds good, I agree. I will close this in favor of #1838.

PR #1839, opened 2026-09-10
python: add `Writer.flush()` to finish the open chunk on demand

simobenziane:
The Python `Writer` finalizes a chunk only when its size exceeds `chunk_size` (`__maybe_finalize_chunk`); there is no public way to close the chunk in progress. The C++ writer has `McapWriter::closeLastChunk()` and the TypeScript writer has `flush()` for exactly this — a recorder that wants large chunks for compression but a bounded crash-loss window calls it on a timer. This adds `Writer.flush()`: finish the open chunk if it holds any message, then write the buffered records to the stream. No behaviour changes for callers that never call it; `finish()` is unchanged. Two tests: `flush()` writes the open chunk below `chunk_size` and the written prefix is readable by a linear scan; `flush()` with no open chunk writes nothing.

Context: a companion-robot recorder (Python, pydantic messages as JSON channels) measured its crash-loss window as one chunk's fill time plus a 1 s file flush; with 64 KiB chunks at ~50 KB/s that is ≈ 2.4 s, and the alternative — 8 KiB chunks — doubles the file. `flush()` on the timer makes the window the timer alone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/mcap/mcap/writer.py
+++ b/python/mcap/mcap/writer.py
@@ -501,9 +501,8 @@ class Writer:
 
     def flush(self):
         """Finishes the chunk in progress, if any, and writes everything buffered so far to
-        the output stream -- the Python counterpart of the C++ writer's ``closeLastChunk()`` and
-        the TypeScript writer's ``flush()``. Call it on a timer to bound how many messages a
-        crash can lose regardless of ``chunk_size``; a chunk with no messages is not written.
+        the output stream. A chunk with no messages is not written. Compression works per
+        chunk, so flushing often reduces the compression ratio.
         """
         self.__finalize_chunk()
         self.__flush()
--- a/python/mcap/mcap/writer.py
+++ b/python/mcap/mcap/writer.py
@@ -499,6 +499,15 @@ class Writer:
         self.__chunk_indices.append(chunk_index)
         self.__chunk_builder.reset()
 
+    def flush(self):
+        """Finishes the chunk in progress, if any, and writes everything buffered so far to
+        the output stream -- the Python counterpart of the C++ writer's ``closeLastChunk()`` and
+        the TypeScript writer's ``flush()``. Call it on a timer to bound how many messages a
+        crash can lose regardless of ``chunk_size``; a chunk with no messages is not written.
+        """
+        self.__finalize_chunk()
+        self.__flush()
+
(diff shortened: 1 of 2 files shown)


python/mcap/mcap/writer.py
```diff
@@ -499,6 +499,15 @@ def __finalize_chunk(self):
         self.__chunk_indices.append(chunk_index)
         self.__chunk_builder.reset()
 
+    def flush(self):
+        """Finishes the chunk in progress, if any, and writes everything buffered so far to
+        the output stream -- the Python counterpart of the C++ writer's ``closeLastChunk()`` and
+        the TypeScript writer's ``flush()``. Call it on a timer to bound how many messages a
```
gasmith: I think it's kinda weird for this docstring to be referencing methods and behavior from other language bindings. A self-contained explanation of what this does and how to use it is sufficient.

python/mcap/mcap/writer.py
```diff
@@ -499,6 +499,15 @@ def __finalize_chunk(self):
...
+        crash can lose regardless of ``chunk_size``; a chunk with no messages is not written.
+        """
```
gasmith: Ultra-nit: The timer would put a lower bound on the age of messages that might be lost relative to the time of the crash, not the number of messages... I think people broadly understand how "flush" works and how to use it to ensure that buffered messages are persisted to disk. Maybe we don't need to try and describe this in such detail.

On the other hand, it might be worth mentioning that compression benefits from larger chunks, and so flushing too frequently may reduce the compression ratio.

gasmith (approved):
LG, though I have some nit-picky comments about the doc string that I'll leave to your discretion.

python/mcap/mcap/writer.py
simobenziane: Agreed — the docstring is self-contained now (no other bindings named).

python/mcap/mcap/writer.py
simobenziane: Dropped the crash-loss sentence and added the compression note instead: a chunk with no messages is not written; compression works per chunk, so flushing often reduces the ratio. Thanks for the review.

gasmith:
@simobenziane thanks for the contribution! Sorry it took a while to get back to it.

PR #1841, opened 2026-09-16
rust: reuse zstd compression context across chunks

csmith105:
### Changelog

Writing zstd-compressed MCAP files no longer spawns a fresh set of `compression_threads` worker threads for every chunk -- the compression context (and its thread pool) is now reused across chunks in the same file.

### Docs

None.

### Description

`ChunkWriter::new()` builds a brand-new `zstd::stream::raw::Encoder` -- and, when `compression_threads > 1`, spawns that many new OS threads via `NbWorkers` -- every time a chunk starts, then drops it entirely in `ChunkWriter::finish()`. On a file with hundreds of chunks, that's hundreds of thread-pool spin-ups and tear-downs over the life of one `Writer`, for zero benefit: the crate already implements `Operation::reinit()` for `Encoder`, which calls `ZSTD_reset_session_only` -- resetting the compression session for a new frame while leaving every parameter, including `NbWorkers`, exactly as it was. Nothing was calling it.

This is not just a bare-thread-spawn cost either: recreating the `Encoder` re-initializes zstd's internal compression buffers and window tables from scratch each time, at whatever `compression_level` was configured. For a downstream writer producing a real 223-chunk file, this showed up as ~450 `clone3` syscalls over a single write session, plus measurable wall-clock overhead on top of that.

The fix: `Writer<W>` now holds the most recently finished chunk's encoder in a `reusable_zstd_encoder` field instead of letting `ChunkWriter::finish()` drop it. `ChunkWriter::new()` calls `.reinit()` on that encoder when one's available, and only falls back to building a fresh one for the very first chunk. The two zstd-only inputs to `ChunkWriter::new()` (`compression_threads`, and now the reused encoder) are bundled into one `ZstdChunkState` struct so the function doesn't trip clippy's `too_many_arguments`. Everything is `#[cfg(feature = "zstd")]`-gated the same way `compression_threads` already was.

**Testing:**
- `cargo test -p mcap --all-features`: 71/71 pass (after `git lfs pull` for fixture data).
- `cargo clippy -p mcap --all-targets -- --no-deps -D warnings`: clean, no new `#[allow]`s.
- `cargo fmt --all -- --check`: clean.
- Manual: rebuilt a downstream Rust writer (a nav-filter replay tool) against this patch via `[patch.crates-io]` on two different machines/architectures (x86_64 and arm64). Ran it against the same real input file before and after.

`strace -f -c` on one write session: `clone3` called **447** times. Wall clock ~23.1s.

`mcap info` on the output: 223/223 chunks, all channels/messages present.

</td><td>

Same session, same input: `clone3` called **3** times (one real encoder init, matching the two other transient threads in the process). Wall clock ~21.6s.

`mcap info` on the output: identical -- 223/223 chunks, same channel/message counts, same durations. Byte-for-byte equivalent decode.

</td></tr></table>

No functional change to the format written -- same compression, same chunk boundaries, same parameters per chunk. Only the lifetime of the in-process `Encoder` changes.


csmith105:
@gasmith any thoughts on this?

kvnloo:
I traced the reuse path through the "zstd" Rust implementation and the mechanism looks sound to me: "Encoder::reinit()" resets the compression session for a new frame while retaining the context parameters, which is what we want here for keeping the "NbWorkers" setup alive across chunks.

The before/after thread-count measurement is also a strong validation of the intended effect.

One thing I'd like to see pinned down with a focused regression test is the lifecycle boundary itself. The existing suite passing is useful, but this PR specifically depends on the second and later chunks going through the reused-encoder path.

Could we add a small test that:

1. configures zstd + a small "chunk_size" so several chunks are guaranteed,
2. writes enough messages for at least 2–3 chunks,
3. finishes and reopens the file,
4. verifies the expected chunk/message counts and message payloads?

That would make a future change to "ChunkWriter::finish()", "zio::Writer", or encoder ownership fail directly if reuse ever starts leaking frame state or producing a malformed subsequent frame.

On non-WASM MCAP already enables zstd's "zstdmt" feature, so if it's practical it may also be worth running that case with "compression_threads > 1" to exercise the exact context configuration this optimization is intended to preserve.

Otherwise the change looks nicely scoped.

AI assistance disclosure: reviewed with OpenAI ChatGPT GPT-5.6 Sol.

csmith105:
> I traced the reuse path through the "zstd" Rust implementation and the mechanism looks sound to me: "Encoder::reinit()" resets the compression session for a new frame while retaining the context parameters, which is what we want here for keeping the "NbWorkers" setup alive across chunks.
> 
> The before/after thread-count measurement is also a strong validation of the intended effect.
> 
> One thing I'd like to see pinned down with a focused regression test is the lifecycle boundary itself. The existing suite passing is useful, but this PR specifically depends on the second and later chunks going through the reused-encoder path.
> 
> Could we add a small test that:
> 
> 1. configures zstd + a small "chunk_size" so several chunks are guaranteed,
> 2. writes enough messages for at least 2–3 chunks,
> 3. finishes and reopens the file,
> 4. verifies the expected chunk/message counts and message payloads?
> 
> That would make a future change to "ChunkWriter::finish()", "zio::Writer", or encoder ownership fail directly if reuse ever starts leaking frame state or producing a malformed subsequent frame.
> 
> On non-WASM MCAP already enables zstd's "zstdmt" feature, so if it's practical it may also be worth running that case with "compression_threads > 1" to exercise the exact context configuration this optimization is intended to preserve.
> 
> Otherwise the change looks nicely scoped.
> 
> AI assistance disclosure: reviewed with OpenAI ChatGPT GPT-5.6 Sol.

Thanks @kvnloo, I've added the specific test. Note that my agent caught the following:

> I stress-tested the test itself by temporarily deleting the reinit() call to confirm it would catch exactly the failure mode they described ("leaking frame state"). It didn't fail — `libzstd` already resets its compression context for a fresh frame after a clean `ZSTD_e_end` completion, independent of an explicit `ZSTD_reset_session_only` call, so that specific omission isn't something a correctness-only test can catch here. I left a comment in the test being upfront about this rather than overclaiming what it proves. The test still gives real regression coverage on the actual reuse plumbing (mem::take, `ChunkWriter::finish()`'s encoder hand-back, multi-chunk byte layout under a multithreaded context) — just not that one specific mechanism.

csmith105:
@kvnloo, @gasmith, @bennetthardwick, @james-rms any additional comments?

kvnloo:
nice mutation check. i took another pass at the ownership invariant and i think there are two small things worth tightening.

first, i think the test comment still slightly overstates what the regression test protects. removing "reinit()" is one useful mutation, but if the encoder hand-back were removed entirely, or "ChunkWriter::new()" simply went back to constructing a fresh encoder every chunk, this round-trip test should still pass. so i'd describe this as protecting multi-chunk correctness under the new implementation, while the before/after "clone3" measurement is the evidence that the reuse optimization itself is actually happening.

there's also one lifecycle edge i hadn't noticed before: after "Writer::finish()" calls "finish_chunk()", the encoder gets handed back into "self.reusable_zstd_encoder", but no future chunk can legally start after the writer is finished. that means a finished-but-still-live "Writer" can keep the zstd context / worker pool / allocated buffers alive until "Writer" is dropped. since "compression_threads" defaults to the physical CPU count, that retention could be nontrivial.

would it make sense to clear "reusable_zstd_encoder" immediately after the final "finish_chunk()?" in "Writer::finish()"? that seems to make the ownership boundary exact: retain the context only while another chunk can still exist, then release it once reuse is impossible.

other than that, i still think the context-reuse direction itself is the right root fix.

csmith105:
@kvnloo 

Agreed on all points, see [#ea6b4a0](https://github.com/foxglove/mcap/pull/1841/commits/ea6b4a086bc77899b00e6a726d3aa3ea6d65bef0)

kvnloo:
heya 👋 thanks for working through this with me and for being patient with the extra passes - really appreciate the collaboration here!

the latest change addresses both things i was worried about: the test now describes its guarantee precisely, and dropping the reusable encoder at "finish()" makes the ownership boundary clean.

looks good to me now. nice work getting this across the line.

csmith105:
@kvnloo we (me and Anduril) appreciate your collaboration, this PR will speed up one of our internal processes by ~15% and on an HPC that spawns hundreds of these jobs slash thread creation counts from thousands to dozens.

james-rms (approved):

james-rms:
Thanks for your contribution @csmith105!

kvnloo:
wait that's crazy, i was just talking to one of my friends at anduril about some of the stuff he's working on, i was essentially asking him how to fast track learning about robotics based on my past experience & how i can best contribute 😁

what a small world haha

csmith105:
It's a cool company, lots of small teams doing neat things in the drone space. I'll keep it sterile in the PR comments, feel free to connect on LinkedIn and we can chat. https://www.linkedin.com/in/csmith105/

